PR #6, opened 2015-12-15
find packages and pip

lasote:


memsharded:
Not sure it is a good idea to merge to master, as it will get published to RTD, but the features are still not released. I suggest opening develop branch on docs, and merge there.

Anyway, the PR is good, can be merged, good work!

PR #33, opened 2016-01-20
Correct wording for SPHINXBUILD to 'command line'

dragly:
Setting the SPHINXBUILD environment  variable doesn't work, so the wording should be to set this on the command line.


lasote:
The **Makefile** file is generated by sphinx (http://www.sphinx-doc.org) and I have checked that the file is generated with the same statement in the latest version. I think this issue should be opened in sphinx repository, maybe the authors want to check why if it really doesn't work with env variable.

Thanks!

PR #39, opened 2016-02-09
This needs to be called "requires" in conan 0.7.2

monsdar:
I tried the example and had problems. Turns out that the option block needs to be named "requires" instead of "requirements"

### DIFF
--- a/examples/zmq_protobuf.rst
+++ b/examples/zmq_protobuf.rst
@@ -161,7 +161,7 @@ If not created, then create also a ``conanfile.txt`` with the following content:
 
 .. code-block:: text
 
-   [requirements]
+   [requires]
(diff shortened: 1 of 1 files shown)


memsharded:
Yes, thanks, that was a mistake, merged.
If you are trying the zmq and protobuf libs, I am currently working on them, probably upload new packages soon for more configs, and also containing FindCmake files for them, so they work almost transparently for existing cmake projects. 

Thanks for your contribution!

monsdar:
Yeah, I plan on pushing some projects that are mainly based on ZMQ and Protobuf. Looking forward to your changes. [Here's](https://github.com/monsdar/FitzComm) a first try to use Conan and CMake... Being able to strip the Conan-dependencies out of the main CMake would be a great thing.

PR #48, opened 2016-05-13
Fixed the qbs acronym expansion

arBmind:
See https://doc.qt.io/qbs/ for the intended name.


memsharded:
Ups, great, thank you!

PR #54, opened 2016-06-30
Release/0.10

lasote:
I'm not sure about the section I put the scopes docs... You only know why you shouldn't use options when you know about package creation.


memsharded:
Already merged, good job!

Issue #63, opened 2016-08-30
update option in conan is not properly documented


memsharded:
As pointed out by @mnowy in https://github.com/conan-io/conan/issues/101#issuecomment-243455607, it is true that this important feature is undocumented, this has to be fixed.

PR #64, opened 2016-08-31
Update option documented

lasote:
Docs for --update option (#63). Directly to master!


memsharded:
It is fine, but I think it deserves a how-to or another section specifying a bit the flows and the reasons.

PR #68, opened 2016-09-06
Fix name of Qbs - it's now just "Qbs"

jakepetroules:
We internally decided not to call it "Qt build suite" anymore.

See https://codereview.qt-project.org/#/c/161228/ and https://doc.qt.io/qbs/


memsharded:
Thanks Jake for fixing this!

PR #70, opened 2016-09-08
Replace test by test_package

menuet:
Hi,
I saw a few places where "test" was still used instead of "test_package"

### DIFF
--- a/packaging/testing.rst
+++ b/packaging/testing.rst
@@ -111,7 +111,7 @@ use the **--build=never** option:
    ...
    Hello world!
    
-The ``conan test`` command receives the same command line parameters as ``conan install`` so you
+The ``conan test_package`` command receives the same command line parameters as ``conan install`` so you
 can pass to it the same settings, options, and command line switches.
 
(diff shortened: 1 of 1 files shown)


memsharded:
Thanks very much for fixing this! ``test`` should still work, but the correct syntax is indeed ``test_package``.
BTW, for PRs to the docs the policy is this:
- If it is fixing something of the current released conan, then it can go directly to master branch, so it is inmediately published to docs.conan.io
- The develop branch is to document changes that have been implemented in conan, but are not released yet, so they are in the develop branch of the code repo only.

I will just merge this one to master right now.
Thanks again for your contribution!

menuet:
Thanks for your thanks. I will try to remember about the policy for next time :-)
By the way, I really like the project conan. Keep going !

memsharded:
Changes now live: http://docs.conan.io/en/latest/packaging/testing.html
I have changed the default branch in the repo from develop=>master, so next time, it should be by default :) Thanks for your support!

PR #79, opened 2016-10-13
Fix comments in custom_cache.rst

sqrt:
Replace incorrect ``//`` comments with ``#`` in bash examples


memsharded:
Hi Julius, thanks for fixing this!

PR #105, opened 2016-11-29
Conan 0.16.0 requires node-semver

tivek:
Since version 0.16.0 conan requires the python port of node-semver. For Arch Linux it is packaged on AUR as python-node-semver.


memsharded:
Thanks for fixing this (and creating the arch linux 0.16 pkg :)

PR #140, opened 2017-01-18
Review of reference/commands.rst

fbauzac:
Some rephrases.

Split long verbatim lines, because otherwise the end cannot be seen on
the docs.conan.io website (need to horizontal scroll).

Add link to JWT.


memsharded:
Good job, as usual :)

PR #157, opened 2017-02-27
Add python-distro to the list of AUR dependencies

tivek:
python-distro was missing from the list of AUR dependencies for Arch Linux, add it to the instructions.


memsharded:
Yes, that is true, it is used now for better detection of the linux distro by ``OSInfo``, thanks!

PR #159, opened 2017-02-28
Document CMake configure & build wrappers (#158)

sixten-hilborn:
See #158.

### DIFF
--- a/conanfile_py.rst
+++ b/conanfile_py.rst
@@ -66,6 +66,8 @@ Build automation
 One advantage of using ``conanfile.py`` is that the project build can be further simplified,
 using the conanfile.py ``build()`` method.
 
+.. _building_with_cmake:
+
 Building with CMake
 ___________________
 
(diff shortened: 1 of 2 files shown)


conanfile_py.rst
```diff
@@ -88,13 +90,13 @@ If you are building your project with CMake, edit your ``conanfile.py`` and add
    
       def build(self):
          cmake = CMake(self.settings)
-         self.run('cmake "%s" %s' % (self.conanfile_directory, cmake.command_line))
-         self.run('cmake --build . %s' % cmake.build_config)
+         cmake.configure(self)
```
lasote: At least at the end of this section I would like to mention that `configure` and `build` are only helpers and you can still use the "command line version".

lasote:
Very good job @sixten-hilborn! I did just an observation but it can be merged.

conanfile_py.rst
sixten-hilborn: Sure! Agree the "command line version" should be mentioned, it's still the one currently used by most packages.
I can do an update maybe this evening or tomorrow.

memsharded:
The docs are very good, so they can just go live today (I changed the PR to master, develop will not go live until next release). If you can, just add a few lines at the beginning explaining the old command-line mode, before introducing the improved helper. If not, we will add them, dont worry. Thanks very much, great job!

PR #161, opened 2017-02-28
docs for build_id & multi-config

memsharded:


packaging/package_info.rst
```diff
@@ -0,0 +1,190 @@
+Package identifiers and information
+====================================
+
+Packages provide two types of information for consumers. The first one would be the package
+identifiers (ID), which is a SHA-1 hash of the package configuration (settings, options, requirements),
+that allows consumers to reuse an existing package without building it again from sources.
+
+The other type of information would be C/C++ build information, as include paths, library names, or
+compile flags
```
lasote: It looks confusing to me. I wouldn't say that a package identifier is an information for me, the consumer.

packaging/package_info.rst
```diff
@@ -0,0 +1,190 @@
...
+        cmake = CMake(self.settings)
+        if cmake.is_multi_configuration:
+            cmd = 'cmake "%s" %s' % (self.conanfile_directory, cmake.command_line)
+            self.run(cmd)
+            self.run("cmake --build . --config Debug")
+            self.run("cmake --build . --config Release")
+        else:
+            for config in ("Debug", "Release"):
+                self.output.info("Building %s" % config)
+                self.run('cmake "%s" %s -DCMAKE_BUILD_TYPE=%s'
```
lasote: new configure/build helpers?

packaging/package_info.rst
```diff
@@ -0,0 +1,190 @@
...
+that allows consumers to reuse an existing package without building it again from sources.
+
+The other type of information would be C/C++ build information, as include paths, library names, or
+compile flags
+
+
+Package IDs
+------------
+
+package_id
```
lasote: I'm not sure about this file in the docs. Maybe all this should go to conanfile.py page.

lasote: And maybe the conanfile.py page should be splitted in several subpages. It's becoming too big.

packaging/package_info.rst
memsharded: Not really. The helpers use the standard ``cmake.command_line`` which depend on `sèttings.build_type`` in this case not defined. So I am manually overriding the behavior.  Probably it could be wrapped in another cmake helper method.

packaging/package_info.rst
memsharded: Nop, the ``conanfile_py.rst`` if you meant that, is just for package consumers. This is only relevant for package creators, and I think it should go inside the "Creating packages section".

packaging/package_info.rst
memsharded: Well, it is printed in the ``conan info`` command, is useful information for consumers of the package, to know which binary they are actually using for a given configuration. 

But, yes, I agree that the title/description could be improved.

PR #163, opened 2017-03-05
Minor updates to SystemPackageTool documentation

sixten-hilborn:


reference/conanfile.rst
```diff
@@ -762,7 +761,7 @@ SystemPackageTool methods:
 - **update()**: Updates the system package manager database.
```
memsharded: Probably clarify that update() is already automatically called before first "install"

memsharded:
Many thanks!

PR #173, opened 2017-03-16
Small but very confusing typo

tridius:
Attribute incorrectly set as "export_sources" instead of "exports_sources". Tripped me up while copying and pasting.


memsharded:
good catch, thanks for fixing it!

memsharded:
good catch, thanks for fixing it!

PR #185, opened 2017-04-01
adding tools.chdir mention to basic example

digitalist:
Had some acute butt hurts trying to adapt basic example from tutorial, until I noticed this "cwd" thing - and found tools.chdir. Guess this can help some newbie like me

### DIFF
--- a/packaging/getting_started.rst
+++ b/packaging/getting_started.rst
@@ -48,7 +48,13 @@ Let`s have a look to the root package recipe ``conanfile.py``:
 
         def source(self):
             self.run("git clone https://github.com/memsharded/hello.git")
-            self.run("cd hello && git checkout static_shared")
+            
+            # self.run("cd hello && git checkout static_shared") 
+            # or better, change working directory and launch multiple commands:
+            with tools.chdir("hello"):
+                self.run("git checkout static_shared")
+                # ...maybe init submodules:
+                # self.run("git submodule update --init")
             # patch to ensure compatibility
             tools.replace_in_file("hello/CMakeLists.txt", "PROJECT(MyHello)", '''PROJECT(MyHello)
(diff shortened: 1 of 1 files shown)


memsharded:
Hi @digitalist 

Thanks very much for your contribution. It didn't look bad, but there is some reason we don't introduce the ``tools.chdir`` directly in the getting started. It is a bit more obvious to most users what the ``cd hello &&`` command is doing, than the equivalent python context trick. I think the best could be to point it out in a comment, or maybe link to the ``tools.chdir`` docs page.

Why did you close your PR? It seems that you used a different email in your commit (your local gitconfig) that doesn't match your github account. Was that the reason that the CLA is not validated?

digitalist:
Hi, thanks for your reply, I'll explain... (screams with a very low, quiet voice) 
It's very obvious, just until that moment you try to "_Aha.. let's take this example and try something more advanced!_" and then you bump into some CWD-related error, but you don't know it's CWD-related and you just start checking your scripts, re-reading docs, etc., etc,. until you decide to debug...! 

`tools.chdir` is lot more obvious because it does not hide this subtle but annoying detail.  In a siimple case `cd && `  is enough  and that's why I left  this as an option in a one-liner comment.  
I'm long-time (>7 years) python user and >5 years desktop (no jokes here) Linux user, so, I guess, if this thing wasn't obvious to me, maybe it can scare other users.

I closed my PR because github hates me... err, no, there was some silly stupid error: I accepted CLA, tests passed, and then github starts barking at me - "We don't know your email, please, add your email... " - though  it's a same email as in other commits and PRs. If you'll be so kind, I can recreate this commit, but I don't want to bother you with debugging a github bugs :-)

memsharded:
Perfect, we will manage! Just created an issue for it: https://github.com/conan-io/docs/issues/186

PR #191, opened 2017-04-11
docs for build_requires

memsharded:
I think it would be good to add some "build-tool", like cmake example, also links to Android, etc.
Do you have cmake examples? Otherwise I will build them myself.

### DIFF
--- a/mastering.rst
+++ b/mastering.rst
@@ -11,6 +11,7 @@ This section provides an introduction to important productivity features and use
    mastering/conanfile_py
    mastering/profiles
    mastering/build
+   mastering/build_requires
    mastering/conditional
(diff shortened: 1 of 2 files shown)


lasote:
This cmake-installer should work well as a build_require, despite of being created as a regular requirements it should work: https://github.com/lasote/conan-cmake-installer/blob/master/conanfile.py

The same for the mingw installer, In fact, for the next release, I would want to take a look to the conan-package-tools and change the way it runs mingw builds, also there is one or two open issues.

https://github.com/lasote/conan-mingw-installer/blob/master/conanfile.py

Please, try it and make a PR to the recipes is something is not working well.

PR #235, opened 2017-06-19
Remove python-fasteners from AUR dependencies

tivek:
python-fasteners was moved from AUR to the official Arch Linux community repo. It can now be fetched automatically by pacman and makepkg.


tivek:
It seems python-monotonic was also moved to community. This makes the manual installation even simpler. Updated patch to reflect that.

memsharded:
Thanks very much for keeping this updated!!

Issue #233, opened 2017-06-19
Document Gitlab CI integration


memsharded:
Developed by @uilianries in https://github.com/conan-io/conan/pull/1401

uilianries:
Okay, I can update docs tonight :smile:

memsharded:
Good, you are an awesome contributor to the project! 
Please if so, do it to the "develop" branch, that we merge to master and release when 0.25 is out.

Thanks again once more for your incredible job.

PR #237, opened 2017-06-20
#233 Update Gitlab integration

uilianries:
This PR includes:

* Updated `conan new` reference to support Gitlab CI.
* Added step-by-step how to create a new Conan project in Gitlab.

### DIFF
--- a/reference/commands/new.rst
+++ b/reference/commands/new.rst
@@ -7,7 +7,7 @@ conan new
 .. code-block:: bash
 
    $ conan new [-h] [-t] [-i] [-c] [-s] [-b] [-cis] [-cilg] [-cilc] [-cio]
-               [-ciw] [-gi] [-ciu CI_UPLOAD_URL]
+               [-ciw] [-ciglg] [-ciglc] [-gi] [-ciu CI_UPLOAD_URL]
                name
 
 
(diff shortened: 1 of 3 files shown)


reference/commands/new.rst
```diff
@@ -64,3 +66,8 @@ package testing files.
 
    $ conan new mypackage/1.0@myuser/stable -t -cilg -cio -ciw
 
+- Crete files for gitlab (linux) Continuous integration and set upload conan server:
```
memsharded: Crete => create

uilianries:
@memsharded Done! I'm using spell checking in my text editor, it warned me about this typo but I didn't see. Anyway, we could add some spell checker to analyzes each new push, should be faster to find typo erros. What do you think?

memsharded:
Sounds good, but what do you mean? Some tool to run in travis? Will it reject PR if errors are encountered? Because the false positives with so many acronyms, brands, etc, are very common.

uilianries:
Yes, run in Travis some tool to check.  
So, we will have a new problem, as you mentioned, false positives. I think in 2 solutions, reserved words, as brands could be add in a suppress file, as done in clang-tidy and cppcheck.  
Or, use `allow_failures` in Travis and just check (user review) when some error was detected.

memsharded:
Personally, it sounds a bit of overhead right now, to setup and run it. The truth is that right now there is no travis CI for the docs, it is directly run in readthedocs. I am mostly fine with a few typos that will eventually be detected and fixed. But just my personal opinion.

uilianries:
Yes, maybe just docs review is enough for now, I agree. I think that to maintain a white list for each new commit or review Travis warnings will be more effort than necessary, maybe will hide real typo-errors. Also, develop this feature will spend some hours to create travis script. Anyway, thanks to discuss the idea.

PR #240, opened 2017-06-28
Add documentation on ``cpp_info`` attribute

cleeland:
This is duplicated from the ``package_info()`` documentation in methods.rst.   DRY would suggest that it should exist in only one place and the other link to that place.  Since it's an attribute, this seems like the more logical place to put the authoritative documentation.

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -439,6 +439,60 @@ In the package recipe methods, some attributes pointing to the relevant folders 
 
 When executing local conan commands (for a package not in the local cache, but in user folder), those fields would be pointing to the corresponding local user folder.
 
+cpp_info
+--------
+
+Typically set in ``package_info()``, the ``cpp_info`` attribute has the following properties you can assign/append to:
+
+.. code-block:: python
+
+   self.cpp_info.includedirs = ['include']  # Ordered list of include paths
+   self.cpp_info.libs = []  # The libs to link against
+   self.cpp_info.libdirs = ['lib']  # Directories where libraries can be found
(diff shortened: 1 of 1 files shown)


memsharded:
Hi Chris,

Thanks for contributing this PR to the docs. The truth about ``self.cpp_info`` is that is not a class-wide attribute, it only makes sense within the ``package_info()`` method, this is why it is documented in that method. 

Same happens to other elements of conanfile, like the ``self.copy`` method. While it seems that it is a class method, it is not. It is a callable object that can be different (actually it is different) in different scopes (package, imports). This is the reason that ``self.copy`` is not documented as a separate method, but only inside of the respective ``package()`` and ``imports()`` methods.

So, as ``cpp_info`` is actually None (check the declaration in the source code for conans/model/conan_file.py constructor) in all other methods of the recipe, using it from elsewhere in the recipe other than ``package_info()`` will make it crash badly, so IMO it is much better and induce less errors to keep it documented just in ``package_info()`` and not as a class attribute.

Thanks very much in any case for your feedback and contributing!

cleeland:
Ah.  That makes sense from a developer's perspective.  However, since `cpp_info` "looks" like an attribute to a user of conan, I would suggest adding something like the following to the attributes documentation in lieu of what I suggested:

> Only valid inside `package_info()`.  See _link to cpp_info documentation under package_info docs_ for more information.

memsharded:
Yes, totally make sense, much better, just added it: http://docs.conan.io/en/latest/reference/conanfile/attributes.html#cpp-info Thanks for the suggestion.

PR #250, opened 2017-07-05
Fixed self.copy() description as per code

shreyasbharath:


reference/conanfile/methods.rst
```diff
@@ -107,7 +107,7 @@ The syntax of copy is as follows:
 
 .. code-block:: python
 
-   self.copy(pattern, dst, src, keep_path=False, symlinks=None, excludes=None)
+   self.copy(pattern, dst, src, keep_path=True, links=False, symlinks=None, excludes=None, ignore_case=False)
```
memsharded: Actually the ``links`` argument is kept in code for backwards compatibility, but it is confusing from the UI. If symlinks is defined, links will take its value. Maybe it could be removed already, but I am always afraid to break users.

shreyasbharath: Ah okay, should I remove it here then?

memsharded: Yes, that is right, there was a mistake in the docs, keep_path default is True. But better not show ``links`` it might eventually be removed. Thanks!

shreyasbharath: Done.

memsharded:
Great, thanks!

Issue #254, opened 2017-07-17
Document conans.util.file


memsharded:
save, load, mkdir, etc

MarcoKoch:
I think, with #255 merged this can be closed then. Or is there anything else missing?

PR #255, opened 2017-07-18
Document new public file utils

MarcoKoch:
This adds documentation for the functions added to `conans.tools` in conan-io/conan#1514.
See #254.


memsharded:
It is so great when someone contributes both to the code and the documentation. Thanks so much! :)

MarcoKoch:
Code is only as good as its documentation ;)

PR #261, opened 2017-07-19
Update conanfile_txt.rst 'requires' description

dennisfen:
Make description of 'requires' section consistent with example: package owner should be 'pocoproject' instead of 'lasote'.


memsharded:
Thanks for fixing this!

PR #268, opened 2017-07-28
Fixed typo

ovidiub13:


memsharded:
Merged to master too, so it goes live (for fixes that are in the live docs, PR can go to master, so they are published directly).

Thanks for fixing this!

PR #280, opened 2017-08-17
updated definitions to defs

solvingj:
I changed to match line 118, which i think is the new variable name, not sure.


memsharded:
Yes, in the code you can see it, it is called ``definitions`` now. Thanks!

PR #310, opened 2017-08-30
Feature/runtime detection

### LINKED ISSUE #300: 0.26: compiler.runtime not enforced

lasote:
Close #300


integrations/cmake/cmake_multi_generator.rst
```diff
@@ -29,9 +29,9 @@ switch among configurations:
 
 .. code-block:: bash
 
-    $ conan install -s build_type=Release -s compiler.runtime=MD ...
```
lasote: I've removed this because it was not even possible to specify the runtime without specifying first the compiler, right? so, better to remove it.

PR #325, opened 2017-09-20
0.27

lasote:


reference/commands/config.rst
```diff
@@ -35,3 +39,65 @@ Manages conan.conf information.
...
+The **profiles files**, that will be overwritten if already present, but won't delete any other profile file that the user
+has in the local machine.
+
+
+Allowed files in the zipfile:
+
++--------------------------------+----------------------------------------------------------------------+
+| File                           |                                                                      |
++================================+======================================================================+
+| pylintrc                       | Overrides the local ~/.conan/pylintrc                                |
```
memsharded: Local pylintrc isn't a thing. It is just another location to be pointed by a conf variable if desired, but it is not a default, nor it is loaded automatically.

reference/commands/config.rst
```diff
@@ -35,3 +39,65 @@ Manages conan.conf information.
...
+
+The zip can contain only a subset of all the allowed configuration files, only the present files will be
+replaced, except the **conan.conf** file, that will apply only the declared variables in the zipped ``conan.conf`` file
+and will keep the rest of the local variables.
+
+The **profiles files**, that will be overwritten if already present, but won't delete any other profile file that the user
+has in the local machine.
+
+
+Allowed files in the zipfile:
```
memsharded: All files are allowed in the zipfile, not only these. They will be copied verbatim if not matching the special ones.

reference/commands/config.rst
```diff
@@ -35,3 +39,65 @@ Manages conan.conf information.
...
+
+**Examples**:
+
+- Install the configuration from an url:
+
+.. code-block:: bash
+
+    $ conan config install http://url/to/some/config.zip
+
+- Refresh the configuration again:
```
memsharded: Say that conan.conf stores the URL to apply it

reference/config_files/settings.yml.rst
```diff
@@ -11,6 +11,9 @@ that commonly make sense to most users. Anyway, you can add/remove/modify those
 possible values in the ``settings.yml`` file, according to your needs, just be sure to share changes with
 colleagues or consumers of your packages.
 
+If you want to distribute a unified the ``settings.yml`` file
```
memsharded: not finished

reference/commands/config.rst
lasote: fixed

reference/commands/config.rst
lasote: fixed

reference/commands/config.rst
lasote: fixed

reference/config_files/settings.yml.rst
lasote: fixed

PR #326, opened 2017-09-21
add tools.collect_libs()

ovidiub13:


reference/tools.rst
```diff
@@ -513,3 +513,20 @@ Replaces the ``prefix`` variable in a package config file ``.pc`` with the speci
...
+
+- **conanfile**: A `ConanFile` object from which to get the `package_folder`
+- **folder**: The subfolder where the library files are. Defaulted to "lib"
+
+**Example:**
+
+.. code-block:: python
+
+     def package_info(self):
+        self.cpp_info.libs = tools.collect_libs()
```
lasote: The example should use the "self"

ovidiub13: but this was moved to the tools module since 0.27.

ovidiub13: https://github.com/conan-io/conan/blob/develop/conans/client/tools/files.py#L207

ovidiub13: http://conanio.readthedocs.io/en/latest/changelog.html?highlight=collect_libs#september-2017

ovidiub13: Sorry, I misunderstood. I've added it now.

PR #333, opened 2017-10-02
Added clarity to CONAN_SYSREQUIRES_SUDO

solvingj:
I don't know if my changes are correct, so please make any changes necessary.  It seemed that what was documented was probably not clear or sufficient, so I figured this would get it improved one way or another.


memsharded:
It is good, thanks!

PR #334, opened 2017-10-02
Added note about Win10 long paths feature only working with Python3.6

sztomi:
After some experimentation and research I found that earlier Python versions can't take advantage of the long paths feature in newer Windows 10 builds. First version to support it is Python 3.6. I think it's important to note this in the docs.

https://stackoverflow.com/questions/46488118/win32-long-paths-from-python/46489473#46489473

https://docs.python.org/3/using/windows.html#removing-the-max-path-limitation


memsharded:
You are right, thank you

PR #336, opened 2017-10-04
minor formatting fix

gesellc:
Missing blank leads to wrong inline code highlighting


memsharded:
Thanks for fixing this! 
It seems that you might have signed the commit with a different email than the one used in github. Coud you please check the above CLA notice, and how to fix it? Thanks!

PR #371, opened 2017-11-06
Added example Apache configuration for using Conan installed from pip.

Lyberta:
Added example configuration for running Conan server installed from pip behind Apache as discussed in conan-io/conan#1986


memsharded:
Excellent, many thanks!

PR #387, opened 2017-11-29
Removes .txt extension in profiles

danimtb:


memsharded:
I think it is a bit more elegant the profiles without extensions

PR #399, opened 2017-12-10
meson full package example

memsharded:


reference/build_helpers/meson.rst
```diff
@@ -35,3 +35,71 @@ Methods
...
+        name = "Hello"
+        version = "0.1"
+        settings = "os", "compiler", "build_type", "arch"
+        generators = "pkg_config"
+        exports_sources = "src/*"
+
+        def build(self):
+            meson = Meson(self)
+            meson.configure(source_dir="%s/src" % self.source_folder, 
+                            build_dir="build")
```
lasote: I prefer the explicit use of `self.build_folder`

reference/build_helpers/meson.rst
```diff
@@ -35,3 +35,71 @@ Methods
     - **args**: A list of additional arguments to be passed to the ``make`` command. Each argument will be escaped according to the current shell. No extra arguments will be added if ``args=None``
     - **build_dir**: Default conanfile.build_folder
     - **targets**: A list of targets to be built.
```
lasote: In the example at the top of the page, change the `meson.configure()` with a `meson.configure(build_dir="%s/build" % self.build_folder)`

lasote:
I've changed to develop the PR. I think it's better to fix it and then release the docs.

PR #424, opened 2017-12-14
Formatted reference/environment variables

danimtb:
- Added deafult values (when possible)
- Set examples in code-blocks
- Added `See also`, `Note` and `Warning` text boxes
- Changed headers highlight

### DIFF
--- a/reference/env_vars.rst
+++ b/reference/env_vars.rst
@@ -1,15 +1,18 @@
 .. _env_vars:
 
 Environment variables
-=============================
+=====================
 
 These are the environment variables used to customize conan.
 
-Most of them can be set in the ``conan.conf`` configuration file (inside your ``<userhome>/.conan`` folder)
+Most of them can be set in the ``conan.conf`` configuration file (inside your ``<userhome>/.conan`` folder).
 
+Variables
+---------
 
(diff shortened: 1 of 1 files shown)


reference/env_vars.rst
```diff
@@ -1,15 +1,18 @@
...
 Environment variables
-=============================
+=====================
 
 These are the environment variables used to customize conan.
 
-Most of them can be set in the ``conan.conf`` configuration file (inside your ``<userhome>/.conan`` folder)
+Most of them can be set in the ``conan.conf`` configuration file (inside your ``<userhome>/.conan`` folder).
 
+Variables
```
memsharded: Why this new level? Seems uninformative.

danimtb: Just to get the headers highlight right (There is a CMake Related Variables after at the same highlight level).

Any suggestion for a more informative name?

memsharded: I think I prefer the current levels, that is too much of nesting, and I think it might put the last variables like CONAN_USER_HOME_SHORT, etc under ``CMake Related Variables``

danimtb: Oh Ok, now I see it.

PR #426, opened 2017-12-14
Format autotools build helper

danimtb:
- Added signature to methods and brief description.
- Added constructor section.
- Formatted attributes section.
- Default values to parameters and attributes (when possible).

### DIFF
--- a/reference/build_helpers/autotools.rst
+++ b/reference/build_helpers/autotools.rst
@@ -44,6 +44,71 @@ Constructor
 Parameters:
     - **conanfile** (Required): Conanfile object. Usually ``self`` in a conanfile.py
 
+Attributes
+----------
+
+You can adjust the automatically filled values modifying the attributes like this:
+
+.. code-block:: python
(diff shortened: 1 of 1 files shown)


reference/build_helpers/autotools.rst
```diff
@@ -1,58 +1,160 @@
...
+            env_build = AutoToolsBuildEnvironment(self)
+            env_build.fpic = True
+            env_build.libs.append("pthread")
+            env_build.defines.append("NEW_DEFINE=23")
+
+            with tools.environment_append(env_build.vars):
+                self.run("./configure")
+                self.run("make")
+
+fpic
```
memsharded: I think for this I prefer the table style. More compact.

danimtb: I also think a table is more compact in this case as there is no much additional information. However, I think it is better to have a uniform style format

lasote: If it is only for the build helpers I think it's ok. Other tables like the integrations section are different.

Issue #443, opened 2017-12-22
Add warning to tools.collect_libs() reference about linking order


danimtb:
Issue https://github.com/conan-io/conan/issues/2201

`tools.collect_libs()` does not follow a predefined dependency graph of the libraries to link with and might not not sort them in a suitable way.

PR #444, opened 2017-12-22
tool.collect_libs warning about order

danimtb:
Issue #443

### DIFF
--- a/reference/tools.rst
+++ b/reference/tools.rst
@@ -826,8 +826,8 @@ libraries or with complex names like ``libmylib-x86-debug-en.lib``.
 
 .. warning::
 
-    This tool collects the libraries searching directly inside the package folder and returns the in
-    no specific order. If libraries are inter-dependent, then package_info() method should order
+    This tool collects the libraries searching directly inside the package folder and returns them
+    in no specific order. If libraries are inter-dependent, then package_info() method should order
     them to achieve correct linking order.
 
 .. _pkgconfigtool:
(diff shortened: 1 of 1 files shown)


reference/tools.rst
```diff
@@ -810,7 +810,8 @@ tools.collect_libs()
 
     def collect_libs(conanfile, folder="lib")
 
-Fetches a list of all libraries in the package folder.
+Fetches a list of all libraries in the package folder. Useful to collect libraries with complex
```
memsharded: Not only. Also useful if the libraries are not inter-dependent.

reference/tools.rst
```diff
@@ -823,6 +824,13 @@ Fetches a list of all libraries in the package folder.
     - **conanfile** (Required): A `ConanFile` object from which to get the `package_folder`.
     - **folder** (Optional, Defaulted to ``"lib"``): The subfolder where the library files are.
 
+.. warning::
+
+    This tool collects the libraries searching directly inside the directory tree of the package
+    folder. It does not follow any kind of dependency graph, name convention or build order, so it
```
memsharded: => This tool collects the libraries searching directly inside the package folder, and returns them in no specific order. If the libraries are inter-dependent, then the ``package_info()`` method should order them to achieve correct linking order.

reference/tools.rst
```diff
@@ -823,6 +824,12 @@ Fetches a list of all libraries in the package folder.
     - **conanfile** (Required): A `ConanFile` object from which to get the `package_folder`.
     - **folder** (Optional, Defaulted to ``"lib"``): The subfolder where the library files are.
 
+.. warning::
+
+    This tool collects the libraries searching directly inside the package folder and returns the in
```
memsharded: and returns **them**

PR #449, opened 2017-12-29
Fix warnings in bash code blocks

danimtb:
**Do not use ' in bash code blocks**

Fixes warning generarintg docs: `WARNING: Could not lex literal_block as "bash". Highlighting skipped.`

### DIFF
--- a/reference/commands/creator/create.rst
+++ b/reference/commands/creator/create.rst
@@ -68,9 +68,9 @@ created correctly. Check the 'conan test' command to know more about the
                             binary is not built with the current recipe or when
                             missing binary package. --build=[pattern] Build always
                             these packages from source, but never build the
-                            others. Allows multiple --build parameters. 'pattern'
+                            others. Allows multiple --build parameters. "pattern"
                             is a fnmatch file pattern of a package name. Default
-                            behavior: If you don't specify anything, it will be
+                            behavior: If you dont specify anything, it will be
                             similar to --build=never, but package recipes can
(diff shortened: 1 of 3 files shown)


reference/commands/creator/create.rst
```diff
@@ -68,9 +68,9 @@ created correctly. Check the 'conan test' command to know more about the
                             binary is not built with the current recipe or when
                             missing binary package. --build=[pattern] Build always
                             these packages from source, but never build the
-                            others. Allows multiple --build parameters. 'pattern'
+                            others. Allows multiple --build parameters. "pattern"
                             is a fnmatch file pattern of a package name. Default
-                            behavior: If you don't specify anything, it will be
```
lasote: Should it be a way to escape it?

memsharded: Backslashes should make it: http://docutils.sourceforge.net/docs/user/rst/quickref.html#escaping

danimtb: For command line help we should use ``.. code-block:: console``. Howeever the highligth in console mode is too plain and does not show the same as bash. I found a similiar issue here: https://github.com/sphinx-doc/sphinx/issues/4098

Apparently, bash code does not allow ' quotes, that is why we are seeing the warning message all the time. 

Also skipping the ' quotes inside a code-block does not make it work

memsharded: Oh, yes, that is true. I am fine with this merge without the quotes.

PR #438, opened 2017-12-21
Add a Repositories page

lamefun:
Conan docs need to link to active repositories, or else people will look at the official repositories and see only 21 active packages, and will rightly assume that the community has abandoned Conan and it's not long for this world. For now I added bincrafters. Does anyone know any other active repositories?

PR #453, opened 2018-01-02
Docs introduction refactor

danimtb:
Includes commit from PR #438 

1. Getting started:
    - [x]  End with intro to profiles, link with reference
2. Using packages:
    - [ ] Repeat using profiles
3. Creating packages:
    - [x] Remove: Uploading, bintray, running & deploying
    - [x] Add: ABI compatibility & package_id() stuff from “how-to”
    - [ ] Add: Local flow
4. Uploading packages:
    - [x] Server: conan-server , Artifactory, etc
    - [x] Popular repos: bincrafters
    - [x] Conan-server (From “running your server”)
    - [x] Bintray (from “creating”)

**Not done:**
  - Add: Local flow -> I think creating packages has a good understandable flow as it is now. WDYT?
  - Repeat using profiles  -> Well explained in getting started and didnt find a suitable connection inside 
    *Using Packages* section


creating_packages.rst
```diff
@@ -0,0 +1,18 @@
+.. _packaging:
```
memsharded: I suggest leaving this section named as "packaging", so it is easier to review. Otherwise, it seems that everything is new, I can't see changes. This rename can be done in a further PR, or leave it, the name "packaging" is not that bad.

creating_packages/package_repo.rst
```diff
@@ -3,26 +3,29 @@
 Recipe and sources in the same repo
 ===================================
 
-In the previous package we implemented a ``source()`` method that fetched the source code from github. An alternative approach would be embedding the source code into the package recipe, so it is self-contained and it doesn't require to fetch code from external origins when it is necessary to build from sources.
+In the previous package we implemented a ``source()`` method that fetched the source code from
```
memsharded: Also, maybe leave format (like making multiline) changes to the other PR, as the review forces to re-read everything and compare manually with the old text

creating_packages/existing_binaries.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Sometimes, it is necessary to create packages from existing binaries, like binaries from third
+parties, or previously built by another process or team not using conan, so building from sources is
+not wanted. You would want to package local files in two situations:
+
+ - When it is not possible to build the packages from sources (only pre-built binaries available).
+ - When you are developing your package locally and want to export the built artifacts to the local
+   cache.
+   As you don't want to rebuild again (clean copy) your artifacts, you don't want to call
+   **$ conan create**.
```
memsharded: Regarding format, I am not sure the bold is the best for command line examples.  I am fine with using a different one, like italics for files/folders.

danimtb:
These are the possibilities to make emphasis in text: http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#emphasis

I tried to follow this rules:
- Bold text (strong emphasis) to highlight important texts
- Italics (emphasis) to refer to file names, paths or dirs.
- Bold with $ to refer to command line commands. e.i. **$ conan create** or **
- Inline literals (``like this``) to refer to code or text inside recipies. e.i. ``package()``, ``cmake`` (generator in *conanfile.txt*) or ``settings`` (variable inside a ``ConanFile`` class in *conanfile.py*)
- Other names like CMake, Autotools, Conan, Visual Studio should not be emphasized, just use a capital letter or follow the convention (like first and second capitalized letters for CMake)

creating_packages/existing_binaries.rst
danimtb: yes, not sure about bold for command line commands but I couldn't find other option, that's why I used the $

danimtb:
An yes, it is a refactor. Moving some file from one place to another an moving parts of texts to other files

creating_packages/package_repo.rst
danimtb: Many work to revert here. I think this is a good opprotunity to make some formatting. Otherwise it would never be done. I agree it makes it harder to review though...

memsharded: No problem, do not revert it. Just for next time

memsharded:
Agree with all formatting but the **$ conan create**. Because when they are used in code-blocks they are rendered as inline, so I think it reads better in the same font and style. What do you think @lasote?

creating_packages.rst
danimtb: Packaging is not a bad name. However I think it is easier and intuitive to find files that have same name as the header. If a section is called "Creating packages" I would expect the content to be in a file with a similar name.

memsharded: Yes, I was thinking also of so many broken links. Everytime we link a docs page from github pointing to that "packaging" section, will be broken. But I am fine with the change too, no big issue

danimtb:
Maybe ``$ conan create ..`` to match code-bloc style and differenctiate from code or text like ``default_options``?

memsharded:
You mean, using alwas the ``$`` to differentiate it is command line? Yes, could be.

PR #456, opened 2018-01-04
Local development flow

lasote:
- Almost all contents from bincrafters blogpost


danimtb:
It will need some formatting, especially the extracted part, but same applies to whole docs. We could do it in a separate PR

lasote:
I agree with both of you. Let's merge it and I'll open an issue, so if we have more time we can improve it.

PR #460, opened 2018-01-04
Profiles in using packages section

danimtb:
Removed profiles from mastering and splitted themin using_packages and reference


using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+    CC=/usr/bin/clang
+    CXX=/usr/bin/clang++
+
+And you can use it instead of command line arguments as:
+
+.. code-block:: bash
+
+    $ conan create demo/testing -pr=myprofile
+
+You can use ``$PROFILE_DIR`` in your profile and it will be replaced with the absolute path to the profile file.
```
lasote: the PROFILE_DIR only in the reference I would say

using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+
+If we continue with the example of Poco, we could have a handy profile to help us build our project with the desired configuration and avoid
+the ussage of all the command line arguments when installing the dependency packages.
+
+A profile to build our depenencies as **shared** and and in **debug** mode will look like this:
+
+.. code-block:: text
+   :caption: *poco_debug_shared_profile*
+
+
```
lasote: `include(default)`
Without this, the default settings are not present

using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+.. code-block:: text
+
+    [options]
+    *:shared=True
+
+They would contain the desired configuration, for example:
+
+.. code-block:: text
+   :caption: myprofile
```
lasote: In this first example, I wouldn't introduce options, only the env and settings stuff and I would name it "clang_3.5" instead of myprofile, to be reused later

lasote: Explain where to save the file

using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: text
+
+   [env]
+   PYTHONPATH=$PROFILE_DIR/my_python_tools
+
+.. note::
+
+    If you specify a profile in a conan command, like `conan create` or `conan install` the base profile ``~/.conan/profiles/default`` won't
+    be applied.
```
lasote: From the comment above to here, I would remove it in favor of the reference

using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+
+.. code-block:: bash
+
+    $ conan install . -pr=poco_debug_shared_profile
+
+We could also create a new profile to use a different compiler in a different OS, even use release as build type and store it in our project
+directory:
+
+.. code-block:: text
+   :caption: *poco_apple_clang*
```
lasote: In this example, I would include the previous `clang_3.5` and introduce, in addition to the *:shared=True, a specific option for Poco, like `Poco:enable_apacheconnector=False`. The [settings] section wouldn't be necessary anymore

using_packages/using_profiles.rst
```diff
@@ -0,0 +1,121 @@
...
+    Poco:shared=True
+    OpenSSL:shared=True
+
+With this we could just install using the profile:
+
+.. code-block:: bash
+
+    $ conan install . -pr=poco_debug_shared_profile
+
+We could also create a new profile to use a different compiler in a different OS, even use release as build type and store it in our project
```
lasote: the `in a different OS` may be complex. I would change only the compiler for this example.

reference/profiles.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    [build_requires]
+    Tool1/0.1@user/channel
+    Tool2/0.1@user/channel, Tool3/0.1@user/channel
+    *: Tool4/0.1@user/channel
+
+Profile files are useful to gather some configurations that would have to be introduced as arguments in command line instructions like
+``$ conan install`` or ``$ conan create``. Therefore, profile files ussage is highly recommended as anyone could share the whole
+configuration to build a project with settings and options, set set environment variables needed for compilation and even set requirements
+for building.
```
lasote: Less literature, reference section. Something like:

Use ``--profile`` with ``conan install`` and ``conan create`` commands to apply the profile

reference/profiles.rst
```diff
@@ -0,0 +1,243 @@
...
+    compiler=clang
+    compiler.version=3.5
+    compiler.libcxx=libstdc++11
+
+    [options]
+    MyLib:shared=True
+
+    [env]
+    CC=/usr/bin/clang
+    CXX=/usr/bin/clang++
```
lasote: The only new things in this example are the use of CC, CXX, can be mentioned to the way to override the default compiler in our system if we have more than one, also for pointing to a toolchain when we are cross-building.

reference/profiles.rst
```diff
@@ -0,0 +1,243 @@
...
+Profiles can be located in different folders, for example, the default ``<userhome>/.conan/profiles``, and be referenced by absolute or
+relative path:
+
+.. code-block:: bash
+
+    $ conan install --profile /abs/path/to/profile  # abs path
+    $ conan install --profile ./relpath/to/profile  # resolved to current dir
+    $ conan install --profile profile  # resolved to user/.conan/profiles/profile
+
+You can list and show existing profiles with the ``$ conan profile`` command:
```
lasote: More bullet and short: "Listing existing profiles:" "Showing a profile"

reference/profiles.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    [env]
+    PYTHONPATH=$PROFILE_DIR/my_python_tools
+
+.. note::
+
+    If you specify a profile in a conan command, like `conan create` or `conan install` the base profile ``~/.conan/profiles/default`` won't
+    be applied. If you want to apply it use the ``include`` directive explained later in this page.
+
+Package settings and env vars
```
lasote: Remove all this, I think for the reference is enough to say that the profiles can be in the default profile page at ~/.conan/profiles or in any other folder. Maybe reuse this example to explain how to use a profile to point to a different compiler.

PR #464, opened 2018-01-05
Android studio with profile, working

lasote:


integrations/android_studio.rst
```diff
@@ -14,153 +14,104 @@ that can be cross-compiled to Android could be used using the same procedure.
...
-|wizard7|
-
 
 **conanfile.txt**
 
 .. code-block:: text
 
     [requires]
-    libpng/1.6.23@lasote/testing
+    libpng/1.6.23@conan/stable
```
memsharded: But this still not in conan-center, right?

lasote: You are right! Maybe we should add it

lasote: better to wait. there is a pending inclusion request for libpng that seems to be more up to date, but the GH link is broken, so, I'll keep the transit reference

PR #466, opened 2018-01-08
makefiles howto

memsharded:


lasote:
Add a mention of `configure()` method, very common when the makefile is generated with the autotools.

PR #479, opened 2018-01-15
Feature/automatic pythonpath

lasote:


danimtb:
can this be merged @lasote ?

lasote:
If you like it... LoL

reference/tools.rst
```diff
@@ -478,13 +478,18 @@ Parameters:
...
     def pythonpath(conanfile)
 
 This is a context manager that allows to load the PYTHONPATH for dependent packages, create packages
 with python code, and reuse that code into your own recipes.
 
+It is automatically applied
+
 .. code-block:: python
 
     from conans import tools
```
danimtb: If I understood this feature well, using `tools.pythonpath()` is no longer needed if `apply_env `is `True`. So this code

```
def build(self):
    with tools.pythonpath(self):
        from module_name import whatever
        whatever.do_something()
```
can still be used even if `apply_env` is `True` and would be the same as typing:

```
def build(self):
    from module_name import whatever
    whatever.do_something()
```
is that right?

lasote: True. Do you want me to add these comments?

danimtb:
It would be nice to have them, yes! :D

2018-02-13 9:46 GMT+01:00 Luis Martinez de Bartolome Izquierdo <
notifications@github.com>:

lasote:
Done

PR #484, opened 2018-01-16
Update package_repo.rst

menuet:


danimtb:
Same here! Thanks again

PR #493, opened 2018-01-23
Cross build profiles

lasote:
Real and generic profiles to target Linux/android

### DIFF
--- a/systems_cross_building/cross_building.rst
+++ b/systems_cross_building/cross_building.rst
@@ -146,8 +146,8 @@ Linux to Windows
 A **bin/example.exe** for Win64 platform has been built.
 
 
-Windows to Raspberry PI
-.......................
+Windows to Raspberry PI (Linux/ARM)
+...................................
 
 - Install the toolchain: http://gnutoolchains.com/raspberry/
   You can choose different versions of the GCC cross compiler, choose one and adjust the following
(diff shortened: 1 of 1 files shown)


systems_cross_building/cross_building.rst
```diff
@@ -157,22 +157,41 @@ Windows to Raspberry PI
...
+    standalone_toolchain=C:/sysgcc/raspberry
+    cc_compiler=gcc
+    cxx_compiler=g++
+
     [settings]
     os_build=Windows
     arch_build=x86_64
     os=Linux
-    arch=armv7hf
+    arch=armv7
```
memsharded: Actually, this would be armv6 for older RPI (maybe a note somewhere?)

systems_cross_building/cross_building.rst
```diff
@@ -157,22 +157,41 @@ Windows to Raspberry PI
...
-    compiler.version=6.3  # We have chosen gcc 6.3 installer
+    compiler.version=6
     compiler.libcxx=libstdc++11
     build_type=Release
 
     [env]
-    CC=arm-linux-gnueabihf-gcc
-    CXX=arm-linux-gnueabihf-g++
+    CONAN_CMAKE_FIND_ROOT_PATH=$standalone_toolchain/$target_host/sysroot
+    PATH=[$standalone_toolchain/bin]
```
memsharded: If PATH is defined and most tools (AR, LD, STRIP) are in that path, is it always necessary to define the below env-vars?

lasote: No, but it is harmless and covers all cases.

systems_cross_building/cross_building.rst
lasote: Right. Rpi 1 is armv6, it has to be updated in this table: http://docs.conan.io/en/latest/systems_cross_building/cross_building.html#arm-architecture-reference

PR #514, opened 2018-02-05
Completed docs about CONAN_CMAKE_PROGRAM

lasote:
https://github.com/conan-io/conan/issues/2249

### DIFF
--- a/reference/env_vars.rst
+++ b/reference/env_vars.rst
@@ -176,13 +176,16 @@ CONAN_MAKE_PROGRAM
 
 **Defaulted to**: Not defined
 
-Specify an alternative ``make`` program to use with the :ref:`build helper AutoToolsBuildEnvironment<autotools_reference>` make() method. 
+Specify an alternative ``make`` program to use with:
+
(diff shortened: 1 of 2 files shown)


reference/env_vars.rst
```diff
@@ -176,13 +176,16 @@ CONAN_MAKE_PROGRAM
...
+
+    - The build helper :ref:`AutoToolsBuildEnvironment<autotools_reference>`. Will invoke the specified executable in the `make` method.
+    - The build helper :ref:`build helper CMake<cmake_reference>`. By adjusting the CMake variable `CMAKE_MAKE_PROGRAM <https://cmake.org/cmake/help/v3.0/variable/CMAKE_MAKE_PROGRAM.html>`_.
 
 For example:
 
 .. code-block:: bash
 
-    CONAN_MAKE_PROGRAM = ``mingw32-make``.
+    CONAN_MAKE_PROGRAM="/path/to/mingw32-make"
```
memsharded: But specifying it without the path is also valid, right?

lasote: Yes, but it has to be in the path. We can add a comment.

PR #523, opened 2018-02-09
doc: Fix conan requires converted to mailto link

liberforce:
The @ sign in "Poco/1.8.0.1@pocoproject/stable" was detected as being part of an email address and incorrectly converted to a mailto link.


memsharded:
Good catch, thanks!

PR #530, opened 2018-02-13
Avoid dangerous rm -rf

liberforce:
`rm -rf *` may be dangerous if the `cd` before failed.
`rm -rf example-hello-build/*` might be dangerous if a space is added before the `*`.
Other danger is the user calling that from bash history but from the wrong directory.

A possible improvement would be:
`cd example-hello-build && rm -rf *`

As a bonus, I added a fix for a bash comment.

### DIFF
--- a/using_packages/workflows.rst
+++ b/using_packages/workflows.rst
@@ -43,7 +43,7 @@ Now you are ready to build:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


memsharded:
Yes, you are totally right. Thanks for catching this one, hopefully will avoid some problems to some user.

PR #539, opened 2018-02-16
Added method execution order for conan create

### LINKED ISSUE #535: Inlcude conan create method execution flow

danimtb:
closes #535


danimtb:
I think it should be added to the `$ conan install` command referencem page. Don't you think?

memsharded:
``$ conan install`` is for consumption, you probably are not that concerned about which methods of the recipe are being executed...

But yes, please comment in create that those 5-11 wouldn't be executed and add them to ``install`` reference, for completitude.

PR #543, opened 2018-02-20
keep_path paramenter for copy inside imports()

danimtb:
Documentation for PR conan-io/conan#2495

### DIFF
--- a/reference/conanfile_txt.rst
+++ b/reference/conanfile_txt.rst
@@ -112,17 +112,22 @@ List of files to be imported to a local directory. Read more: :ref:`imports<impo
     bin, *.dll -> ./bin # Copies all dll files from packages bin folder to my local "bin" folder
     lib, *.dylib* -> ./bin # Copies all dylib files from packages lib folder to my local "bin" folder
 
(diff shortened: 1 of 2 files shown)


reference/conanfile_txt.rst
```diff
@@ -114,18 +114,15 @@ List of files to be imported to a local directory. Read more: :ref:`imports<impo
 
 The first item is the subfolder of the packages (could be the root "." one), the second is the pattern to match. Both relate to the conan local cache. The third (after the arrow) item, is the destination folder, living in user space, not in the conan cache.
 
-
-The ``[imports]`` section also support the same arguments as the equivalent ``imports()`` method in conanfile.py, separated with an @
-
-- root_package: fnmatch pattern of the package name ("OpenCV", "Boost") from which files will be copied. Default: all packages in deps
```
memsharded: I think I'd keep here a copy of the arguments, linking conanfile.txt users to conanfile.py could be confusing to them.

danimtb: I think it is harder to maintain this way as explanations need to match in both places. But it is ok, as it is something that is not willing to change in the future.

PR #557, opened 2018-02-28
Clarify Visual Studio integration docs

claremacrae:
This is an attempt to write up the helpful advice given to me in conan-io/docs/issues/556

It would be better if someone could explain how to work around the current cmake limitations - or point to instructions on this.

### DIFF
--- a/integrations/visual_studio.rst
+++ b/integrations/visual_studio.rst
@@ -13,14 +13,20 @@ Conan can be integrated with **Visual Studio** in two different ways:
 With CMake
 ----------
 
+Use this, or **cmake_multi**, if you are using cmake to machine-generate your Visual Studio projects.
+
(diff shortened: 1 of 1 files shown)


memsharded:
Thanks Clare for the contribution!

> could explain how to work around the current cmake limitations

Right now it is a bit of chicken and egg problem. CMake *find-package* limitations don't allow to find both debug and release packages. So you cannot use ``multi`` generators with find_package. And not using ``multi`` generators forces developers to go to command line to install the correct Debug or Release dependencies when the build-type is switched in the IDE.

There is no current workaround to this, with current state. It is necessary to use one or the other approach. I personally prefer the single-conf approach, not using ``multi`` generator and going to the command line to switch installed dependencies. This will be much improved when full integration with IDEs via plugins is implemented.

Issue #558, opened 2018-02-28
Can the text CONAN_PKG::Hello1 be explained in 'cmake_multi generator' docs?


claremacrae:
In this page:
http://docs.conan.io/en/latest/integrations/cmake/cmake_multi_generator.html
(revision https://github.com/conan-io/docs/commit/9f1a0b2b71b457b4d09c9207b8e13c92f87aade0)

... in the "Targets approach" section, there is this example:

```
    project(MyHello)
    cmake_minimum_required(VERSION 2.8.12)

    include(${CMAKE_BINARY_DIR}/conanbuildinfo_multi.cmake)
    conan_basic_setup(TARGETS)

    add_executable(say_hello main.cpp)
    target_link_libraries(say_hello CONAN_PKG::Hello1)
```

What is the `Hello1` string, in `CONAN_PKG::Hello1`?

It doesn't appear anywhere else on the page.

danimtb:
Hi @claremacrae!

Last part of the docs is considered as reference and some things may not be full explained in some sections. I think you are looking for this exlanation http://docs.conan.io/en/latest/reference/generators/cmake.html#targets-generated-by-conanbuildinfo-cmake

claremacrae:
Thank you. Sorry, but I still don’t understand what Hello1 is there - there is no other reference to it on the page.

* The project is called MyHello
* The target is called say_hello
* Hello1 seems undefined

What am I missing?

Also, I noticed that the link says that this mechanism only works with CMake 3 and above, but the code snippet says 2.8.12 is required to use that approach.

danimtb:
Hello1 is just an example of a package name (``CONAN_PKG::PkgName``), so it has nothing to do with the target name nor the project's.

That section is not written as a tutorial, just reference for users. If you think it would be better to have a less confusing name, please submit a PR and we will be glab to merge it :)

Regarding CMake version I think you are right, Targets are supported from CMake 2.8.12

claremacrae:
I've submitted a PR to try to explain Hello1...

PR #561, opened 2018-03-01
Fix CMake minimum version to use targets

danimtb:
Info here: https://cmake.org/cmake/help/v3.0/policy/CMP0023.html

Related to #558


reference/generators/cmake.rst
```diff
@@ -165,7 +165,7 @@ There are other methods automatically called by ``conan_basic_setup()`` but you
 Targets generated by conanbuildinfo.cmake
 -----------------------------------------
 
-If you use ``conan_basic_setup(TARGETS)``, then some cmake targets will be generated (this only works for CMake>3)
+If you use ``conan_basic_setup(TARGETS)``, then some cmake targets will be generated (this only works for CMake > 2.8.12)
```
memsharded: No, the TARGETS approach actually need something like 3.3

danimtb:
See https://github.com/conan-io/conan/blob/38fddb79ee2a4f6229b11d40994675c6f3fef488/conans/client/generators/cmake_common.py#L148

PR #564, opened 2018-03-04
export-pkg --package-folder docs

memsharded:


creating_packages/package_dev_flow.rst
```diff
@@ -148,8 +148,19 @@ This parameters takes the same parameters as ``package()`` but in this case ther
 | ``install-folder`` |                    |
 |                    |                    |
 | ``build-folder``   |                    |
+|                    |                    |
+| ``package-folder`` |                    |
 +--------------------+--------------------+
 
+There are 2 modes of operation:
+
+- Using ``source-folder`` and ``build-folder`` does not require the execution of the above ``$ conan package``.
```
danimtb: Why state that the ``$ conan package`` above is not needed? I mean, I know you don't but I think it is better to encourage users to execute it to make sure an to stick to a the same flow. I'd rather omit this *skip* of ``package()``

creating_packages/package_dev_flow.rst
```diff
@@ -148,8 +148,19 @@ This parameters takes the same parameters as ``package()`` but in this case ther
...
+  It will use the ``package()`` method to extract the artifacts from those folders and create the package,
+  directly in the conan local cache.
+- Using the ``package-folder`` argument (incompatible with the above 2), will not use the ``package()`` method,
```
danimtb: Not sure we should state that here there is a way to "split" the flow. I see this as a fix for ``cmake.install()`` and maybe we could add this information at the bottom of this command or just as a note for those users.

creating_packages/package_dev_flow.rst
memsharded: It is important that users understand that ``export-pkg`` in this case works from the source and build folder, and not from the package folder, and that it will execute the ``package()`` method. Otherwise there can be confusion about what the command is actually doing.

creating_packages/package_dev_flow.rst
memsharded: Is not only a fix for ``cmake.install()``, but also for pre-compiled binaries that already come in the final package layout, and still want to maintain the ``package()`` method to be able to package from build folder. Actually, this is the complete workflow that uses the output of the ``$ conan package`` command.

creating_packages/package_dev_flow.rst
danimtb: > does not require the execution of the above ``$ conan package``

It is OK with me, I would just omit the above sentence and keep the rest of the explanation

PR #567, opened 2018-03-06
Don't use $ for command names, improve bincrafter's blog link

liberforce:
Using $ as a prefix when you're describing a prompt is ok, as it's a reminder.
It helps knowing the command should be (in that specific context) called as a user ($ prefix) and not as root (# prefix).
However, sprinkling it everywhere adds more noise than information, and makes it harder to read.

### DIFF
--- a/creating_packages/package_dev_flow.rst
+++ b/creating_packages/package_dev_flow.rst
@@ -12,8 +12,8 @@ some costly operations:
 But sometimes, specially with big libraries, while we are developing the recipe, **we cannot afford** to perform every time these
 operations.
 
-The following section is the local development flow description based on the Bincrafters community at
-`their Blog <https://bincrafters.github.io>`_.
+The following section is the local development flow description based on the
+`Bincrafters community blog <https://bincrafters.github.io>`_.
 
 ----
 
@@ -23,8 +23,8 @@ phase.
 
 We will use the following `conan flow example <https://github.com/memsharded/example_conan_flow.git>`_ to follow the steps in the order below:
 
-$ conan source
-^^^^^^^^^^^^^^
+conan source
+^^^^^^^^^^^^
 
 You will generally want to start off with the :command:`conan source` command. The strategy here is that you’re testing your source method in
 isolation, and downloading the files to a temporary sub-folder relative to the *conanfile.py*. This just makes it easier to get to the
@@ -50,8 +50,8 @@ This method outputs the source files into the source-folder.
 Once you’ve got your source method right and it contains the files you expect, you can move on to testing the various attributes and methods
 relating to the downloading of dependencies.
 
-$ conan install
-^^^^^^^^^^^^^^^
+conan install
+^^^^^^^^^^^^^
 
 Conan has multiple methods and attributes which relate to dependencies (all the ones with the word "require" in the name). The command
 :command:`conan install` activates all them.
@@ -75,8 +75,8 @@ This also generates *conaninfo.txt* and *conanbuildinfo.xyz* (extension depends 
 (``install-folder``), which will be needed for the next step. Once you've got this command working with no errors, you can move on to
 testing the ``build()`` method.
 
-$ conan build
-^^^^^^^^^^^^^
+conan build
+^^^^^^^^^^^
 
 The build method takes a path to a folder that has sources and also to the install folder to get the information of the settings and
 dependencies. It uses a path to a folder where it will perform the build.
@@ -104,8 +104,8 @@ dependencies. It uses a path to a folder where it will perform the build.
 This is pretty strightforward, but it does add a very helpful new shortcut for people who are packaging their own library. Now, developers
 can make changes in their normal source directory and just pass that path as the ``--source-folder``.
 
-$ conan package
-^^^^^^^^^^^^^^^
+conan package
+^^^^^^^^^^^^^
 
 Just as it sounds, this command now simply runs the ``package()`` method of a recipe. It needs all the information of the other folders in
 order to collect the needed information for the package: header files from source folder, settings and depency information from the install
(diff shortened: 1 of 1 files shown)


danimtb:
Hi @liberforce,

Thank you for this PR. We really appreaciate every improvement to the documentation as it is the reference for conan users.

I think the Bincrafters link is better that way. However, using $ for inline command line commands in a text is part of our [documentation style guidelines](https://github.com/conan-io/docs#style-guidelines). We are currently migrating to keep the same style in the whole docs on the purpose of making the reading uniform and pleasant.

We think this way readers will associate different names of files, commands, code... in a more intuitive way.

Please note that we will consider any suggestion to improve this (based on the example of other software documentation similar to conan for example) and we work everyday to make docs better.

Thanks again!

liberforce:
Hi @danimtb, thanks for your quick answer.

I think this part of the documentation guidelines isn't adequate, as it adds visual clutter while giving no information.

It mixes a command (a concept) and a call to a command in a specific shell context (an instance). Adding `$` or `#` in front of calls to commands might add some value, to clarify the call context (even if I think it's too unix-ish for a cross-platform tool). However for a command name, it adds visual clutter to something that should be read as an english sentence.

If I write:
> The `$ conan build` command allows to build your project

I don't see any added value compared to:

> The `conan build` command allows to build your project

As I said, a command name is a concept, not an instance. Type `conan build` in a shell and you will have a `error too few arguments` error. The `$` and `#` are for real commands you type in a shell. They should be in a separate bash-formatted block. Command names are inline in a sentence, with a visual style that already make them stand out of the sentence, that should be enough.

Now you wanted examples of similar tools. Look at the pip documentation:
https://pip.pypa.io/en/stable/user_guide/#installing-packages
It only uses `$` for example of use of commands, i.e. commands that are really typed in a shell. They are inconsistent though, because sometimes there's a `$`, sometimes there's not, which is bad. They should use one or the other style. But on command names they are consistent: they never add a prefix.

Now npm:
https://docs.npmjs.com/getting-started/installing-npm-packages-locally
They use a different style for command names, and add prefixes to commands issued in a shell.

Now cargo, rust's package manager:
https://doc.rust-lang.org/cargo/guide/creating-a-new-project.html
Same thing: different style for the command name, prefix when issued in a shell.

So please consider changing the guidelines to this respect.

liberforce:
I'd be happy to do the changes in the whole repository if that's something you agree with.

danimtb:
Thank so much for the suggestions. I really think all of them are valid and good examples of documentation.

I see your point about inline commands (like: ``$ conan install``) about prefixes not adding infomration to readers/users.

So finally, I think the use of the prefix was about differentiating the highlight of code and the one used for commands.
Do you have any suggestion for that?

liberforce:
> So finally, I think the use of the prefix was about differentiating the highlight of code and the one used for commands.

From what I understand, you mean you use the `$` to differenciate highlighted text that is a command from highlighted text that is _not_ a command. The context of the sentence in itself is enough to differenciate those. Most of the time you currently already have something like:

> the `$ conan install` command

There is no ambiguity here. The context make it clear it is a command, it's written. Even without that, I fail to see how the reader can be mistaken.

For example in `creating_packages/package_dev_flow.rst`:

> ``source()`` method the next time you issue ``$ conan create`` using ``--keep-source``:

Compare this to:
> ``source()`` method the next time you issue ``conan create`` using ``--keep-source``:

What differentiates the command from the other highlighted elements ? Well, it starts with `conan`. Adding a `$` character is hence redundant. Even the parenthesis after `source` feel redundant as the very next word says it's a method.

There's also a reason why ithe option is written like this:

> using ``--keep-source``

and not

> using the ``--keep-source`` option

That's because the writer assumed the reader has enough background to know that something that starts with two dashes is an option and doesn't need to be told everytime. It's considered explicit by itself. In that case, I'm ok with both ways though.

Those are popular and widespread GNU-style conventions. If you're afraid people don't know them, either link to an external reference that explains them, or just use what you already have in your documentation style guidelines and add a "conventions" page at the beginning of the document.

BTW, like `pip`, the `git` book also doesn't use `$` in front of command names either.
https://git-scm.com/book/en/v2/Getting-Started-Installing-Git

danimtb:
I unerstand all your points and agree ``$`` doesn't add any value to inline commands.
However I think it would be nice for readers to highlight conan commands in a different way the inline code is done.

I haven't found a way to do so and I don't see the use of ``$`` very important regarding information to users. So, in case we dont find an alternative to differentitate them, I am fine with this PR removing $ from headers, but I would leave the inline text ones.

liberforce:
I didn't select the examples I gave: all the tools I looked at didn't differentiate the notation between their command names and options or other commands. Those are popular tools so I really don't feel like it's critical nor even useful for understanding.

So I've searched more:

Don't differentiate commands | Differentiate commands
------------ | -------------
git | [RubyGem](http://guides.rubygems.org/publishing/) (adds bold)
pip | 
npm |
cargo |
subversion |
gradle |
perforce |
[Portage](https://wiki.gentoo.org/wiki/Handbook:AMD64/Working/Portage) (special case: they differentiate options alone from options in the context of a command call, but don't differentiate commands)

So if you really *really* want to differentiate conan command names and make them stand out from the rest, I guess you could use bold, like in "use the **`conan install`** command".

danimtb:
Hi @liberforce!

We have finally changed the way to highlight inline commands: https://github.com/conan-io/docs#text-emphasishighlighting

Please, update this PR with the proposed guidelines and it could be merged.

Thank you so much for your research and suggestions, this made us improve the documentation style for future users! 👍

danimtb:
Hi @liberforce!

We really want to push this PR forward with the new changes, please address the changes and we will merge it :)

liberforce:
Hi, sorry for the delay was a bit busy. I've updated the pull request. Thanks for taking comments into account, this is greatly appreciated. I also saw that some options were marked using the `:command:` style, is this what you want? Options are different from commands, and the [text emphasis docs](https://github.com/conan-io/docs#text-emphasishighlighting) don't tell what style to use with options.

memsharded:
Hi @liberforce 

I understand that the ``:command:`` syntax applies not only to pure commands, but to the command line, including command line arguments.

liberforce:
> Command inline literals to refer to command line commands.

I don't see the option case as obvious, as an option alone is not a valid command line. So maybe the example sentence could be changed to reflect that:

> To create a conan package you can use :command:`conan create`. 
> You may call it with the :command:`--keep-source` option to avoid deleting and fetching the source.

But if you think the doc is explicit enough, that's fine for me.

memsharded:
Sure, here we go: https://github.com/conan-io/docs/commit/0a5d33f7ef7527d1e261ec6acd3a79eb488ad2a3

PR #607, opened 2018-03-26
Update using_profiles.rst

ClausKlein:
fix used profile file name in example


memsharded:
Nice catch, thanks for fixing this!

PR #608, opened 2018-03-26
Update getting_started.rst

ClausKlein:
fix the --build=missing options


memsharded:
In theory the command line parser accepts ``--build missing`` too, without the "=". But I much prefer to use it whenever possible, because it removes some ambiguities to the parser. 

So thanks :)

PR #612, opened 2018-03-28
[CLion.rst] Automatically run conan install on first build

tkhyn:
I was too lazy to remember the full command to install conan deps in the correct install folder :)


integrations/clion.rst
```diff
@@ -10,14 +10,15 @@ Just include the ``conanbuildinfo.cmake`` this way:
...
 .. code-block:: cmake
 
-   if(EXISTS ${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-       include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-       conan_basic_setup()
-   else()
-       message(WARNING "The file conanbuildinfo.cmake doesn't exist, you have to run conan install first")
+   if(NOT EXISTS ${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+       message(WARNING "The file conanbuildinfo.cmake doesn't exist, running conan install")
+       execute_process(COMMAND conan install .. -s build_type=${CMAKE_BUILD_TYPE} --install-folder=${CMAKE_BINARY_DIR})
```
memsharded: This has a problem, is that it won't use the same cmake current configuration, but the default machine configuration, which might be a problem, they don't necessary match.

memsharded:
Thanks very much for your contribution Thomas,

The problem is that the approach is not valid for all cases. When the default conan profile is different to the current CLion-cmake configuration, the approach fails. And it is something that will happen quite often.

I think you want to keep an eye into this repo: https://github.com/conan-io/conan-clion-plugin, it is ongoing work, but expected quite soon.

The other repo that might interest you is this cmake wrapper over conan, that automatically tries to get all the cmake configuration and translate it to conan settings. Should be more complete than your approach: https://github.com/conan-io/cmake-conan

tkhyn:
Oops, yes indeed. As it worked perfectly in my case with matching conan profile and Cmake config it did not occur to me that they could be different!

Thanks for the links, I look forward to using the ``conan-clion-plugin``!

(and thanks for all your work at the same time, I've just taken inspiration from your pybind11-example repo. It looks like you've worked on everything I need now (namely dependencies management for C/C++ and python to C/C++ bindings) to :)

PR #613, opened 2018-03-28
Docs for --json install output

lasote:


reference/commands/output/install.rst
```diff
@@ -0,0 +1,87 @@
+
+.. install_json::
```
memsharded: .. _install_json:

PR #615, opened 2018-04-01
Update vs2017_cmake.rst

raulbocanegra:
Add how to use tasks.vs.json to run conan install or any other conan related command from Visual Studio IDE directly

### DIFF
--- a/howtos/vs2017_cmake.rst
+++ b/howtos/vs2017_cmake.rst
@@ -160,6 +160,9 @@ Another alternative is using file `tasks <https://docs.microsoft.com/en-us/cpp/i
 
 All you need is to right click on your `conanfile.py`-> Configure Tasks (see the `link above <https://docs.microsoft.com/en-us/cpp/ide/non-msbuild-projects#define-tasks-with-tasksvsjson>`_) and add the following to your `tasks.vs.json`.
 
+.. warning:
+    The file `tasks.vs.json` is added to your local .vs folder so it is not supposed to be added to your version control system. There is also feature `request <https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/33814138-add-macro-buildroot-to-tasks-vs-json>`_ to improve this process.
+    
 .. code-block:: text
     :emphasize-lines: 7,9,16,18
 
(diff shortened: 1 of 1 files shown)


memsharded:
Hi @raulbocanegra 

This is interesting, but I have a couple questions:

- The configuration is not specified. cmake-conan auto-detects cmake configuration, but the tasks are not using any profile or setting, to differentiate between debug or release builds, or even the compiler (could be different)
- How do you know the install folder? It is not possible to know it in advance, right? So you cannot commit to version control these tasks to be used by other developers.

Is it intended only like a local helper?
Many thanks for your contribution!

raulbocanegra:
Well, it's a basic example.

> The configuration is not specified. cmake-conan auto-detects cmake configuration, but the tasks are not using any profile or setting, to differentiate between debug or release builds, or even the compiler (could be different)

You cann add the profile or settings to the conan install command, just wanted to simplify the example.

> How do you know the install folder? It is not possible to know it in advance, right? So you cannot commit to version control these tasks to be used by other developers.

Well, as you said, right now it's not poosible to know it in advance so I have to hardcode it like it's done in the conan install examples above in the same document, so you need to create one command for each configuration/build folder. Anyway, I have added a [suggestion](https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/33814138-add-macro-buildroot-to-tasks-vs-json) into Visual Studio user voice in order to be able to access the build folder from the tasks.vs.json via ${buildRoot} macro in CMakeSettings.json file of Visual Studio. We may need some votes to be heard probably :).

memsharded:
>  Anyway, I have added a suggestion into Visual Studio user voice in order to be able to access the build folder from the tasks.vs.json via ${buildRoot} macro in CMakeSettings.json file of Visual Studio. We may need some votes to be heard probably :).

That is great, thanks for the suggestion! Hopefully they implement it.

I think it would be better to explicitly:
- Add ``-s build_type=Release/Debug`` in the commands
- Put a clear comment and a warning (``.. warning:``) note that the folder is local and this is not to be committed to version control
- I am fine linking from that warning note the suggestion to Msft, so hopefully it gets more votes and attention :)

raulbocanegra:
I will update it with your suggestions

raulbocanegra:
I think it's done. feel free to add any suggestions.

memsharded:
It is perfect now, many thanks!

PR #631, opened 2018-04-18
Conan docset for Dash/Zeal

ytimenkov:
**Why**:
I found myself using Conan documentation a lot (says much about its quality 🌹), but docs.conan.io is slow (at best few seconds delay) and lacks of quick index.
Dash (or open source variant Zeal) provides quick local documentation and also reduces load on servers.

I would like to have docset for Conan too.

1. This prototype is fairly non-intrusive and kind of produces usable documentation.
2. I tried to put commands into `Makefile` but it finding `doc2dash` (like `sphinx-build`) didn't work. So was interesting to experiment with pipenv. I don't have strong opinion about it.
3. Patching css: it seems it is done for other docsets too. ToC us already there, so there is no need to take this space. Maybe there is a better solution.
4. Using generic headers vs more specialized entries: It seems possible to create custom sections based on path, but it's rather hackish. I wonder if it's possible to use standard Sphinx constructs like  `.. attribute::` and style them, so better index is generated even for online docs.

Screenshot:
image


ytimenkov:
> Is Dash just for Apple systems? This PR would serve for users interested in it, would be able to generate the docs for Dash?

Original dash is for Apple, but there are OSS cross-platform alternatives which use same format. I use Zeal on Windows (as you can see on screenshot).

I found it when I was looking for faster alternative to cppreference.com. I considered Qt documentation and similar tools, but they didn't have good up to date c++ reference.

Dash/Zeal provide literally instant index search compared to google + online docs combo (even with 100 Mbps channel) and has really wide range of docsets available.

Try it for Python reference :smile:

I'd like Conan docset to be published to the official dash feed, but of course you can host own feed.

> How does affect changing the titles to: .. attribute:: name for the regular sphinx generation? It is not intended as a replacement for the normal docs, right?

That's my concern. I tried few approaches: initially doc2dash didn't generate any sections at all, so converting headers to attributes did help. Then I went through the doc2dash issue tracker and found gist with example how to extend it and make it generate sections as "User guide".

Lastly I've realized that thanks to document organization it's possible to override types and sort sections into right categories completely non-intrusively, but I left it as `.. attribute::` just to show effect.

So either I could tweak script further to post-process documentation without any changes to .rst files (essentially it does require `make html`, but as I wrote above I couldn't add invocation of doc2dash without breaking).

or we could reuse richer Sphinx markup for entities and style them to look as before.

memsharded:
> So either I could tweak script further to post-process documentation without any changes to .rst files (essentially it does require make html, but as I wrote above I couldn't add invocation of doc2dash without breaking).

Of course, post-process of docs without any changes to .rst would be optimal for moving this forward faster.

I also think that it would be better to separate the introduction of Pipfile and lock file, so far it is using a plain ``requirements.txt`` and as only requiring Sphinx, it works pretty well, so adding there the necessary requirement would be good in this stage. At some moment we will be modernizing our repos to Pipfiles, but better one thing at a time :)

ytimenkov:
Good! I was thinking the same way during the lunch.

build.py
```diff
@@ -0,0 +1,58 @@
+from subprocess import check_call
+
+BUILD_DIR = "_build"
+
+check_call(
```
ytimenkov: Given that doc2dash is installed via requirements, is it OK to assume that it is available globally and I can just use it?
I mean `Makefile` allows overriding path to `sphinx-build` is this needed for dash too?

In other words I suggest removing call to `sphinx-build` from this file and add `Makefile`target which runs this python script.

Does it sound OK?

memsharded: Yes, yes, I think so (the only requirement is that doc2dash is something that installs nicely on most platforms). So it can be assumed that it is there and use it.

Makefile targets sounds perfect!

ytimenkov:
Now it should look better and less intrusive.

I had to go through attributes.rst and update label anchors and use them in converter scripts.

I'm also a bit confused about branches and where to merge: master and develop look diverged :(

ytimenkov:
I've "created" a feed to try the docs (Following https://kapeli.com/docsets#dashdocsetfeed)

https://github.com/ytimenkov/docs/releases/download/1.2.3-docs.1/Conan_feed.xml

Get Dash/Zeal and add URL above to feeds. 

Unfortunately due to some redirects from Github Zeal came up with GUID instead of proper name 😞

danimtb:
Hi @ytimenkov!

I tried Zeal with the feed you provided  and its looking really nice. Search is quite fast!

We would need more info on how to maintain this once it is merged to the master branch:

- Do we have to follow any rule to tag attributes, classes and so on?
  I have seen in your python script tou are tagging based on the current doc structure, so I wonder if it would be better to do it tagging each of our `.rst` somehow it gets the Zeal/Dash categorizing automated.

- How is the feed generated and how can we publish the Conan feed to an official Zeal/Dash one?

Regarding the base branch of this PR: Dont worry, we'll keep this pointing to master as this is not tied to any particular conan release and that is why you see the branches diverged.

Thanks a lot for your hard work on this!

ytimenkov:
@danimtb (trying my best from the phone)

> Do we have to follow any rule to tag attributes, classes and so on?
I have seen in your python script tou are tagging based on the current doc structure, so I wonder if it would be better to do it tagging each of our .rst somehow it gets the Zeal/Dash categorizing automated.

Initially I converted headings to specific domain anchors like `:py:attribute` or `py:method`. They handled by the script out of box, but need to be styled properly (something to what is done for commands). I think in long run it would be beneficial to have proper markup for all special entities rather than generic heading. But after discussion with @memshared we decided to make initial version less intrusive.

So currently yes, all attributes need to have anchor starting with `_attribute_`. Not best from maintenance perspective, but worst that can happen is it will be missing from the index.

I think **Class** where I put helpers can also be categorized automatically. Putting generators into "plugin" section is pure test of concept. In general or should be possible to create extra domains (in terms of Sphinx) and map them to dash categories to decouple from document structure.

I could experiment little but more with it.

> How is the feed generated and how can we publish the Conan feed to an official Zeal/Dash one?

Here are my guessings.

Feed is just an xml file which acts as a symlink to the latest version of tarball with docset. (Any) change in version in it tells the client that an update is available. Both Xml and tarball can be served from anywhere.

Here is doc to contribute to the official dash: https://github.com/Kapeli/Dash-User-Contributions#contribute-a-new-docset

As I understand you create a PR with a *binary tarball* and it gets added.

So I think that when you decide to upload new version of Conan docs to docs.conan.io you can generate dash docset at the same time and create PR. This could probably be automated.
Or you can serve 2 files from any place or cdn you want.

danimtb:
Hi @ytimenkov 

Since Conan 1.5 we re-enabled the possibility to download the Conan documentation in a PDF.
This was somehting requestd by users in #682 and now we freeze the documentation for every Conan release.

Having a look to your Zeal/Dash proposal again looks like very convenient for tool documentation and I really see yhe value of this PR.

However, seeing the amount of work required to make it work and auto-update the official repository, the fact the we are no experts on Dash/Zeal and that there was no feedback from the community requesting this kind of format, we fail to see the additional value of this now that offline documentation is well covered with PDFs.

Feel free to open an issue to get feedback from users to add Dash/Zeal support in the future and we would be glad to come to this PR and get the valuable information.

Thanks a lot for your work and hope this makes sense 😄

ytimenkov:
No worries! Good to hear about pdf.

PR #634, opened 2018-04-20
Update commands help + script

danimtb:
Docs for PR conan-io/conan#2770
Issue conan-io/conan#2766

- Updated messsages
- Added internal reference to all commands -> .. _conan_install:``
- Changed `code-block` of arguments description to `text` type to avoid problems in highlight
- Added double newline at end of commands help

- Added a quick python script to update this messages with a single command (could be removed and I will keep it for myself):
```bash
$ python command_help_update.py conandev.bat
```


reference/commands/commands_help_update.py
```diff
@@ -0,0 +1,103 @@
+import os
```
memsharded: Better put it in the root folder, not here.

PR #635, opened 2018-04-21
Allowing requiring color output with ``CONAN_COLOR_DISPLAY=1``

lachlankrautz:
- Change CONAN_COLOR_DISPLAY
  - default undefined
  - set to 1 to force color

### DIFF
--- a/reference/env_vars.rst
+++ b/reference/env_vars.rst
@@ -78,9 +78,10 @@ CONAN_COLOR_DISPLAY
 
 **Defaulted to**: Not defined
 
-Useful to remove or require colored output.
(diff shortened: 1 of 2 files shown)


reference/env_vars.rst
```diff
@@ -76,9 +76,11 @@ instead of Yellow, Cyan, White.
 CONAN_COLOR_DISPLAY
 -------------------
 
-**Defaulted to**: ``True``/``1``
+**Defaulted to**: Not defined
```
lasote: Mention that by default (not defined) it will use color display when tty is detected.

PR #653, opened 2018-05-14
Default options on conditional requirements

kznamst:
Added an explicit example on how to specify the default_options for a conditional requirement.


danimtb:
Thanks for the contribution!

PR #677, opened 2018-06-08
Updated virtualrunenv generator and some fixes

### LINKED ISSUE #676: Confusing how-to of shared libraries regarding virtualrunenv
The page https://docs.conan.io/en/latest/howtos/manage_shared_libraries.html is not using virtualrunenv, but instead virtualenv and defining ``package_info()``. This section has to be updated, it is not clear.

Comes from https://github.com/conan-community/conan-tbb/issues/6

cc /@db4

danimtb:
closes #676

### DIFF
--- a/howtos/manage_shared_libraries/env_vars.rst
+++ b/howtos/manage_shared_libraries/env_vars.rst
@@ -24,10 +24,8 @@ the right environment variable set, so they will be able to locate the (A) share
 Similarly if you use the :ref:`virtualenv generator<virtual_environment_generator>` and you
 activate it, you will get the paths needed to locate the shared libraries in your terminal.
 
-
 Example
 -------
-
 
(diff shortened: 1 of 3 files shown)


howtos/manage_shared_libraries/env_vars.rst
```diff
@@ -215,14 +166,12 @@ the environment variables poiting to the "lib" and "bin" folders.
...
 
 .. code-block:: bash
 
-    conan install .
-    source activate
-    toolA --someparams
-
+    $ conan install .
+    $ source activate_run
+    $ toolA --someparams
```
memsharded: I think the security restriction of OSX is also active here, and it might be ncessary to do the DYLD_LIBRARY_PATH = in the command line.

PR #679, opened 2018-06-08
Clarify paths relative to src in self.copy() in package()

### LINKED ISSUE #577: Clarify copy() method "src" parameter
I find it a bit difficult to understand how `copy()` works.
http://docs.conan.io/en/latest/reference/conanfile/methods.html#package

The doc says (emphasis mine):
> The folder where you want to search the files in the **build folder**. If you know that your libraries when you build your package will be in build/lib, you will typically use build/lib in this parameter.
Leaving it empty means the **root build folder** in local cache.

As source, install, build and package folders can be located pretty much anywhere using the `--source-folder`-like options, `build/lib` doesn't give me much information. It's a relative path, but relative to what?

When I initialize a new package with `conan new`, the sample contains in the `package()` method:
`self.copy("*.h", dst="include", src="hello")`

So now here `src` seems relative to the source folder, as it's where the code is checked out in the `hello` example, but the doc says `src` is relative to the build folder. So how does it works?

danimtb:
closes #577

### DIFF
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -176,10 +176,13 @@ And it will copy the lib to the package folder *lib/Mylib.lib*, which can be lin
     If you are using CMake and you have an install target defined in your CMakeLists.txt, you might be able to reuse it for this
     ``package()`` method. Please check :ref:`reuse_cmake_install`.
 
-The ``package()`` method will be called twice if the attribute ``no_copy_source`` is defined and ``True``. One will copy from the *source*
-folder (typically packaging the headers and other data files), and the other will copy from the *build* folder, packaging the libraries and
-other binary artifacts. Also, when the local :command:`conan package` command is issued with :command:`--source-folder` and :command:`--build-folder`, it will
-execute two times, one in each folder, in the same way.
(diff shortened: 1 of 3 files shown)


reference/conanfile/methods.rst
```diff
@@ -176,10 +176,13 @@ And it will copy the lib to the package folder *lib/Mylib.lib*, which can be lin
     If you are using CMake and you have an install target defined in your CMakeLists.txt, you might be able to reuse it for this
     ``package()`` method. Please check :ref:`reuse_cmake_install`.
 
-The ``package()`` method will be called twice if the attribute ``no_copy_source`` is defined and ``True``. One will copy from the *source*
-folder (typically packaging the headers and other data files), and the other will copy from the *build* folder, packaging the libraries and
-other binary artifacts. Also, when the local :command:`conan package` command is issued with :command:`--source-folder` and :command:`--build-folder`, it will
-execute two times, one in each folder, in the same way.
+Normally during :command:`conan create` this method will copy all the indicated files from build folder (``src``) as source folder content is
```
memsharded: I think this explanation is even more complicated than the original one. Maybe split the cases in bullets.

PR #689, opened 2018-06-17
Explain wildcard filter for unzip and untargz

MathPlayer:
Add documentation to explain how to filter paths when decompressing files.
This explains the implementation for conan-io/conan#2938.

### DIFF
--- a/reference/tools.rst
+++ b/reference/tools.rst
@@ -202,7 +202,7 @@ tools.unzip()
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


MathPlayer:
The `changelog.rst` on the develop branch does not contain yet an entry for the new / unreleased version. I don't know what is the preferred way to proceed in adding the information there since there is already a pull requests doing this (see #685).

memsharded:
Thanks for the good docs!

Don't worry about the entry in the changelog, we will add it.

PR #698, opened 2018-06-21
Constrain settings and options section

### LINKED ISSUE #692: Add section to explain how to constraint settings and options
Add answer to question in conan-io/conan#1313 to docs.

Probably in mastering conan section.

danimtb:
closes #692

### DIFF
--- a/mastering/conditional.rst
+++ b/mastering/conditional.rst
@@ -76,8 +76,8 @@ There are two approaches for this situation:
 
 - **Constrain settings inside a recipe**:
 
-  This approach constrains the settings inside a recipe to a subset of them and it is normally used in recipes that are never supposed to be
-  used with the restricted settings.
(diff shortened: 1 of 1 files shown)


mastering/conditional.rst
```diff
@@ -46,5 +46,56 @@ Here is an example of what we could do in our **configure method**:
...
+
+  This same method is also valid for ``options`` and ``config_options()`` method and it is commonly used to remove options for one setting:
+
+  .. code-block:: python
+
+      def config_options(self):
+          if self.settings.os == "Windows":
+              del self.options.fPIC
+
+- **Constrain settings inside a recipe**:
```
memsharded: I am not a big fan of this feature. It is verbose and can become ugly if trying to constrain some subsetting. Generally prefer the configure() method and only for some special cases where it is 100% that such code will never, ever work with that setting. Please add some note or comment, suggesting this, if possible.

PR #726, opened 2018-07-13
Note about pylint plugin

ytimenkov:
Documentation part for conan-io/conan#3117

Please advise on better location/title. Otherwise I'm not a fan of reading lengthy docs :)

I was thinking about mentioning at least 2 IDEs (VS Code and PyCharm) but I don't have the latter...


danimtb:
hi @ytimenkov, thanks a lot for the docs!

It looks good to me. WDYT @memsharded ?

PR #728, opened 2018-07-16
Review "Define ABI compatibility" section and documented requires modes

### LINKED ISSUE #718: `0.1` version should be `1.0`
On https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html?highlight=transitive#the-problem-of-dependencies

> Then, in this case, MyLib/0.1 has to build a new binary for the new dependency version. Otherwise, it will maintain the old, buggy addition() version. Even if MyLib/0.1 hasn’t change a line, not the code, neither the recipe, still the resulting binary would be different.

This `0.1` version should be `1.0` in this paragraph

### LINKED ISSUE #719: Document self.info.requires modes.
in https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html?highlight=transitive#versioning-schema

```python
def package_id(self):
    # Default behavior, only major release changes the package ID
    self.info.requires["MyOtherLib"].semver_mode()

    # Any change in the require version will change the package ID
    self.info.requires["MyOtherLib"].full_version_mode()

    # Any change in the MyOtherLib version, user or channel will affect our package ID
    self.info.requires["MyOtherLib"].full_recipe_mode()

    # Any change in the MyOtherLib version, user or channel or Package ID will affect our package ID
    self.info.requires["MyOtherLib"].full_package_mode()

    # The requires won't affect at all to the package ID
    self.info.requires["MyOtherLib"].unrelated_mode()
```

Please describe the following in the docs:
1. semver_mode()
2. full_version_mode()
3. full_recipe_mode()
4. full_package_mode()
5. unrelated_mode()

PLUS any other modes not in the code extract above.

danimtb:
closes #718, closes #719

### DIFF
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -24,7 +24,7 @@ When this package is installed by a *conanfile.txt*, another package *conanfile.
 
 The process will be:
 
-1. Conan will get the user input settings and options. Those settings an options can come from command line, profiles or from the values
+1. Conan will get the user input settings and options. Those settings and options can come from command line, profiles or from the values
    cached in the latest :command:`conan install` execution.
 2. Conan will retrieve the ``MyLib/1.0@user/channel`` recipe, read the ``settings`` attribute, and assign the necessary values.
 3. With the current package values for ``settings`` (also ``options`` and ``requires``), it will compute a SHA1 hash that will be the binary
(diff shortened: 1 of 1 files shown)


creating_packages/define_abi_compatibility.rst
```diff
@@ -24,39 +24,33 @@ When this package is installed by a *conanfile.txt*, another package *conanfile.
...
-2. Conan will retrieve the ``MyLib/1.0@user/channel`` recipe, read the ``settings`` attribute, and
-   assign the necessary values.
-3. With the current package values for ``settings`` (also ``options`` and ``requires``), it will
-   compute a SHA1 hash, that will be the binary package ID, e.g.
-   ``c6d75a933080ca17eb7f076813e7fb21aaa740f2``.
-4. Conan will try to find the ``c6d75...`` binary package. If it's present conan will retrieve it,
-   if not, it can be built from sources with :command:`conan install --build`.
-
-If the package is installed again with different settings, for example, for 32bits architecture:
+1. Conan will get the user input settings and options. Those settings an options can come from command line, profiles or from the values
```
memsharded: settings "and" options

danimtb:
Agree! better an unified table

PR #731, opened 2018-07-16
Line from config.rst missing in official docs. (second try)

pepe82sh:
`code-block` directive without a language specification doesn't seem to work properly in sphinx. [The official documentation](https://docs.conan.io/en/latest/reference/commands/consumer/config.html) is missing that line completely. I couldn't find a version in the repo that has that error, so I assume it's not generated correctly.


danimtb:
This is it! Thanks a lot @pepe82sh 😄

Issue #716, opened 2018-07-09
Document cmake/make/make install recipe for in-house monolithic code base


mristin:
Hi conan team!

While we were trying to migrate the compilation and releasing of our (still monolithic) code base to conan, we were a bit confused how to use it with cmake wrapper from within a python build script.

Our usual work flow included:

* cd to the build directory, cmake {src} 
* make
* make check
* make install

We needed the "install" step so that the rpaths are removed, while we required the rpaths to be specified in the build phase to facilitate the unit testing (otherwise `make check` fails).

We looked at the documentation at https://docs.conan.io/en/latest/howtos/cmake_install.html. Unfortunately, it did not really help us further as this work flow was nowhere documented in full.

Here is what we come up with for the `conanfile.py`:

```python
import conans
import conans.tools


class Pqry(conans.ConanFile):
    name = "Pqry"
    version = "1.0.0"
    settings = "os", "compiler", "build_type", "arch"
    options = {"shared": [True, False]}
    generators = "cmake"
    keep_imports = True

    requires = (
        # some libraries here...
    )

    default_options = ("shared=True",
                       # some parameters here...
                       )

    def build(self):
        cmake = conans.CMake(self)
        cmake.configure()
        cmake.build()

    def package(self):
        cmake = conans.CMake(self)
        cmake.install()
```

In our python build script (which is our main point of entry), we wrote:

```python
import conans

def install_codebase(
    repo_dir: pathlib.Path, 
    build_dir: pathlib.Path, 
    release_dir: pathlib.Path) -> None:
    """
    Installs the Parquery codebase.

    :param repo_dir: where Parquery repository resides
    :param build_dir: where the build files will be stored
    :return:
    """
    source_dir = repo_dir / "src" / "cpp"
    cmake_build_debug_dir = source_dir / "cmake-build-debug"
    cmake_build_debug_dir.mkdir(exist_ok=True)
    build_dir.mkdir(exist_ok=True, parents=True)

    conan, _, _ = conans.client.conan_api.ConanAPIV1.factory()

    # Install files for clion
    conan.install(cwd=cmake_build_debug_dir.as_posix(), path="..", build=["missing"])

    # Install dependencies and build in the build directory
    conan.install(cwd=build_dir.as_posix(), path=source_dir.as_posix(), build=["missing"])

    # Specifying the argument 'package_folder' is necessary 
    # since it needs to be set at the build time for CMake.
    conan.build(cwd=build_dir.as_posix(),
[...]

PR #750, opened 2018-07-23
Updated reuse_cmake_install with an example of package()

mristin:
This update is related to the issue #716 .

### DIFF
--- a/howtos/cmake_install.rst
+++ b/howtos/cmake_install.rst
@@ -5,16 +5,13 @@ How to reuse cmake install for package() method
 
 It is possible that your project's *CMakeLists.txt* has already defined some
 functionality that extracts the artifacts (headers, libraries, binaries) from
-the build and source folder to a predetermined place.
+the build and source folder to a predetermined place and does the post-processing (*e.g.*, strips rpaths). For example,
+one common practice is to use CMake `install <https://cmake.org/cmake/help/latest/command/install.html>`_ directive
+to that end.
 
-The conan ``package()`` method does exactly that: it defines which files
-have to be copied from the build folder to the package folder.
+When using conan, you need to wrap that functionality in the conan ``package()`` method.
 
-If you want to reuse that functionality, you can do it with cmake.
-
-Invoke cmake with ``CMAKE_INSTALL_PREFIX`` using the ``package_folder`` variable.
-If the ``cmake install`` target correctly copies all the required libraries, headers, etc. to the ``package_folder``,
-then the ``package()`` method is not required.
+The following excerpt shows how to build and package with CMake within conan:
 
 .. code-block:: python
 
@@ -22,16 +19,34 @@ then the ``package()`` method is not required.
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
-        cmake.install()
-        # equivalent to
-        # args += ['-DCMAKE_INSTALL_PREFIX="%s"' % self.package_folder]
-        # self.run('cmake "%s/src" %s %s'
-        #          % (self.source_folder, cmake.command_line, ' '.join( args ) ) )
-        # self.run("cmake --build . --target install %s" % cmake.build_config)
 
     def package(self):
-        # nothing to do here now
+        cmake = CMake(self)
+        cmake.install()
 
-The ``package_info()`` method is still necessary, as there is no possible way to
-automatically extract the information of the necessary libraries, defines and flags for different
-build configurations from the cmake install.
+If you want to specify a different ``CMAKE_INSTALL_PREFIX`` for your package, you need to specify it as
+``package_folder`` argument both in a call to conan ``build()`` and ``package()``, respectively.
+(Maybe a bit unexpectedly, you can not specify ``package_folder`` only in the call to ``package()`` since CMake needs
+to know the install prefix already at the building stage.)
+
+The following excerpt from a custom script (meant to install conan, build and package the code) shows you how to set
+``package_folder`` properly:
+
+.. code-block:: python
+
+    import conans
+
+    def conan_build_and_install_my_codebase():
+        conan, _, _ = conans.client.conan_api.ConanAPIV1.factory()
+
+        source_folder = "/some/path/to/the/sources"
+        build_folder = "/some/path/to/the/build/folder"
+        package_folder = "/some/path/to/the/install/prefix"
+
+        conan.install(cwd=build_folder, path=source_folder, build=["missing"])
+
+        # We already need to specify the package_folder in the build() since CMake already needs
+        # the install prefix at this stage.
+        conan.build(cwd=build_folder, conanfile_path=source_folder, package_folder=package_folder)
+
+        conan.package(build_folder=build_folder, path=source_folder, package_folder=package_folder)


danimtb:
Hi @mristin, thanks a lot for this PR! Could you please check the CLA sign? We need it to get this PR accepted.

Regarding the changes in this PR, one of the things that made us not kind of recommend this usage is because package() method would be called twice in case you define [no_copy_source = True](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#no-copy-source).

Also, as conanfile methods are executed independently, you have to create a new CMake object in package() and configure same definitions, for example: https://github.com/bincrafters/conan-mbedtls/blob/414f0a54ba604f5fcd56b85a03874b717ae5036b/conanfile.py#L54

So finally I would say it is a trade-off. And I'd say that if we are going to make this change, it will be better to include the wholse solution as in the example linked above.

mristin:
@danimtb thank you for the review!

Please apologize for somewhat amateurish pull requests  -- I'm new to conan and we have just started to use it in the company. 

> Regarding the changes in this PR, one of the things that made us not kind of recommend this usage is because package() method would be called twice in case you define no_copy_source = True.

After reading the docs you pointed me to I'm a bit puzzled why the `package()` needs to be called twice. Wouldn't CMake (_i.e._ `cmake.install()`) be responsible for all the copying including the headers or any other source files? Could you please give me a bit more details so that I can also update the docs in this pull request? I suppose this merits at least a short paragraph.

> Also, as conanfile methods are executed independently, you have to create a new CMake object in package() and configure same definitions, [...]

Could you please have a look at the commit 0675a46b038118feb494ccd92d24cbd18454d28a and let me know if that is what you had in mind or something more complex?

danimtb:
Hi @mristin,

Don't worry, every PR (especially to the docs) is always welcomed and we really appreciate the feedback from users.

Regarding to the commit you pointed: yes, that is exactly what I was trying to highlight.

For the case of `package()` method being called twice, this happens when you have `no_copy_source = True`. First, let me explain how Conan behaves:

- Conan retrieves the sources from `exports`, `exports_sources` & `source()` and puts them in a source folder.
- Before compiling, Conan creates a copy of the source folder into the build folder
- Then the build starts.
- After the build Conan creates the package collecting the files and artifacts from the build folder (using package() method of the recipe).

Now, if you set  `no_copy_source = True` Conan behaves this way:

- Retrieve sources (same as above).
- Build starts in the build folder (sources are in source folder -NOT copied to build folder-
- After the build, Conan creates the package copying files from the source folder (this calls `package()`) and copying the artifacts from the build folder (this calls `package()` again).

So in the case you were using `cmake.install()` in the package method and `no_copy_source`, the install step will be done twice. Files will be overwritten with issues but install steps are sometimes time expensive and this will double the "packaging" time.

mristin:
@danimtb thanks for your explanation! 

I'm still a bit puzzled about this step:
> After the build, Conan creates the package copying files from the source folder (this calls package()) and copying the artifacts from the build folder (this calls package() again).

Why would you call the same function twice that doesn't include any parameter to change its behavior? I looked into:
https://github.com/conan-io/conan/blob/c9df68faf57003dbe480f5dbbbe887ec8aeed4a9/conans/model/conan_file.py#L247

```python
    def package(self):
        """ package the needed files from source and build folders.
        E.g. self.copy("*.h", src="src/includes", dst="includes")
        """
self.output.warn("This conanfile has no package step")
```
so there seems to be no side effects. Are there parameters I'm not aware of (like `package_folder` dynamically changing between the two steps in the concrete instance of `ConanFile`)?

mristin:
Hi @danimtb,
Just a kind reminder in case you missed the last message.

danimtb:
Hi @mristin,

As I told you before (sorry about the mistake):

> So in the case you were using `cmake.install()` in the package method and `no_copy_source`, the install step will be done twice. Files will be overwritten witho**OUT** issues but install steps are sometimes time expensive and **this will double the "packaging" time**.

Currently I thinks the best would be to make the changes you are suggesting in this PR (cmake.install() in package()) and include a warning message regarding ``no_copy_source``. For the future I think we should improve this in the conan client regarding other build helpers too.

howtos/cmake_install.rst
```diff
@@ -5,33 +5,54 @@ How to reuse cmake install for package() method
...
+
+If you want to specify a different ``CMAKE_INSTALL_PREFIX`` for your package, you need to specify it as
+``package_folder`` argument both in a call to conan ``build()`` and ``package()``, respectively.
+(Maybe a bit unexpectedly, you can not specify ``package_folder`` only in the call to ``package()`` since CMake needs
+to know the install prefix already at the building stage.)
+
+The following excerpt from a custom script (meant to install conan, build and package the code) shows you how to set
+``package_folder`` properly:
+
+.. code-block:: python
```
danimtb: Please do not include code using the Conan internal API as this is not stable enough to recommend its usage in the documentation.

howtos/cmake_install.rst
```diff
@@ -5,33 +5,54 @@ How to reuse cmake install for package() method
...
+        build_folder = "/some/path/to/the/build/folder"
+        package_folder = "/some/path/to/the/install/prefix"
+
+        conan.install(cwd=build_folder, path=source_folder, build=["missing"])
+
+        # We already need to specify the package_folder in the build() since CMake already needs
+        # the install prefix at this stage.
+        conan.build(cwd=build_folder, conanfile_path=source_folder, package_folder=package_folder)
 
-The ``package_info()`` method is still necessary, as there is no possible way to
```
danimtb: Please, do not remove this note as it is relevant to propagate the package information and remind the user to do so

howtos/cmake_install.rst
```diff
@@ -5,33 +5,54 @@ How to reuse cmake install for package() method
...
-        # self.run('cmake "%s/src" %s %s'
-        #          % (self.source_folder, cmake.command_line, ' '.join( args ) ) )
-        # self.run("cmake --build . --target install %s" % cmake.build_config)
 
     def package(self):
-        # nothing to do here now
+        cmake = self.configure_cmake()
+        cmake.install()
+
+If you want to specify a different ``CMAKE_INSTALL_PREFIX`` for your package, you need to specify it as
```
danimtb: Dont see the point here to change CMAKE_INSTALL_PREFIX as you want it to be pointing to the package folder as it is already. Cant see the use case

mristin:
Hi @danimtb ,
I tried to write down what we discussed here regarding the different work flows when `no_copy_source` is set.

I also added a more verbose paragraph on `package_info()`. That part was also not clear to me, and having the redundant explanation why it is needed (aside the main `package_info()` page) would help understanding, IMO. 

I removed the example that leverages internal conan API.

howtos/cmake_install.rst
mristin: Please mind that we don't want to change `CMAKE_INSTALL_PREFIX`. We want to change the default `package_folder` to an arbitrary folder. Conan will pass this value to CMake which eventually sets `CMAKE_INSTALL_PREFIX`. 

We have a release script which we use to build and install in arbitrary folders. We use conan only to manage the dependencies, but don't have any downstream conan consumers.

I wanted to include this example to help other people as it took us long to figure it out, but I suppose it makes most sense if we just keep using it internally and leave it undocumented here as long as API is not stable.

howtos/cmake_install.rst
```diff
@@ -5,33 +5,69 @@ How to reuse cmake install for package() method
...
+The following excerpt shows how to build and package with CMake within conan. Mind that you need to configure CMake
+both in ``build()`` and in ``package()`` since these methods are called independently.
 
 .. code-block:: python
 
-    def build(self):
+    def configure_cmake(self):
         cmake = CMake(self)
         cmake.configure()
+        cmake.definitions["SOME_DEFINITION"] = "On"
```
memsharded: I think this ``cmake.definitions`` should be **before** ``cmake.configure()``, otherwise it is ignored.

howtos/cmake_install.rst
```diff
@@ -5,33 +5,69 @@ How to reuse cmake install for package() method
...
     def package(self):
-        # nothing to do here now
+        cmake = self.configure_cmake()
+        cmake.install()
 
-The ``package_info()`` method is still necessary, as there is no possible way to
-automatically extract the information of the necessary libraries, defines and flags for different
-build configurations from the cmake install.
+    def package_info(self):
+        self.cpp_info.includedirs = ['include']
```
memsharded: These are the defaults, it is not necessary to define them, and it might be confusing to do so, without a clear warning or something.

danimtb:
@memsharded done! https://docs.conan.io/en/latest/howtos/cmake_install.html

- Fixed definitions.
- Changed package_id()
- Removed redundant information.
- Linked to no_copy_source reference

PR #751, opened 2018-07-23
Added example subsection to _cmake_reference

mristin:
This update is related to the issue #716 .

### DIFF
--- a/reference/build_helpers/cmake.rst
+++ b/reference/build_helpers/cmake.rst
@@ -309,6 +309,7 @@ The following example of ``conanfile.py`` shows you how to manage a project with
 
     def configure_cmake(self):
         cmake = CMake(self)
+        cmake.configure()
 
         # put definitions here so that they are re-used in cmake between
         # build(), test() and package()
@@ -319,10 +320,7 @@ The following example of ``conanfile.py`` shows you how to manage a project with
     def build(self):
         cmake = self.configure_cmake()
         cmake.build()
-
-    def test(self):
-        cmake = self.configure_cmake()
-        self.run("make check")
+        cmake.test()  # run all the unit tests after the build
 
     def package(self):
         cmake = self.configure_cmake()
--- a/reference/build_helpers/cmake.rst
+++ b/reference/build_helpers/cmake.rst
@@ -30,15 +30,15 @@ There are two ways to invoke your cmake tools:
 
     class ExampleConan(ConanFile):
         ...
-        
+
         def build(self):
             cmake = CMake(self)
             # same as cmake.configure(source_folder=self.source_folder, build_folder=self.build_folder)
-            cmake.configure()  
+            cmake.configure()
             cmake.build()
             cmake.test() # Build the "RUN_TESTS" or "test" target
             # Build the "install" target, defining CMAKE_INSTALL_PREFIX to self.package_folder
-            cmake.install() 
+            cmake.install()
 
 
 Constructor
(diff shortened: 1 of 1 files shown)


reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,42 @@ Environment variables
...
 :ref:`CMAKE RELATED VARIABLES<cmake_related_variables>` section.
+
+Example
+-------
+The following example of ``conanfile.py`` shows you how to manage a project with conan and CMake.
+
+.. code-block:: python
+
+    import conans
+    import conans.tools
```
danimtb: `from conans import ConanFile, CMake, tools`

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,42 @@ Environment variables
...
+
+    class SomePackage(conans.ConanFile):
+        name = "SomePackage"
+        version = "1.0.0"
+        settings = "os", "compiler", "build_type", "arch"
+        options = {"shared": [True, False]}
```
danimtb: Remove not relevant information for this example: options & default_options, keep_imports, requires

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,38 @@ Environment variables
...
+        # build(), test() and package()
+        cmake.definitions["SOME_DEFINITION_NAME"] = "On"
+
+        return cmake
+
+    def build(self):
+        cmake = self.configure_cmake()
+        cmake.build()
+
+    def test(self):
```
memsharded: This is good for ``test_package`` functionality, but this code is not run for normal packaging.
Maybe many users want to make sure to run the tests in the ``build()`` method, so they are run for every build of the package, and abort the packaging if they fail. In that case, the ``test()`` method will be removed, and the build would be:

```python
   def build(self):
        cmake = self.configure_cmake()
        cmake.build()
        cmake.test()
```

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,38 @@ Environment variables
...
+        cmake = self.configure_cmake()
+        self.run("make check")
```
memsharded: The ``make check`` is too Makefile specific. I think the general approach should be ``cmake.test()``. If for some reason it is not working, please report.

reference/build_helpers/cmake.rst
marko-ristin-parquery: I added the example commands for two scenarios: one that you suggest (run unit tests after the build), and another command to run a custom make command.

reference/build_helpers/cmake.rst
marko-ristin-parquery: See the comment re this change above.

mristin:
(Sorry, I closed the issue unintentionally.)

mristin:
Hi @memsharded,
Just a kind reminder in case the last commit got lost in the backlog -- would you mind reviewing?

mristin:
Hi @memsharded ,
I'm staring at the code but can't see where I missed the `cmake.configure()` :-(

If you compare against the master branch:

```python
    def configure_cmake(self):
        ...

    def build(self):
        cmake = self.configure_cmake()
        ...

    def package(self):
        cmake = self.configure_cmake()
        ...
```

Could you please point me?

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,40 @@ Environment variables
...
+        settings = "os", "compiler", "build_type", "arch"
+        generators = "cmake"
+
+    def configure_cmake(self):
+        cmake = CMake(self)
+
+        # put definitions here so that they are re-used in cmake between
+        # build() and package()
+        cmake.definitions["SOME_DEFINITION_NAME"] = "On"
+
```
memsharded: Maybe here, insert the call to ``cmake.configure()``?

mristin: Exactly :smile: added, sorry for the blind spot.

memsharded:
Many thanks!

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,41 @@ Environment variables
...
+    from conans import ConanFile, CMake
+
+    class SomePackage(ConanFile):
+        name = "SomePackage"
+        version = "1.0.0"
+        settings = "os", "compiler", "build_type", "arch"
+        generators = "cmake"
+
+    def configure_cmake(self):
+        cmake = CMake(self)
```
bryceschober: IMO, `set_cmake_flags=True` should be mentioned more prominently in the docs, and this is a good opportunity. It's the kind of option that I thought should be default behavior as newbie.

danimtb: It is well documented here https://docs.conan.io/en/latest/reference/build_helpers/cmake.html#constructor and has not raised any particular issue so far on its default value

bryceschober: Sorry, I was off-base on the current documentation context.

reference/build_helpers/cmake.rst
```diff
@@ -292,3 +292,41 @@ Environment variables
...
+
+    from conans import ConanFile, CMake
+
+    class SomePackage(ConanFile):
+        name = "SomePackage"
+        version = "1.0.0"
+        settings = "os", "compiler", "build_type", "arch"
+        generators = "cmake"
+
+    def configure_cmake(self):
```
johnmcfarlane, line 310: Indentation doesn't look right in this example.

memsharded: Definitely wrong, thanks

memsharded: https://github.com/conan-io/docs/pull/1954

PR #753, opened 2018-07-23
Py2 Deprecation

markgalpin:
Are we ready for this?

### DIFF
--- a/introduction.rst
+++ b/introduction.rst
@@ -122,13 +122,13 @@ If you have any question regarding Conan updates, stability, or any clarificatio
 
 Python 2 Deprecation Notice
 ---------------------------
-Through 1.6, all Conan features are fully supported in both Python2 and Python3.
-After 1.6, all features built prior to 1.6 will continue to be fully tested in Python 2 and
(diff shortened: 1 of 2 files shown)


introduction.rst
```diff
@@ -118,7 +118,20 @@ This means that public repositories, like conan-center assume the use of the lat
 
 If you have any question regarding Conan updates, stability, or any clarification about this definition of stability, please report in the documentation issue tracker: https://github.com/conan-io/docs.
 
-
+.. _python2:
+
+Python 2 Deprecation Notice
+---------------------------
+Before 1.6, all Conan features are fully supported in both Python2 and Python3.
```
danimtb: **were** fully

memsharded:
I agree, I think this should go in 2 places:
- The "install" section in the docs
- A link in the downloads page in conan.io

markgalpin:
I put it here (there's already a warning in Install, which I may strengthen later) because it feels like it impacts/modifies the stability commitment.  If we don't think so, I can move this other places.

memsharded:
I think the main point should be the installation section, and in any case, link it briefly from the stability subsection of the introduction.

PR #760, opened 2018-08-01
Provide guidance on reusing cmake install for package development

bryceschober:
Based on my frustrations and experiences in getting cmake.install() usage working with conan-package-tools, and learning from conan-io/conan-package-tools#247 and conan-io/conan/#2350.

### DIFF
--- a/howtos/cmake_install.rst
+++ b/howtos/cmake_install.rst
@@ -35,3 +35,57 @@ then the ``package()`` method is not required.
 The ``package_info()`` method is still necessary, as there is no possible way to
 automatically extract the information of the necessary libraries, defines and flags for different
 build configurations from the cmake install.
+
+Reusing cmake install for package developers
+--------------------------------------------
+
+Using the local development workflow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using the :ref:`local development workflow<package_dev_flow>`, the :command:`conan package` 
+command can obviously be skipped, since it has nothing to do. But when using the 
+:command:`conan export-pkg` command, you will need to specify the package-folder parameter:
+
+.. code-block:: bash
+
+    $ conan export-pkg . user/testing --package-folder=tmp/package
+
+Using the ``conan create`` workflow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In order to support the :command:`conan create` command (and the excellent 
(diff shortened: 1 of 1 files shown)


howtos/cmake_install.rst
```diff
@@ -35,3 +35,57 @@ then the ``package()`` method is not required.
...
+
+Reusing cmake install for package developers
+--------------------------------------------
+
+Using the local development workflow
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When using the :ref:`local development workflow<package_dev_flow>`, the :command:`conan package` 
+command can obviously be skipped, since it has nothing to do. But when using the 
+:command:`conan export-pkg` command, you will need to specify the package-folder parameter:
```
memsharded: This is not completely true. The ``package-folder`` is only necessary if the package has already been created locally (maybe ``conan package``), but probably the most common usage of ``conan export-pkg`` is directly from the ``build-folder``, and in this case the ``package-folder`` cannot be specified. This addition to the docs should be clear in that sense too.

bryceschober: But in this case we're doing a `cmake.install()`. Surely that doesn't install into the `build-folder` in normal circumstances. And the linked docs for the local development workflow don't say anything about `cd`'ing into the package directory.

Honestly, it's awkward to have my edits in any one of the three possible contexts (package dev, cmake, cmake-install howto) because they only really matter at the intersection of the three. On the other hand, it also seems to me that the intersection of those three contexts would be very common for newer package developers on the most common build platform, a use-case that matters a lot.

All that said, I'm not sure where to go from here...

bryceschober: @memsharded can you offer more guidance? I actually don't understand how your comment applies to my situation, where `conan package` has no implementation..

danimtb: I think what @memsharded is saying is that it is not necessary to provide a ``--package-folder`` when using `export-pkg`. ``--package-folder`` can be used after ``conan package -pf tmp/package`` but not mandatory either

danimtb:
@bryceschober could you update this PR with the latest changes? There was PR recently merged providing this section with some more information. Maybe it helps to clarify some doubts. Thanks! 😄

bryceschober:
I've reviewed the changes that would need to be merged, and it appears that others have filled the need I had at the same time and gotten their docs updates merged in 0fc2d339 and 443eaa4 to the point that this PR is redundant.

PR #772, opened 2018-08-21
Updated `conan upload` docs to include info about --query (#2445)

joechrisellis:
* Updated `conan upload` command signature in `reference/commands/creator/upload.rst`.
* Added an example of using the new `--query` option in `reference/commands/creator/upload.rst`.
* A few words about the distinction between `--all`, `--package`, `--query`, and using no option at all in `uploading_packages/uploading_to_remotes.rst`.

### DIFF
--- a/reference/commands/creator/upload.rst
+++ b/reference/commands/creator/upload.rst
@@ -6,8 +6,8 @@ conan upload
 
 .. code-block:: bash
 
-    $ conan upload [-h] [-p PACKAGE] [-r REMOTE] [--all] [--skip-upload]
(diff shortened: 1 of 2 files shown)


markgalpin:
I think its important to make explicit that the complex string for the query needs to be in quotes, you have it nicely in the example in the reference, but nothing in the "uploading packages" page.

joechrisellis:
I can add something to make that clearer, no problem :)

memsharded:
Thanks for the docs, they are good! I changed the PR to the docs "develop" branch, as it is not a released feature yet. When 1.7 is out, the docs develop branch is merged to master and published.

Issue #780, opened 2018-08-24
`short_paths` documentation misleading


liberforce:
The [documentation for `short_paths`](https://docs.conan.io/en/latest/changelog.html?highlight=conan_user_home_short#id19) says:

> If one of the packages you are creating hits the limit of 260 chars path length in Windows, add short_paths=True in your conanfile.py

But in fact Conan enables `short_paths` by default by on Windows since [Conan 0.30.1](https://docs.conan.io/en/latest/changelog.html?highlight=conan_user_home_short#id19).

I had enabled long paths on Windows Server 2016, and found that weird that I had a `C:\.conan` dir poping, even if I didn't have `short_paths=True` anywhere.

Furthermore, the doc for [CONAN_USER_HOME_SHORT](https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_user_home_short#conan-user-home-short) states that:

> Please note that this only works with Python 3.6 and newer.

This was added on commit 54a28e57954494b0913ee39d9f68505d8660a60a.
However, I tested successfully Conan (all versions since 1.0 to the last to this day) on Windows 2016 Server on AWS with a [python 2.7.15 installed though chocolatey](https://chocolatey.org/packages/python2). Long paths work nicely, so I think the [changes made for python 3.6 to support long paths](https://hg.python.org/cpython/rev/26601191b368) may have been backported to 2.7 (it's just a change in the manifest). It works nicely with no workaroundand, and `CONAN_USER_HOME_SHORT=None` solved my problem.

To sum up:
1. `short_paths=True` being the default on Windows should be in the `short_paths` documentation and not only in the changelog. Maybe you should add uses cases of when this is really required, as in the Windows case it seems not useful because of the default setting.
2. There should be a sentence saying it can be disabled using `CONAN_USER_HOME_SHORT`, and a link to that environment variable.
3. Long paths work on Windows Server 2016 too, not only Windows 10
4. When enabled system-wide, long paths work with python >= 3.6 and (at least) python>= 2.7.15. That mentions appear twice, so it should be fixed in `reference/conanfile/attributes.rst` and `reference/env_vars.rst`.

liberforce:
The fact that this works on all up-to-date python versions (2.7.x or 3.x) seems comfirmed by [this comment](https://github.com/conan-io/conan/issues/997#issuecomment-281839477) as well as those [first 3 comments](https://bugs.python.org/issue27731#msg272385).

liberforce:
[...]

PR #786, opened 2018-08-28
Windows 10 & 2016 Server MAX_PATH limit can be bypassed even with python 2.7

liberforce:
MAX_PATH fixes for #780.


liberforce:
Broken CI fixed in #787 .

PR #790, opened 2018-08-28
New and improved generator for B2 (Boost Build).

grafikrobot:


integrations/boost_build.rst
```diff
@@ -2,6 +2,11 @@
 |boost_build_logo| Boost Build
 ______________________________
 
+.. container:: out_reference_box
```
danimtb: Don't know if this is a highlighted box in restructuredtext. Maybe use ``.. important::`` ?

reference/generators/b2.rst
```diff
@@ -0,0 +1,218 @@
+.. _b2_generator:
+
+`B2`
+====
+
+.. container:: out_reference_box
```
danimtb: same here

reference/generators/b2.rst
```diff
@@ -0,0 +1,218 @@
...
+    This is the reference page for the ``B2``, aka *Boost Build*, generator. It is
+    a multi-generator to match the multi-build nature of ``B2``.
+
+Usage
+-----
+
+.. code-block:: bash
+
+    $ conan install -g b2 -s build_type=Release ...
+    $ conan install -g b2 -s build_type=Debug ...
```
danimtb: Does this generator need both commands to be executed for it to work?

reference/generators/b2.rst
grafikrobot: I was following what other generator references do. For example: https://raw.githubusercontent.com/conan-io/docs/develop/reference/generators/cmake.rst

reference/generators/b2.rst
grafikrobot: No, like the cmake multi, only one is needed. I added some comments to clarify that though.

integrations/boost_build.rst
grafikrobot: Changed it to caution as that expresses the sentiment more accurately. Note.. I'm a RST noob. So I was copying from other Conan docs what seemed appropriate.

reference/generators/b2.rst
danimtb: ok, that's fine!

PR #793, opened 2018-08-29
Feature/python requires

memsharded:


lasote:
What about the current howto? https://docs.conan.io/en/latest/howtos/python_code_reuse.html

PR #794, opened 2018-08-29
Fix "let's" spelling

liberforce:


memsharded:
Hi @liberforce 

Nice fixes, many thanks. I think this can go live, so it can be done directly to master, and not wait until the next conan release to be published. Could you please change your PR to master, please?

liberforce:
Here we go...

PR #805, opened 2018-09-04
Add spell checking infrastructure and fix a massive amount of typos

liberforce:


danimtb:
Great addition @liberforce!
I think it would be nice to run this in CI and error if there is any mistake.
Does the spellcheck errors if there are mistakes or just creates the output file?

We would have to check that the output file is not included in the deployment stage.
Thanks!

liberforce:
I didn't want something too disruptive for now, so it uses the default behavior: it returns warnings, in the output file as well as stdout/stderr. Those are warnings only, not errors, so they don't block the build. The thing is that it's quite easy to break the check with false positives, for example by naming a user, typing paths or citing some class from a code example.

When it breaks:
- if it's really a typo, you can fix it
- if it's a false positive seldom used, you may disable it locally on the file using the [spelling directive](https://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html#private-dictionaries)
- if it's a false positive widely used, add it to the global dictionary in `spelling_wordlist.txt`

So errors are easy to fix, but they may frighten some people wanting to contribute. There's also some people that are (wrongly) sure a word they wrote is correct might add it to the list to pass the checks instead of fixing the error. Therefore, I think looking at the warnings in the travis logs every now and then, or running the spellcheck before a release would be enough. This also gives more control on the words whitelisted in `spelling_wordlist.txt`. This doesn't prevent contributors to check spelling before submitting, it's just that it would not be mandatory.

For deployment, the code does this:

```
copytree("_build/html/", tmp_dir)
shutil.copy2("_build/latex/conan.pdf", tmp_dir)
shutil.rmtree("_build")
```

So stuff in `_build` is all ignored except for html and pdf output.

danimtb:
Nice clarification. I think is enough to output spell warnings.

Could you please include the `make spelling` call in the CI script for both deploy and not deploy cases?

Thanks!

liberforce:
Yeah, that's on the way. I'm just trying to understand how I can avoid stupid false positives like "doesn" in "doesn't" or "ve" in "we've".

memsharded:
Yes, please, do not introduce errors on spelling, as false positives will be too annoying. It is enough to WARN about them. Otherwise, this PR looks a very good contribution, great job, thanks!

liberforce:
Thanks. I was tired of having a "typos" branch and submitting a PR every now and then, just to find more typos a few hours after when reading another page of documentation. This will make it easier to fix all the typos at once.

liberforce:
Should be good now. All the warnings have been fixed or silenced, and the spellchecker is run on deploy too like @danimtb asked.

PR #817, opened 2018-09-11
Document CONAN_EXPORTED

### LINKED ISSUE #518: Document CONAN_EXPORTED
set by the CMake helper, what is the use case?

https://github.com/conan-io/conan/issues/2408

jgsogo:
Close https://github.com/conan-io/conan/issues/2408
Close #518


jgsogo:
Yes, I think it is more like a `bugfix` than something new, but I didn't know what is the fastest path to deployment 👍

PR #820, opened 2018-09-13
Document ConanInvalidConfiguration, fix some indentation issues

jgsogo:
Documentation for https://github.com/conan-io/conan/pull/3517


danimtb:
I think this worths a mention in the `configure()` method from reference:
https://docs.conan.io/en/latest/reference/conanfile/methods.html#configure-config-options

Please explain explicitly there that the configure() should be used to raise when there is a configuration not supported and illustrate it with an example.

danimtb:
Missing changelog.

Will probably conflict with #821. Let's see which one is merged first ⚔️

lasote:
@danimtb you lose

PR #828, opened 2018-09-19
Added depdenencies example and build folder to meson helper

### LINKED ISSUE #807: Meson example should show dependency consumption within `meson.build` file
I'd like to suggest an improvement to the documentation on this page:  https://docs.conan.io/en/latest/reference/build_helpers/meson.html

So the magic with conan is that I am getting some `*.pc` files generated automatically, but the current example does not show how to consume the `*.pc` dependencies from the `meson.build` file.  I think that is the most important point, but the example doesn't cover it.

For example, if my conanfile.py depends on Boost and Poco ...

```python
from conans import ConanFile, tools, Meson
import os


class PanoptoBase(ConanFile):
    default_options = \
        "boost:shared=False"

    exports_sources = "include/*"

    generators = "pkg_config"

    requires = \
        "Poco/1.8.0.1@pocoproject/stable", \
        "boost/1.66.0@conan/stable"

    settings = \
        "os", \
        "compiler", \
        "build_type"

    def build(self):
        meson = Meson(self)
        meson.configure(build_folder="./build", pkg_config_paths="./build")
        meson.build()
```

... then I will have 5 `*.pc` files after I execute `conan install ..` shown here:

```
* boost.pc
* bzip2.pc
* OpenSSL.pc
* Poco.pc
* zlib.pc
```

So then the `meson.build` file should show how I suck those dependencies in from the `*.pc` files as follows:

```
project(
    'Whatever.Lib',
    'cpp',
    version: '0.0.0',
    meson_version: '>= 0.47.0',
    default_options: [ 'buildtype=debug',
                       'cpp_std=c++11',
                       'warning_level=1']
)

includeFolder = include_directories('include')

static_library(
    'package_name',
    [   'src/whatever.cpp',
    ],
    include_directories: includeFolder,
    dependencies:
    [   dependency('boost'),
        dependency('bzip2'),
        dependency('OpenSSL'),
        dependency('Poco'),
        dependency('zlib')
    ]
)
```

danimtb:
closes #807

### DIFF
--- a/reference/build_helpers/meson.rst
+++ b/reference/build_helpers/meson.rst
@@ -4,9 +4,8 @@ Meson
 =====
 
 If you are using **Meson Build** as your build system, you can use the **Meson** build helper.
-Specially useful with the :ref:`pkg_config generator<pkg_config_generator>` that will generate the ``*.pc``
+Specially useful with the :ref:`pkg_config_generator` that will generate the ``*.pc``
 files of our requirements, then ``Meson()`` build helper will locate them automatically.
-
 
 .. code-block:: python
    :emphasize-lines: 5, 10, 11, 12
@@ -21,9 +20,8 @@ files of our requirements, then ``Meson()`` build helper will locate them automa
 
         def build(self):
             meson = Meson(self)
-            meson.configure()
+            meson.configure(build_folder="build")
             meson.build()
-
 
 Constructor
 -----------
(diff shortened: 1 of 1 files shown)


reference/build_helpers/meson.rst
```diff
@@ -4,10 +4,9 @@ Meson
 =====
 
 If you are using **Meson Build** as your build system, you can use the **Meson** build helper.
-Specially useful with the :ref:`pkg_config generator<pkg_config_generator>` that will generate the ``*.pc``
+Specially useful with the :ref:`pkg_config_generator` that will generate the ``*.pc``
```
jgsogo: ``*.pc`` >> *.pc*

PR #834, opened 2018-09-21
Setting conditional default options

### LINKED ISSUE #816: Document conditional default_options
From https://github.com/conan-io/conan/issues/3519

```python
    def config_options(self):
        if self.options.some_option == None:
            if self.settings.os == 'Android':
                self.options.some_option = True
            else:
                self.options.some_option = False
```
This works, both for overriding from command line and to defined the default if not defined in command line.

danimtb:
Separated attribute sections of ``options`` and ``config_options``.
Cross-referenced sections.

closes #816

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -358,14 +358,14 @@ And it also works with default option values of conditional required dependencie
 For this example running in Windows, the `default_options` for the `Pkg/0.1@user/channel` will be ignored, they will only be used on every
 other OS.
 
-You can even make default options conditional using ``config_options()``:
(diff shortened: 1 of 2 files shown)


garethsb-sony:
Using `config_options`, we can conditionally remove options, and using the `== None` approach, we can effectively conditionally set a default value. (I assume we can do this for individual options by omitting them from `default_options` as well as by not defining `default_options` as a whole in the way demonstrated in the new example?)
What is the best way to conditionally control the set of allowed values? Do we have to list all of them in `options`, and then `raise ConanInvalidConfiguration` in `configure`, or can we do anything in `config_options`? E.g. if certain values for a particular option were only allowed on one OS?

PR #835, opened 2018-09-21
Improved tools.collect_libs() description

### LINKED ISSUE #781: Put note about `collect_libs` special handling libraries for LD
The implementation of `collect_libs` has some helpful path handling specific to `.a` files , however there doesn't seem to be any note about what it's doing in the docs: 

https://docs.conan.io/en/latest/reference/tools.html#tools-collect-libs

```
def collect_libs(conanfile, folder="lib"):
    if not conanfile.package_folder:
        return []
    lib_folder = os.path.join(conanfile.package_folder, folder)
    if not os.path.exists(lib_folder):
        conanfile.output.warn("Lib folder doesn't exist, can't collect libraries: {0}".format(lib_folder))
        return []
    files = os.listdir(lib_folder)
    result = []
    for f in files:
        name, ext = os.path.splitext(f)
        if ext in (".so", ".lib", ".a", ".dylib"):
            if ext != ".lib" and name.startswith("lib"):
                name = name[3:]
            result.append(name)
    return result
```

danimtb:
closes #781 

cc/ @solvingj

### DIFF
--- a/reference/tools.rst
+++ b/reference/tools.rst
@@ -1153,8 +1153,8 @@ tools.collect_libs()
 
     def collect_libs(conanfile, folder="lib")
 
-Fetches a list of all libraries in the package folder. Useful to collect not inter-dependent
-libraries or with complex names like ``libmylib-x86-debug-en.lib``.
+Generates a library names from library files with extensions *.so*, *.lib*, *.a* and *.dylib* located in the package folder. Useful to
+collect not inter-dependent libraries or with complex names like ``libmylib-x86-debug-en.lib``.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


reference/tools.rst
```diff
@@ -1153,8 +1153,8 @@ tools.collect_libs()
 
     def collect_libs(conanfile, folder="lib")
 
-Fetches a list of all libraries in the package folder. Useful to collect not inter-dependent
-libraries or with complex names like ``libmylib-x86-debug-en.lib``.
+Generates a library names from library files with extensions *.so*, *.lib*, *.a* and *.dylib* located in the package folder. Useful to
```
jgsogo: I think that it is better:

```rst
Returns a list with the name of the libraries (files with extensions *.so*, *.lib*, *.a* and *.dylib*) 
located in the package folder inside the ``folder`` directory. Useful [..]
```

danimtb:
Thanks for the fix ☺️

PR #846, opened 2018-09-25
Plugins documentation

danimtb:
- Created new "extending" section as suggested by @jgsogo 
- Moved python requires.
- Created a plugin section in "extending" with examples and more user focused usage.
- Created a plugin section in reference with plugin interface: functions and parameters.
- Mention policy of *plugins* folder in a **conan config install**


extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+                metadata_error = True
+                output.error("%s Conanfile doesn't have '%s'. It is recommended to add it as attribute"
+                            % (test, field))
+
+    def pre_source(output, conanfile, conanfile_path, **kwargs):
+        conanfile_content = tools.load(conanfile_path)
+        if "def source(self):" in conanfile_content:
+            test = "[IMMUTABLE SOURCES]"
+            valid_content = [".zip", ".tar", ".tgz", ".tbz2", ".txz"]
+            invalid_content = ["git checkout master", "git checkout devel", "git chekcout develop"]
```
lasote: I'm sure `chekcout` is invalid. Is this on purpose?

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+.. code-block:: python
+   :caption: *signing_plugin.py*
+
+    import os
+    from conans import tools
+
+
+    def pre_upload_package(output, conanfile_path, reference, package_id, remote, **kwargs):
+        package_path = os.path.abspath(os.path.join(os.path.dirname(conanfile_path), "..", "package", package_id, "conan_package.tgz"))
+        my_signing_package_function(package_path)
```
lasote: suggest modifying the package is not a good idea.

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+        # Check basic meta-data
+        for field in ["url", "license", "description"]:
+            field_value = getattr(conanfile, field, None)
+            if not field_value:
+                output.warn("Conanfile doesn't have '%s'. It is recommended to add it as attribute"
+                            % field)
+
+This plugin comes activated by default.
+
+Conan Center plugin
```
lasote: I wouldn't mention the conan center plugin until we really use it and complete it.

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+=======
+
+.. warning::
+
+    This is an EXPERIMENTAL feature released in 1.8. Subject to breaking changes.
+
+The Conan plugins are Python functions that are intended to extend the Conan functionalities and let users customize the client behavior at
+determined execution points.
+
+Storage, activation and sharing
```
lasote: I think this section or the information contained. shouldn't be in the reference but the extending/plugins.

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+
+Some of this parameters does not appear in the signature of the function as they may not be available always (Mostly depending on the recipe
+living in the local cache or in user space). However, they can be checked with the ``kwargs`` parameter.
+
+.. important::
+
+    Plugin functions should have a ``**kwargs`` parameter to keep compatibility of new parameters that may be introduced in future versions
+    of Conan.
+
+Importing from a module
```
lasote: Same, this is not reference. Move to the other section

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+.. code-block:: python
+   :caption: *example_plugin.py*
+
+    from conans import tools
+
+
+    def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+        conanfile_content = tools.load(conanfile_path)
+        test = "[RECIPE METADATA]"
+        metadata_error = False
```
jgsogo: variable `metadata_error` is not needed.

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+
+.. code-block:: python
+   :caption: *example_plugin.py*
+
+    from conans import tools
+
+
+    def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+        conanfile_content = tools.load(conanfile_path)
+        test = "[RECIPE METADATA]"
```
jgsogo: Use the recipe reference instead of (or added to) this hardcoded string:

```python
test = str(reference)  #  Would resolve to 'name/version@user/channel'
```

or do something with the `reference.name`, `reference.version`,...

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+            field_value = getattr(conanfile, field, None)
+            if not field_value:
+                metadata_error = True
+                output.error("%s Conanfile doesn't have '%s'. It is recommended to add it as attribute"
+                            % (test, field))
+
+    def pre_source(output, conanfile, conanfile_path, **kwargs):
+        conanfile_content = tools.load(conanfile_path)
+        if "def source(self):" in conanfile_content:
+            test = "[IMMUTABLE SOURCES]"
```
jgsogo: Idem. Use recipe reference somewhere

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+Here you can see an example of a simple plugin:
+
+.. code-block:: python
+   :caption: *example_plugin.py*
+
+    from conans import tools
+
+
+    def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+        conanfile_content = tools.load(conanfile_path)
```
jgsogo: No need for `conanfile_content`

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+
+This plugin is only checking the recipe content prior to the recipe being exported and prior to downloading the sources. Basically the
+``pre_export()`` function is checking the attributes of the ``conanfile`` object to see if there is an URL, a license and a description and
+warning the user with a message through the ``output``. This is done **before** the recipe is exported to the local cache.
+
+The ``pre_source()`` function checks if the recipe contains a ``source()`` method (this time it is using the conanfile content instead of
+the ``conanfile`` object) and in that case it checks if the download of the sources are likely coming from immutable places (a compressed
+file or a determined :command:`git checkout`). This is done **before** the **source()** method of the recipe is called.
+
+Any kind of Python scripting can be executed. You can create global functions and called them from different plugin functions, import them
```
jgsogo: `called` >> `call`

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+from a relative module and warn, error or even raise to block the Conan client execution.
```
jgsogo: `block` >> `abort`/`stop`

extending/plugins.rst
jgsogo: Maybe we can live just with the `post_download_package` example

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+        my_signing_package_function(package_path)
+
+    def post_download_package(output, conanfile_path, reference, package_id, remote, **kwargs):
+        package_path = os.path.abspath(os.path.join(os.path.dirname(conanfile_path), "..", "package", package_id, "conan_package.tgz"))
+        my_sign_checker_function(package_path)
+
+Official Plugins
+----------------
+
+There are two official plugins ready to be used in Conan. You could take as a starting point to create your own ones.
```
jgsogo: Link to the repo? All plugins in the official repo will be official?

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+
+.. code-block:: python
+   :caption: *attribute_checker.py*
+
+    def pre_export(output, conanfile, conanfile_path, reference, **kwargs):
+        # Check basic meta-data
+        for field in ["url", "license", "description"]:
+            field_value = getattr(conanfile, field, None)
+            if not field_value:
+                output.warn("Conanfile doesn't have '%s'. It is recommended to add it as attribute"
```
jgsogo: I think that this message should contain the full reference of the recipe... or maybe the output is scoped with it?

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+
+This plugin has been created to perform some the checks that the Conan team make as part of the process of accepting a new library into the
+:ref:`conan_center` central repository in Bintray.
+
+This plugin is not shipped with the Conan plugin but stored in a different repository to improve it separated from the Conan source code.
+
+The plugin performs various checks during development of a package and also during the creation and it has been designed to not
+block the Conan client execution and only printing error traces.
+
+.. tip::
```
jgsogo: I would move this tip to the end or to the beginning of this section, it is like a _to see a full list of available plugins go there_

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+
+They can be easily activated and deactivated from the command line using the :command:`conan config set` command:
+
+.. code-block:: bash
+
+    $ conan config set plugins.attribute_checker  # Activates 'attribute_checker'
+
+    $ conan config rm plugins.attribute_checker  # Deactivates 'attribute_checker'
+
+There is also an environment variable ``CONAN_PLUGINS`` to list the active plugins. Plugins listed in *conan.conf* will be loaded into
```
jgsogo: I can't understand what will happen if I have some plugins listed in `conan.conf` and a different set in `CONAN_PLUGINS``.

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+    **Pre** and **post** ``download_recipe()``/``download_package()`` are also executed when installing new recipes/packages from remotes
+    using :command:`conan create` or :command:`conan install`.
+
+Function parameters
+--------------------
+
+Here you can find the description for each parameter:
+
+- **output**: :ref:`Output object<conanfile_output>` to print formatted messages during execution with the name of the plugin and the
+  function executed, e.g., ``[PLUGIN - complete_plugin] post_download_package(): This is the remote name: default``.
```
jgsogo: Maybe add the conan reference to it?  (and it will invalidate some of my comments above)

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+
+- **conanfile**: It is a regular ``ConanFile`` object loaded from the recipe that received the Conan command. It has its normal attributes
+  and dynamic objects such as ``build_folder``, ``package_folder``...
+
+- **conanfile_path**: Path to the *conanfile.py* file whether it is in local cache or in user space.
+
+- **reference**: Named tuple with attributes ``name``, ``version``, ``user, and ``channel``.
```
jgsogo: ...and the `full_repr` or the `str`, it has been used above.

reference/plugins.rst
```diff
@@ -0,0 +1,282 @@
...
+Table legend:
+  - **Yes**: Availability in ``pre`` and ``post`` functions in any context.
+  - **No**: Not available.
+  - **pre / post**: Availability in both ``pre`` and ``post`` functions with **different values**. e.g. ``conanfile_path`` pointing to user
+    space in ``pre`` and to local cache in ``post``.
+  - **post**: Only available in ``post`` function.
+  - **cache**: Only available when the context of the command executed is the local cache. e.g. :command:`conan create`,
+    :command:`conan install`...
+  - **user space**: Only available when the context of the command executed is the local cache. e.g. :command:`conan build`
+
```
jgsogo: I wonder if this table is needed given the full example provided above.

extending/plugins.rst
```diff
@@ -0,0 +1,134 @@
...
+                        fixed_sources = False
+            else:
+                fixed_sources = False
+                for valid in valid_content:
+                    if valid in conanfile_content:
+                        fixed_sources = True
+
+            if not fixed_sources:
+                output.error("%s Source files does not come from and immutable place. Checkout to a "
+                            "commit/tag or download a compressed source file" % test)
```
jgsogo: I find this `pre_source` example quite difficult. I like the intention, but I think it is very tricky

extending/plugins.rst
danimtb: I'd like to keep it because it could be used as a reference to start creating a plugin. I have added a warning saying this is something under development

reference/plugins.rst
danimtb: I consider it as a summary for clarification, I like to have the "usage" example above and the explanation in the table

jgsogo: I was thinking about maintainability, we have to keep code and the documentation (in two places) synchronized. But it is ok, let's see if in the future this table grows too much.

reference/plugins.rst
danimtb: Plugin functions might not always attached to a reference or recipe, for example in the future we could add a function to be executed right after a Conan command

reference/plugins.rst
danimtb: Will try to stick it to the ``str(reference)`` to avoid documenting the class

extending.rst
```diff
@@ -0,0 +1,12 @@
...
+
+Extending Conan
+---------------
+
+This section provides an introduction to extension capabilities of Conan:
+
+.. toctree::
+   :maxdepth: 2
+
+   extending/python_requires
```
memsharded: Interesing section. I think I'd also move here (not in this PR) the "custom generators" section. Maybe also a good place to document the python API in the future.

extending/plugins.rst
memsharded: Agree, would be better to keep the example more simple and obvious to read.

PR #856, opened 2018-09-27
Fix typo

ssrobins:


memsharded:
Thanks for fixing this typo! 
For fixes that are in the current "live" documentation, you could do the PR to master branch, instead of develop, so it gets published immediately on merging. This typo fix will be included in the next release. Thanks again!

ssrobins:
Good to know, I'll do that in the future.

-Steve

PR #859, opened 2018-10-01
Package Dev Flow: Fix channel incosistency

Talkless:
Package Dev Flow documentation mostly uses `user/channel` specification, though it's `user/testing` in some places. 

This patch fixes this inconsistency.


danimtb:
Same situation as the one described in #860

PR #866, opened 2018-10-08
conan inspect

memsharded:


reference/commands/misc/inspect.rst
```diff
@@ -0,0 +1,52 @@
...
+conan inspect
+=============
+
+.. note::
+
+    This is an EXPERIMENTAL new command in Conan 1.8, and might be subject to changes in future releases.
+
+.. code-block:: bash
+
+    $ .. code-block:: bash
```
danimtb: remove this line

reference/commands/misc/inspect.rst
```diff
@@ -0,0 +1,52 @@
...
+.. code-block:: bash
+
+    $ conan inspect zlib/1.2.11@conan/stable -a=name -a=version -a=options -a default_options -r=conan-center
+    name: zlib
+    version: 1.2.11
+    options
+        shared: [True, False]
+    default_options: shared=False
+
+.. code-block:: bash
```
danimtb: I will include an example with the output of an recipe inspection stored in a remote to show it actually downloads the recipes

uilianries: What happens when no attribute is specified? could it be a good example?

uilianries: thanks!

memsharded: Done:
- Removed the extra code-block
- Fixed command, so it outputs some default fields in case no ``-a`` is specified.

@danimtb An example of the output of a remote is already shown

PR #871, opened 2018-10-09
Fix sentence

Croydon:


memsharded:
Hi @Croydon 

Thanks for fixing this! As it is something that can be published right ahead, you might want to do the PR to the ``master`` branch instead. Please change it. It will also fire a new CI, which failed for some network connection, nothing related to your change. Best!

Croydon:
Done

PR #876, opened 2018-10-10
Recovered 'Edit on GitHub' button

### LINKED ISSUE #873: Docs used to have an "Edit" button in top right, now gone
I don't know when it changed, a while ago I think, but it was a 1 click operation to jump to creating a PR for a page while browsing that page. I can't find that button anymore, there's no only "view page source" which isn't helpful.  Am I missing something?

danimtb:
closes #873

Information is in this link: https://docs.readthedocs.io/en/latest/vcs.html#github

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -59,7 +59,12 @@ versions_dict = data
 
 html_context = {
     "versions": versions_dict,
-    "current_version": version
+    "current_version": version,
+    "display_github": True, # Integrate GitHub
+    "github_user": "conan-io", # Username
+    "github_repo": "docs", # Repo name
+    "github_version": "master", # Version
+    "conf_py_path": "/" # Path in the checkout to the docs root
 }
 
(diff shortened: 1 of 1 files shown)


conf.py
```diff
@@ -59,7 +59,12 @@
 
 html_context = {
     "versions": versions_dict,
-    "current_version": version
+    "current_version": version,
+    "display_github": True, # Integrate GitHub
+    "github_user": "conan-io", # Username
+    "github_repo": "docs", # Repo name
+    "github_version": "master", # Version
```
danimtb: Maybe we should put "develop" here as @lasote thoughts on following the gitflow

danimtb:
image

conf.py
memsharded: No, because most edits done in the Github button, are to the master/production branch.

PR #878, opened 2018-10-10
Minor changes regarding ConanInvalidConfiguration

jgsogo:
Do you think we should document exit error codes?


danimtb:
> Do you think we should document exit error codes?

I think it would be something useful for custom tools like CPT. My only concern is documenting this means it is "stable", but open an issue and we can discuss it there

PR #882, opened 2018-10-10
Use default_options as a dictionary in reference too

danimtb:


reference/conanfile/attributes.rst
```diff
@@ -366,21 +366,21 @@ consistent implementation take into account these considerations:
 default_options
 ---------------
 
-As you have seen in the examples above, recipe's default options can be assigned to the desired value. However, you can also specify
-default option values of the required dependencies:
+As you have seen in the examples above, recipe's default options are declared as a dictionary and can be assigned an initial desired value.
```
jgsogo: default_options ARE the "initial desired value" for the options... 

``As you have seen in the examples above, recipe's default options are declared as a dictionary with the initial desired value of the options``

or 

``As you have seen in the examples above, recipe's default options are declared as a dictionary.``

PR #895, opened 2018-10-16
 Update short_paths documentation

### LINKED ISSUE #780: `short_paths` documentation misleading

liberforce:
Fixes #780.
Fixes some spelling issues too.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -36,7 +36,7 @@ Check https://github.com/conan-io/conan for issues and more details about develo
 
 - Fix: Remove warnings related to ``python_requires()``, both in linter and due to Python2. `#3706 <https://github.com/conan-io/conan/pull/3706>`_
 - Fix: Use *share* folder for ``DATAROOTDIR`` in CMake and AutoTools build helpers. `#3705 <https://github.com/conan-io/conan/pull/3705>`_
-- Fix: Disabled apiv2 until the new protocol becomes stable. `#3703 <https://github.com/conan-io/conan/pull/3703>`_
+- Fix: Disabled `apiv2` until the new protocol becomes stable. `#3703 <https://github.com/conan-io/conan/pull/3703>`_
 
 
 1.8.0 (9-October-2018)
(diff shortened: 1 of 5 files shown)


danimtb:
Hi @liberforce and thanks a lot for the PR.

I think it is fine and could be merged. Could you please resolve the conflicts?

danimtb:
Done!

PR #896, opened 2018-10-16
- update premake documentation

SSE4:
https://github.com/conan-io/conan/pull/3751


SSE4:
I've removed old way and added reference page

integrations/premake.rst
```diff
@@ -1,32 +1,103 @@
...
 
-Now you can use this generator as a requirement in your recipes **but also as a generator**:
+Now let's get the premake project and build it:
 
-.. code-block:: text
+.. code-block:: bash
 
-    [requires]
-    PremakeGen@0.1@memsharded/testing
+    $ git clone https://github.com/SSE4/conan-premake-demo.git
```
danimtb: Having a look at the project and the need to add two addtional remotes to run the example, I think it is too much for a premake example.

- First: If a user wants to use premake, it will have premake already installed, so no need to add premake_installer as build require
- Second: Conan community should not be added as a repository, it only exists for maturing packages
- Third: Avoid using scripts like `./run.cmd`, is better to use eveything explicitly and with clear steps

danimtb:
@SSE4 I have added a built-in example for premake.
It is not fully covered from end to end but it illustrates well the purpose of the generator.

We could improve it (specially the lua script) in the future! Thanks

PR #898, opened 2018-10-17
export-pkg using profile or settings by default

memsharded:
Docs for https://github.com/conan-io/conan/pull/3768

### DIFF
--- a/reference/commands/creator/export-pkg.rst
+++ b/reference/commands/creator/export-pkg.rst
@@ -176,8 +176,14 @@ with the ``export-pkg`` command. Check this :ref:`How to package existing binari
       $ conan source . --source-folder src
       $ conan install . --install-folder build_x86 -s arch=x86
       $ conan build . --build-folder build_x86 --source-folder src
-      $ conan export-pkg . Hello/0.1@user/stable --build-folder build_x86
+      $ conan export-pkg . Hello/0.1@user/stable --build-folder build_x86 -s arch=x86
 
-  In this case, in the :command:`conan export-pkg`, you don't need to specify the :command:`-s arch=x86` or any other setting, option, or profile,
-  because it will all the information in the :command:`--build-folder` the *conaninfo.txt* and *conanbuildinfo.txt`* that have been created with
-  :command:`conan install`.
+
+.. note::
+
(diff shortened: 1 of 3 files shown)


reference/commands/creator/export-pkg.rst
```diff
@@ -176,8 +176,14 @@ with the ``export-pkg`` command. Check this :ref:`How to package existing binari
       $ conan source . --source-folder src
       $ conan install . --install-folder build_x86 -s arch=x86
       $ conan build . --build-folder build_x86 --source-folder src
-      $ conan export-pkg . Hello/0.1@user/stable --build-folder build_x86
+      $ conan export-pkg . Hello/0.1@user/stable --build-folder build_x86 -s arch=x86
```
danimtb: Should we recommend this as default or only for special cases as warned below?

reference/commands/creator/export-pkg.rst
```diff
@@ -176,8 +176,14 @@ with the ``export-pkg`` command. Check this :ref:`How to package existing binari
...
 
-  In this case, in the :command:`conan export-pkg`, you don't need to specify the :command:`-s arch=x86` or any other setting, option, or profile,
-  because it will all the information in the :command:`--build-folder` the *conaninfo.txt* and *conanbuildinfo.txt`* that have been created with
-  :command:`conan install`.
+
+.. note::
+
+  The :command:`conan export-pkg`, if no ``--profile`` (or settings, options, etc) arguments are provided,
```
danimtb: Please rephrase and add a command line example:

```
This command tries to obtain the configuration from the previous installed configuration
(generated with conan install) stored in *conaninfo.txt* file inside the install folder. There are
some corner cases where that information is not enough (like header only libraries with 
dependencies) and it is recommended to use --profile or --settings and --options parameters:

$ conan export-pkg . Hello/0.1@user/stable --profile myprofile
```

reference/commands/creator/export-pkg.rst
memsharded: This is the default one, it is everywhere in the docs. There is no export-pkg without using settings or profile anymore.

PR #900, opened 2018-10-17
Fix typos in the plugins doc

liberforce:


memsharded:
Thanks for fixing this!
Also, remember, if they are typos, they can directly go live immediately, no need to wait for next release (this will be in 1.9). You can do those pull requests branching from "master" branch and to "master". Thanks again!

liberforce:
Oh, right, I overlooked that. Thanks for the reminder.

PR #905, opened 2018-10-19
Proposed doc changes

ssrobins:
Hello,

Here are some fixes for things I ran into while starting to learn conan.  I still got great value from the docs as-is, but this might help a little for the next user.  Thanks for all the hard work!

-Steve

### DIFF
--- a/creating_packages/getting_started.rst
+++ b/creating_packages/getting_started.rst
@@ -34,8 +34,8 @@ This will generate the following files:
 
     conanfile.py
     test_package
-      conanfile.py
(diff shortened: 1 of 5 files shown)


ssrobins:
I can change that line back tonight (10 hours-ish from now).

ssrobins:
Ok, that's all the doc changes I have for now.  Things are green, anything else to wrap up?

memsharded:
Thanks very much @ssrobins this is very nice! This will indeed help other users. Cheers!

PR #907, opened 2018-10-19
Fix a/an grammar uses

liberforce:
Rules for using one the other: https://www.wikihow.com/Sample/An-A-Usage
"A user" respects the same rules as "a union".


memsharded:
Thank you! Merging it directly to master, so it goes live immediately, not in the next release.

PR #934, opened 2018-10-31
Explicitly mention that remote name is uppercased

ttencate:
Responsible code: https://github.com/conan-io/conan/blob/d9a75eff01098338d4d242d8591a704e75bf3a9c/conans/client/userio.py#L105


danimtb:
many thanks! 😄

PR #945, opened 2018-11-08
Review of examples in cmake_paths and cmake_find_package

### LINKED ISSUE #944: cmake_find_package and cmake_paths documentation example of Zlib
Hi first of all i want to congratulate all of of you in conan about the great job that you are doing...

My issue is only a note, because the documentation about [cmake_find_package](https://docs.conan.io/en/latest/integrations/cmake/cmake_find_package_generator.html) and [cmake_paths](https://docs.conan.io/en/latest/integrations/cmake/cmake_paths_generator.html) is not entirely accurate because of the name of the zlib package, the conan name is lowercase and the the script in the CMake installation directory (/Modules) is uppercase (/usr/share/cmake-3.10/Modules/FindZLIB.cmake) (at least on Ubuntu 18) and the example is neither all uppercase neither all lowercase :)

So your example neither work with the Findzlib.cmake generated by conan when using cmake_find_package neither work with system FindZLIB.cmake because of the name differences (linux)

so a more correct example will be using a library/package with the same name.

Hope i can help a little about improving your documentation.

Side note:
in my example i used like this:
```
cmake_minimum_required(VERSION 3.0)
project(helloworld)
set(CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR} ${CMAKE_MODULE_PATH})
add_executable(helloworld hello.cpp)
find_package(ZLIB)
if(ZLIB_FOUND)
    message(STATUS "zlib was found")
    include_directories(${ZLIB_INCLUDE_DIRS})
    target_link_libraries(helloworld ${ZLIB_LIBRARIES})
else()
    message(STATUS "zlib was NOT found!!!!")
endif()
```
and using both generators, but we are using always the FindZLIB.cmake from the system and not the generated one from conan.

danimtb:
closes #944


danimtb:
Hi @nesc1! This PR should fix the documentation with the cnages you requested. Now the example is using the right casing in `find_package()`. Thanks!

lasote:
Solve conflicts and merge

PR #948, opened 2018-11-21
Fixing a target name typo at the cmake_find_package generator reference

VictoriousV:
Added a missing colon to the name of the target.


danimtb:
Thanks!

PR #951, opened 2018-11-23
Adds github settings file for portable labels

danimtb:
Has to be merged to develop branch for changes to take effect

After merge we need to activate the app here: https://probot.github.io/apps/settings/

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -1,0 +1,100 @@
+# https://github.com/apps/settings
+
+repository:
+  description: conan.io reStructuredText documentation
+  homepage: https://docs.conan.io
+  has_wiki: false
+  default_branch: develop
+
+# Labels: define labels for Issues and Pull Requests
+# colors from: https://htmlcolorcodes.com/color-chart/
+labels:
+  - name: "type: question"
+    color: FADBD8
+    old_name: "question"
+  - name: "type: engineering"
+    color: F5B7B1
+  - name: "type: feature"
+    color: EC7063
+    old_name: "feature"
+  - name: "type: look into"
+    color: CB4335
+    old_name: "look into"
+  - name: "type: bug"
+    color: 943126
+    old_name: "bug"
+
+  - name: "complex: low"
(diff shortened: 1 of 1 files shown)


danimtb:
CI was cancel as it is not affected by this PR

jgsogo:
Do you want to keep this file almost-equal for `conan-io/conan` and `conan-community/community` too, or it is just intended for this repo?

If the answer is the first one, then it is ok, otherwise I would add the `has_projects=False` too (I don't know if `has_downloads=False` is needed too).

.github/settings.yml
```diff
@@ -0,0 +1,100 @@
+# https://github.com/apps/settings
+
+repository:
```
jgsogo: What about the name?

```suggestion
repository:
  name: docs
```

danimtb: I think I am going to keep it as generic as possible to try to have the same settings for all repositories

jgsogo: and what about the `description` and `homepage`?

danimtb: Yes, I am going to remove them too

jgsogo:
What about this (https://github.com/probot/probot-config) to share configuration between repos? Will it work? Having a `conan-io/conan: .github/common-settings.yml` and add a reference (and specialize) to it from the `.github/settings.yml` file of every repo

danimtb:
I think we could at this in the future if this kind of settings start to grow. I don't see this labels changing very much in a short term and configs are easy enough to be copy-pasted 😄 

However, if you feel so we can start putting this common labels in conan-io/conan repo and the reuse them in here... I just wanted to start with a smaller repo like this one

danimtb:
I am the one doing it! 🐮 🤠

danimtb:
Done! Bu labels were not update 🤦‍♂️ Maybe we need to push a new change with the app now active... 🤔

jgsogo:
Sure, an empty commit to `develop` should be enough

PR #955, opened 2018-11-27
Make generator documentation

danimtb:
Documentation conan-io/conan#3773


danimtb:
PR conan-io/conan#4003

integrations/make.rst
```diff
@@ -0,0 +1,135 @@
...
+    LDFLAGS         += $(addprefix -L, $(CONAN_LIB_PATHS))
+    LIBS            += $(addprefix -l, $(CONAN_LIBS))
+
+
+    #----------------------------------------
+    #     Make Commands
+    #----------------------------------------
+
+    COMPILE_CXX_COMMAND         ?= \
+        g++ -c $(CXXFLAGS) $(CPPFLAGS) $< -o $@
```
solvingj: `$(CXXFLAGS) ` should just be `$(CFLAGS) ` since there is actually no variable called `$(CXXFLAGS)` in this example, nor is one propagated by Conan.  Conan only propagates a single variable called `CFLAGS`.  Technically, it doesn't matter here because there were no `CFLAGS` defined in this example, so both will result in an empty string, but we should definitely make the example correct.

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
+Variables per package. ``<PKG-NAME>`` placeholder is filled with the name of the Conan package.
+
++--------------------------------------+----------------------------------------------------------------------+
+| NAME                                 | VALUE                                                                |
++======================================+======================================================================+
+| CONAN_ROOT_<PKG-NAME>                | Absolute path to root package folder                                 |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_SYSROOT_<PKG-NAME>             | System root folder                                                   |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_INC_PATHS_<PKG-NAME>           | Headers folders                                                      |
```
solvingj: I've seen references to `CONAN_INC_PATHS` and `CONAN_INCLUDE_PATHS`.  Not sure what the final choice was, but need to make sure references are consistent everywhere.

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_LIB_PATHS_<PKG-NAME>           | Library folders                                                      |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_BIN_PATHS_<PKG-NAME>           | Binary folders                                                       |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_BUILD_PATHS_<PKG-NAME>         | Build folders                                                        |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_RES_PATHS_<PKG-NAME>           | Resources folders                                                    |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS_<PKG-NAME>                | Binary folders                                                       |
```
solvingj: Delete this entry, it looks like a copy/paste error.

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS_<PKG-NAME>                | Library names to link with                                           |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_DEFINES_<PKG-NAME>             | Library definitions                                                  |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_CFLAGS_<PKG-NAME>              | Library C flags                                                      |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_CPPFLAGS_<PKG-NAME>            | Library CXX flags                                                    |
```
solvingj: The variable name `CPPFLAGS` is misleading.  `CPPFLAGS` is actually short for "C Pre-Processor Flags", nothing to do with C++.  In GNU Make, there is a separate implicit variable called `CXXFLAGS` that we don't do anything with in Conan... that would be specifically for the C++ compiler.

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_RES_PATHS                | Aggregated resource folders                                          |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS                     | Aggregated library names to link with                                |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_DEFINES                  | Aggregated library definitions                                       |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_CFLAGS                   | Aggregated C flags                                                   |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_CPPFLAGS                 | Aggregated CXX flags                                                 |
```
solvingj: "CXX flags" should say "C Preprocessor flags"

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_BUILD_PATHS              | Aggregated build folders                                             |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_RES_PATHS                | Aggregated resource folders                                          |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS                     | Aggregated library names to link with                                |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_DEFINES                  | Aggregated library definitions                                       |
++--------------------------------+----------------------------------------------------------------------+
+| CONAN_CFLAGS                   | Aggregated C flags                                                   |
```
solvingj: Change "C flags" to "C Compiler flags" for clarity.

reference/generators/make.rst
```diff
@@ -0,0 +1,82 @@
...
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_RES_PATHS_<PKG-NAME>           | Resources folders                                                    |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS_<PKG-NAME>                | Binary folders                                                       |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_LIBS_<PKG-NAME>                | Library names to link with                                           |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_DEFINES_<PKG-NAME>             | Library definitions                                                  |
++--------------------------------------+----------------------------------------------------------------------+
+| CONAN_CFLAGS_<PKG-NAME>              | Library C flags                                                      |
```
solvingj: Change "C flags" to "C Compiler flags" for clarity.

danimtb:
Hi @solvingj and thanks a lot for the review.

I have tried to follow the rules defined here: https://docs.conan.io/en/latest/reference/build_helpers/autotools.html#environment-variables

With the information indicated here: https://docs.conan.io/en/latest/reference/conanfile/attributes.html#cpp-info

Following that information the mapping will be:

| ``cpp_info`` | *conanbuildinfo.mak* | *Makefile* |
|--------------|----------------------|------------|
| defines      | CONAN_DEFINES        | CPPFLAGS   |
| includedirs  | CONAN_INCLUDE_PATHS  | CPPFLAGS   |
| libdirs      | CONAN_LIB_PATHS      | LDFLAGS    |
| libs         | CONAN_LIBS           | LIBS       |
| cflags       | CONAN_CFLAGS         | CFLAGS     |
| cppflags     | CONAN_CPPFLAGS       | CXXFLAGS   |

As you see, I think it is better to maintain a similar name for the ``cpp_info`` -> ``generator`` conversion but use the Make variable naming for the example (``generator`` -> User side). WDYT?

The controversial part here is the ``cpp_info.cppflags`` that in my opinion it should have been called ``cpp_info.cxxflags``

solvingj:
Yes, I mostly agree with it.  There are still a few issues to work out though. 

**Autotools vs Make : LIBS vs LDLIBS**
Sadly, and for reasons I have never found, Autotools chose to use a new variable name for this purpose.  GNU Make originally used `LOADLIBES`, then changed to `LDLIBS` which is what is used today.  Autotools came along later and used the shortened variable name `LIBS`. In short, I think that using `LIBS` is appropriate for the AutoTools Helper, and using `LDLIBS` is appropriate for this generator. 

**CPP vs CXX**  
Thank you for clarifying.  I actually completely forgot about this naming mismatch between Conan and Make/Autotools. I remember seeing it last year, perhaps I blacked it out.  I guess I had been working so deeply with Make the past few months, when i came back to work on the generator I just lined things up the way that makes sense and never checked the docs. Sorry about that :(

Since it can never be fixed, I've gone ahead and updated my PR for the unit tests based on your feedback: 
https://github.com/conan-io/conan/pull/4018

So, after all that, this looks good :)

danimtb:
Thanks a lot for the feedback. I will come and update this PR with the variables in the test you proposed on conan-io/conan#4018 😃

PR #958, opened 2018-11-29
Description about build helpers and generators

### LINKED ISSUE #946: Missing: What are generators?
If I didn't miss anything there is no explanation what generators are and what they do? This should be added

https://docs.conan.io/en/latest/reference/generators.html

danimtb:
closes #946

### DIFF
--- a/reference/build_helpers.rst
+++ b/reference/build_helpers.rst
@@ -3,7 +3,8 @@
 Build helpers
 =============
 
-There are several helpers that can assist to automate the build() method for popular build systems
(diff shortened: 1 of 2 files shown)


reference/build_helpers.rst
```diff
@@ -3,7 +3,8 @@
 Build helpers
 =============
 
-There are several helpers that can assist to automate the build() method for popular build systems
+Build helpers are Python wrappers of a build tool that help with the conversion of the Conan settings to the build system's ones. They are
```
jgsogo:
```suggestion
Build helpers are Python wrappers of a build tool that help with the conversion of the Conan settings to the build system's ones. They
```

PR #959, opened 2018-11-29
Update install.json output

### LINKED ISSUE #906: Install json output example without 'dependency' field
Currently there is only a description of the 'dependency' parameter here: https://docs.conan.io/en/latest/reference/commands/output/install.html

But the output example does not have it

danimtb:
closes #906

### DIFF
--- a/reference/commands/output/install.rst
+++ b/reference/commands/output/install.rst
@@ -114,7 +114,10 @@ The output JSON contains a two first level keys:
                     }
                 ]
             },
-            ...
+            {"...":"..."
+            }
+        ]
+    }
 
 .. note::
 


reference/commands/output/install.rst
```diff
@@ -50,69 +50,123 @@ The output JSON contains a two first level keys:
...
+                            "version":"1.0.2l",
+                            "description":"OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols",
+                            "public_deps":[
+                                "zlib"
+                            ]
+                        }
+                    }
+                ]
+            },
+            {
```
jgsogo: I will substitute the second element with `...`, is there anything interesting in it?

danimtb: It is always good to have a second example to see if anything changes in respect to the other (although it is not the case)

jgsogo: These are the docs, not testing. I would add all the expected fields (with cardinality if needed) to only one example. We already have ~400 pages in docs, we don't need more 😆

jgsogo: These are the docs, not testing. I would add all the expected fields (with cardinality if needed) to only one example. We already have ~400 pages in docs, we don't need more

PR #963, opened 2018-12-04
 Explantion about 'outdated from recipe' in FAQ

### LINKED ISSUE #478: What does "outdated from recipe" mean, exactly?
If you install a package, and the the recipe is updated (either because it changed on remote, or because it was overwritten manually), the package is marked as "outdated from recipe". But I see only one place in the docs, where this condition is explained: http://docs.conan.io/en/latest/reference/commands/misc/remove.html

And it does not say much. It does not explain how this condition occurs. Or how it is treated in other commands, e.g. what happen if I conan install a package, while it is already in the cache, but with outdated recipe? Could you please clarify this in the docs?

danimtb:
closes #478

### DIFF
--- a/faq/using.rst
+++ b/faq/using.rst
@@ -102,7 +102,7 @@ information is displayed in yellow with:
     # open file.html
     # It will show outdated binaries in yellow.
 
- This information is important to know if the packages are up to date with the recipe or even if the packages are still "accessible" from the
+This information is important to know if the packages are up to date with the recipe or even if the packages are still "accessible" from the
 recipe. That means: if the recipe has completely removed an option (it could be a setting or a requirement) but there are old packages
- that were generated previously with that option, those packages will be impossible to install as their package ID are calculated from the recipe
-file (and that option does not exist anymore).
+that were generated previously with that option, those packages will be impossible to install as their package ID are calculated from the
+recipe file (and that option does not exist anymore).


faq/using.rst
```diff
@@ -82,3 +82,27 @@ configurations).
...
+Basically outdated packages appear when you modify a recipe and export and/or upload it, without re-building binary packages with it. This
+information is displayed in yellow with:
+
+.. code-block:: bash
+
+    $ conan search Pkg/0.1@user/channel --table=file.html
+    # open file.html
+    # It will show outdated binaries in yellow.
+
+This information is important to know if the packages are up to date with the recipe or even if the package are still "accessible" from the
```
jgsogo:
```suggestion
 This information is important to know if the packages are up to date with the recipe or even if the packages are still "accessible" from the
```

faq/using.rst
```diff
@@ -82,3 +82,27 @@ configurations).
...
+recipe. That means: if the recipe has completely removed an option (it could be a setting or a requirement) but there are old packages
+that were generated previously with that option, that package will be impossible to install as the package ID is calculated from the recipe
```
jgsogo:
```suggestion
 that were generated previously with that option, those packages will be impossible to install as their package ID are calculated from the recipe
```

PR #964, opened 2018-12-05
Update changelog: Add include_prerelease and loose option

climblinne:
Add `include_prerelease` and `loose` option to version range expression


danimtb:
Hi @climblinne,

Sorry we didn't notice that the entry was not in the changelog. We have recently introduced an automatic tool to generate it and seems it skipped the description in your PR 😥 

Thanks a lot for your contribution with this new feature and for making us aware.

PR #966, opened 2018-12-06
Update attributes.rst

failshell:
Fixing typo


memsharded:
Thanks for fixing this typo!
This kind of fixes, that could be directly published, can go directly to "master" branch. This was done to develop, it will be published with next release, which will be soon.

PR #969, opened 2018-12-11
deprecating old python reuse

memsharded:
At least comment it in the docs.

Is it something that we want to deprecate? I think so.
Maybe even remove in Conan 2.0 the automatic injection of PYTHONPATH from the dependencies, for all methods?


lasote:
Yes, we can deprecate in favor of an experimental feature because it could change but we know it won't be removed.

PR #1008, opened 2019-01-10
 Added experimental warning in SCM attribute

### LINKED ISSUE #1006: Mark ConanFile scm attribute as experimental
https://docs.conan.io/en/latest/reference/conanfile/attributes.html#scm

danimtb:
closes #1006

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -1026,8 +1026,8 @@ to :ref:`repackage artifacts <repackage>`
 
 .. _scm_attribute:
 
-scm [EXPERIMENTAL]
-------------------
+scm
+---
 
 .. warning::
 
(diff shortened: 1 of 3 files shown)


reference/conanfile/attributes.rst
```diff
@@ -1026,8 +1026,12 @@ to :ref:`repackage artifacts <repackage>`
 
 .. _scm_attribute:
 
-scm
----
+scm [EXPERIMENTAL]
```
memsharded: Do not put EXPERIMENTAL in titles, because later the URLs use it, and then, when it is removed, it will break URLs

danimtb: You are right. I was not aware of that and unfortunately we have experimental in a bunch of titles right now

PR #1026, opened 2019-01-17
CMake default generator

uilianries:
Conan PR: https://github.com/conan-io/conan/pull/4281
Related issue: https://github.com/conan-io/conan/issues/4265


reference/generators/cmake.rst
```diff
@@ -188,3 +188,9 @@ These targets are:
   targets from its ``requires``)
 - Each ``CONAN_LIB`` depends on the direct public dependencies ``CONAN_PKG`` targets of its container package. This guarantees
   correct link order.
+
+Default Targets generated by CMake
+----------------------------------
+
+When a compiler or its version is not detected, the CMake helper generates a default target based on the platform operating system.
```
lasote: When a compiler or its version is not detected, the CMake helper uses a default generator based on the platform operating system.

reference/generators/cmake.rst
```diff
@@ -188,3 +188,9 @@ These targets are:
   targets from its ``requires``)
 - Each ``CONAN_LIB`` depends on the direct public dependencies ``CONAN_PKG`` targets of its container package. This guarantees
   correct link order.
+
+Default Targets generated by CMake
```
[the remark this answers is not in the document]
uilianries: Understood, target could sounds like "target arch".

reference/generators/cmake.rst
uilianries: done!

danimtb:
This fix is actually causing me some trouble now.

Imagine I have a header only package (no settings) that uses ``cmake.install()`` and the helper returns "MinGW Makefiles" as the generator. I fails as I don't have mingw installed and the old behavior was the following: If there is no generator indicated in the cmake configure step, the CMake looks for the available compilers in your machine.

This changed previous behavior for this case and probable we should revert and print a helpful error to "fix" conan-io/conan#4265

PR #1050, opened 2019-01-29
Documentation for JSON output of 'conan info' command

jgsogo:
Documents https://github.com/conan-io/conan/pull/4359

Also removes some `[EXPERIMENTAL]` tags from titles and links.


reference/commands/output/search.rst
```diff
@@ -2,8 +2,8 @@
 .. _search_json:
 
 
-Search output [EXPERIMENTAL]
-----------------------------
+Search output
+-------------
```
danimtb: add under here the warning of experimental please

reference/commands/output/info.rst
```diff
@@ -0,0 +1,118 @@
...
+                "LibE/0.1@lasote/stable"
+            ]
+        },
+        { "...": "..."}
+    ]
+
+
+.. note::
+
+    As this is a marked as *experimental*, some fields may be removed or added.
```
danimtb: Redundant to the top warning message?

jgsogo: We already have the same in the `install` one. I think it is important to say that not only the JSON argument but the fields may be modified (tests are not looking at all the arguments).

danimtb: fine!

danimtb:
Merging this myself!

PR #1064, opened 2019-02-05
Include note about package before to install

uilianries:
Hi!

Someone on Stack Overflow got in trouble when trying execute how-to-launch-conan-install-from-cmake example. The problem is that `Hello/0.1@memsharded/testing` is not installed, and even his remote is not listed there.

Related question: https://stackoverflow.com/questions/53428536

### DIFF
--- a/howtos/cmake_launch.rst
+++ b/howtos/cmake_launch.rst
@@ -45,3 +45,12 @@ If you want to use targets, you could do:
 
     add_executable(main main.cpp)
     target_link_libraries(main CONAN_PKG::Hello)
+
+.. note::
+
+    The package ``Hello/0.1@memsharded/testing`` is only available through `memsharded's repository`_,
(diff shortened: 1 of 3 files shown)


uilianries:
The CI build failed because some links from Microsoft and ARM were removed or redirected. I just updated those links.

danimtb:
Instead of adding the note, could we use a requirement that is available in Conan Center ?

danimtb:
I am going to cherry-pick your commit of the fixed links to it can be merge ASAP. Thanks

uilianries:
> Instead of adding the note, could we use a requirement that is available in Conan Center ?

Sounds easier than executing extra steps to add a new remote :sweat_smile: 

I gonna change it. Thanks!

uilianries:
I've replaced by Catch/2.6.0. Since it's header-only no one will be able to complain about "Hey, that example is not building on my machine!"

danimtb:
Cool!

PR #1065, opened 2019-02-06
Rephrase docs about `apple_dot_clean` tool

### LINKED ISSUE #1063: Confusing explanation of tools.apple_dot_clean()
I don't understand this sentence and I think the explanation should be improved.

> This tool will remove only files matching the pattern if there is file without the prefix too.

https://docs.conan.io/en/latest/reference/tools.html#tools-apple-dot-clean

Consider also including an example.

jgsogo:
closes #1063 

I could add a link, but I don't want users following a link outside our docs... I think it is so specific that if you are reading the docs of this function you already know what you are looking for.

### DIFF
--- a/reference/tools.rst
+++ b/reference/tools.rst
@@ -1531,9 +1531,11 @@ tools.apple_dot_clean()
 
     def apple_dot_clean(folder)
 
-Remove recursively all ``._`` files inside ``folder`` containing metadata information for Apple
-operating systems, these files can appear when unzipping a file that has been created in Macos.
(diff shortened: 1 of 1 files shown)


jgsogo:
_Will remove only the files matching [`._` files] pattern if there is a file without the prefix too_: it will remove `._file.txt` only if `file.txt` exists.

danimtb:
Ok, now I got it. Could you please specify that in the explanation? I know people looking for something like this would know, but I would like to keep it understandable to everyone

danimtb:
I merged it to develop branch 😞

PR #1070, opened 2019-02-11
Adapt docs to new preferred behaviour in 'conan get' command

jgsogo:
Docs for https://github.com/conan-io/conan/pull/4494

### DIFF
--- a/reference/commands/consumer/get.rst
+++ b/reference/commands/consumer/get.rst
@@ -22,7 +22,7 @@ Gets a file or list a directory of a given reference or package.
     optional arguments:
       -h, --help            show this help message and exit
       -p PACKAGE, --package PACKAGE
-                            Package ID (deprecated, use full reference)
+                            Package ID [DEPRECATED: use full reference instead]
       -r REMOTE, --remote REMOTE
                             Get from this specific remote
       -raw, --raw           Do not decorate the text
--- a/reference/commands/consumer/get.rst
+++ b/reference/commands/consumer/get.rst
@@ -6,7 +6,7 @@ conan get
 
 .. code-block:: bash
 
-    $ conan get [-h] [-p PACKAGE] [-r REMOTE] [-raw] reference [path]
+    $ conan get [-h] [-r REMOTE] [-raw] reference [path]
 
 Gets a file or list a directory of a given reference or package.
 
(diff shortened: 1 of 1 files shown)


reference/commands/consumer/get.rst
```diff
@@ -58,7 +56,7 @@ Gets a file or list a directory of a given reference or package.
 
   .. code-block:: bash
 
-      $ conan get zlib/1.2.11@conan/stable -p 09512ff863f37e98ed748eadd9c6df3e4ea424a8
+      $ conan get zlib/1.2.11@conan/stable:09512ff863f37e98ed748eadd9c6df3e4ea424a8
```
uilianries: This is more readable :smile_cat: 

But, as `-p` is suppressed now, but still supported, so why not add a note about that? If I read this documentation, I'll understand it like a breaking change, where `-p` is no longer supported.

danimtb: It has been removed from the help command but the parameter is still usable. Maybe we should leave a deprecation notice with the example below. WDYT?

uilianries: A deprecation notice sounds good!

uilianries: a deprecation notice sounds good!

jgsogo:
This is the output now:

```
⇒  conan get --help
usage: conan get [-h] [-p PACKAGE] [-r REMOTE] [-raw] reference [path]

Gets a file or list a directory of a given reference or package.

positional arguments:
  reference             package recipe reference
  path                  Path to the file or directory. If not specified will
                        get the conanfile if only a reference is specified and
                        a conaninfo.txt file contents if the package is also
                        specified

optional arguments:
  -h, --help            show this help message and exit
  -p PACKAGE, --package PACKAGE
                        Package ID (deprecated, use full reference)
  -r REMOTE, --remote REMOTE
                        Get from this specific remote
  -raw, --raw           Do not decorate the text
```

Shall I use a `[DEPRECATED]` string, @danimtb?

danimtb:
I really like ``[ ]`` for "tagging" things as deprecated, obsolete, experimental and stuff like that. But not mandatory though...

jgsogo:
I like it too, but if I use the _label_ `[DEPRECATED]`, where do I put the reason or alternative? Something like `[DEPRECATED: use full reference instead]`?

danimtb:
yeah, it could be fine like that!

PR #1072, opened 2019-02-12
conan-io/conan#4265 Revert default cmake gen on Windows

uilianries:
Related PR: https://github.com/conan-io/conan/pull/4509

Hi!

This PR reverts the documentation about default CMake generator on Windows.


danimtb:
🤔 I think this whole section shouldn't be here. This should be part of the CMake build helper, not the generator

uilianries:
I think you are right, since the generator section is more about, how to generate conanbuildindfo.cmake, instead of the generator generated by cmake

uilianries:
@danimtb done!

PR #1079, opened 2019-02-18
Update Travis CI integration

danimtb:


integrations/travisci.rst
```diff
@@ -1,90 +1,83 @@
...
+    install:
+      - "[ $CXX = g++ ] && export CXX=g++-6 || true"
+    # Install conan
+      - pip install conan
+    # Automatic detection of your arch, compiler, etc.
+      - conan user
+    script:
+    # Download dependencies and build project
+      - conan install .
+      - make
```
jgsogo: Add a comment here, or maybe the more canonical way: `cmake . -G "Unix makefiles" && make`
And then a comment (or a command) related to _"run your test suite"_

danimtb: you are assuming the usage of CMake in this case and I would like to not focus the attention on that. I can change it and add some comments

jgsogo: yes, I know it, but I prefer a fully working example. It is not necessary to talk about CMake, but if this file is valid for a project using CMake (copy-paste-works) it would be better.

danimtb: Done

integrations/travisci.rst
```diff
@@ -43,7 +43,11 @@ To enable **Travis CI** support, you need to create a *.travis.yml* file and pas
     script:
     # Download dependencies and build project
       - conan install .
-      - make
+    # Call your build system
+      - cmake . -G "Unix makefiles" && make
+      - cmake --build .
```
jgsogo: it is `make` in the line above, and `cmake --build .` here... remove one of them

integrations/travisci.rst
```diff
@@ -1,90 +1,87 @@
...
+    # Install conan
+      - pip install conan
+    # Automatic detection of your arch, compiler, etc.
+      - conan user
+    script:
+    # Download dependencies and build project
+      - conan install .
+    # Call your build system
+      - cmake . -G "Unix makefiles" && make
+      - cmake --build .
```
jgsogo: it is `make` in the line above, and `cmake --build .` here... remove one of them

PR #1081, opened 2019-02-18
Replace /conanbuildinfo*.cmake by /conanbuildinfo*.cmake in all examples

### LINKED ISSUE #1080: CMake examples contain bad example for including conanbuildinfo.cmake
Most (?) CMake examples suggest to include the conanbuildinfo.cmake file as follows:
```
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
```
However, this is incompatible with using conan workspaces. CMAKE_CURRENT_BINARY_DIR should be used instead (as in the conan workspaces example repository).

sigiesec:
Fixes #1080

### DIFF
--- a/creating_packages/getting_started.rst
+++ b/creating_packages/getting_started.rst
@@ -68,7 +68,7 @@ Let's have a look at the root package recipe *conanfile.py*:
             # properly
             tools.replace_in_file("hello/CMakeLists.txt", "PROJECT(MyHello)",
                                   '''PROJECT(MyHello)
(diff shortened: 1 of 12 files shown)


danimtb:
If we consider this change maybe we should also change the CMakeLists.txt template in the ``conan new`` command

danimtb:
Hi @sigiesec,

As we are still discussing in #1080, I am closing this as the "solution" to that issue is not clear. Let's wait until we look into that to make the final PR.

Thanks!

PR #1082, opened 2019-02-19
Fix code snippets

piponazo:
After copying/pasting one of these snippets, I noticed that I was getting the error:

```
ERROR: 
	TypeError: package() takes 0 positional arguments but 1 was given
```


memsharded:
Cool thank you!

This is a fix over the live documentation, so it could have gone to the ``master`` branch, for immediate publication. No issue, the next release will be soon, and it will be published by then. Thanks again for fixing this!

PR #1091, opened 2019-02-26
All references to cpp_info.cppflags are being changed to cxxflags

jgsogo:
Docs for https://github.com/conan-io/conan/pull/4611

As we are deprecating member `cppflags` of `CppInfo` in favor of `cxxflags` all the references to it should be changed.

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -817,7 +817,7 @@ This object should be filled in ``package_info()`` method.
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
 | self.cpp_info.cflags           | Ordered list with pure C flags. Defaulted to ``[]`` (empty)                                             |
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
-| self.cpp_info.cppflags         | Ordered list with C++ flags. Defaulted to ``[]`` (empty)                                                |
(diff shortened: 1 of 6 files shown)


reference/conanfile/attributes.rst
```diff
@@ -817,7 +817,7 @@ This object should be filled in ``package_info()`` method.
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
 | self.cpp_info.cflags           | Ordered list with pure C flags. Defaulted to ``[]`` (empty)                                             |
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
-| self.cpp_info.cppflags         | Ordered list with C++ flags. Defaulted to ``[]`` (empty)                                                |
+| self.cpp_info.cxxflags         | Ordered list with C++ flags. Defaulted to ``[]`` (empty)                                                |
```
danimtb: I wouldn't remove cppflags from here, just document it as deprecated. Rest of the docs are fine without them

jgsogo: Yes, you are right, let the people know about the old name. Thanks!

shelper:
@jgsogo  I am little bit confused by this `cpp_info.cxxflags`, this is for package information, but cxxflags are compiling options, so why we still need it after the package has been packaged? is this the flags will be enforced by the users of the package?

jgsogo:
All the information in the `package_info()` method, like `cpp_info`, is information exposed by the recipe, it is information for the consumers

PR #1097, opened 2019-03-01
Document parallel attribute of CMake build helper

### LINKED ISSUE #1096: CTEST parallel mode enabled if CMake builds in parallel
We are in the process to upgrade from 1.10 to 1.12

I saw from documentation if CMake help builder was constructed (it is by default) using parallel=True then the CTest runs tests in parallel mode as well.

I believe this is an oversight, you should put in CMake.test(...) a new variable that will pass -j to Ctest and not linking the two options build in parallel then tests in parallel.

danimtb:
closes #1096

### DIFF
--- a/reference/build_helpers/cmake.rst
+++ b/reference/build_helpers/cmake.rst
@@ -133,6 +133,23 @@ An example of the value of this property could be:
 
     --config Release
 
+parallel
+++++++++
+
+**Defaulted to**: ``True``
+
+Run CMake process in parallel for compilation, installation and testing. This attribute gets the value for the class constructor and
+appends the ``/MP`` flag in the case of the Visual compiler. However, the parallel executing can be changed for testing for example:
+
+.. code-block:: python
+
+    cmake = CMake(self)
+    cmake.build()  # 'parallel' is enabled by default
+    cmake.parallel = False
+    cmake.test()
+
+In the case of ``cmake.test()`` this flag sets the ``CTEST_PARALLEL_LEVEL`` variable to the according value in :ref:`tools_cpu_count`.
+
 definitions
 +++++++++++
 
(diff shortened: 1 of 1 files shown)


reference/build_helpers/cmake.rst
```diff
@@ -133,6 +133,23 @@ An example of the value of this property could be:
 
     --config Release
 
+parallel
+++++++++
+
+**Defaulted to**: ``True``
+
+Run CMake process in parallel for compilation, installation and testing. This attribute gets the value for the class constructor and
```
uilianries: should we add some comment how [CONAN_CPU_COUNT](https://docs.conan.io/en/latest/reference/env_vars.html#conan-cpu-count) can affect this option?

danimtb: Already referenced in `tools.cpu_count()`

reference/build_helpers/cmake.rst
```diff
@@ -133,6 +133,23 @@ An example of the value of this property could be:
...
+appends the ``/MP`` flag in the case of the Visual compiler. However, the parallel executing can be changed for testing for example:
```
jgsogo: Rephrase this. The flag `parallel` translate into the proper command line argument, for Unix Makefiles it is `-jX` and for MSBuild it is `/m:X`. I think that the `/MP` is a definition you could have in your CMakeLists, but not an argument in the cli.

reference/build_helpers/cmake.rst
```diff
@@ -133,6 +133,23 @@ An example of the value of this property could be:
...
+
+.. code-block:: python
+
+    cmake = CMake(self)
```
jgsogo: I would add the call to `cmake.configure()`, just one line and it turns the snippet into a copy-paste valid one.

reference/build_helpers/cmake.rst
danimtb: I was trying to reference this https://github.com/conan-io/conan/blob/a52dcdc87dd8146b661ed650da616b9fcd2371e3/conans/client/build/cmake_flags.py#L259-L258 but you are right too

PR #1119, opened 2019-03-14
Update binary package reference

maikelvdh:


danimtb:
Hi @maikelvdh and thanks for the fix!

As this is a change affecting the production version of Conan, could you please use master as the base branch for this PR? That way your change will go live inmediately.

Thanks!

danimtb:
Nevermind, adding this to next 1.15 release. Thanks!

PR #1120, opened 2019-03-14
 Refactorize and complete cmake generator section

danimtb:
Before writing the docs for conan-io/conan#4721, I took a look at the cmake generator one and found some missing pieces.

This PRs documents the missing parts and structures and homogenizes the whole section similarly to the build helpers one.

### DIFF
--- a/reference/generators/cmake.rst
+++ b/reference/generators/cmake.rst
@@ -1,190 +1,220 @@
 .. _cmake_generator:
 
-
-`cmake`
-=======
+cmake
+=====
 
 .. container:: out_reference_box
 
     This is the reference page for ``cmake`` generator.
     Go to :ref:`Integrations/CMake<cmake>` if you want to learn how to integrate your project or recipes with CMake.
 
-
-It generates a file named ``conanbuildinfo.cmake`` and declares some variables and methods
+It generates a file named *conanbuildinfo.cmake* and declares some variables and methods.
 
 .. _conanbuildinfocmake_variables:
 
-Variables in conanbuildinfo.cmake
----------------------------------
+Variables in *conanbuildinfo.cmake*
+-----------------------------------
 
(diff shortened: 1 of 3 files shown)


reference/generators/cmake.rst
```diff
@@ -1,190 +1,220 @@
...
+.. code-block:: cmake
+
+    macro(conan_basic_setup)
+        set(options TARGETS NO_OUTPUT_DIRS SKIP_RPATH KEEP_RPATHS SKIP_STD SKIP_FPIC)
+
+Parameters:
+    - ``TARGETS`` (Optional): Setup all the CMake variables by target (only CMake > 3.1.2). Activates the call to the macro
+      ``conan_target_link_libraries()``.
+    - ``NO_OUTPUT_DIRS`` (Optional): Do not adjust the output directories. Deactivates the call to the macro ``conan_output_dirs_setup()``.
+    - ``SKIP_RPATH`` (Optional): **[DEPRECATED]** Use ``KEEP_RPATHS`` instead. Activate ``CMAKE_SKIP_RPATH`` variable in OSX.
```
jgsogo: If it is deprecated, do not document the functionality here

```suggestion
    - ``SKIP_RPATH`` (Optional): **[DEPRECATED]** Use ``KEEP_RPATHS`` instead.
```

danimtb: There were comments regarding deprecation in other issues pointing that we should keep the description. See for example **source_dir** and **build_dir** in here: https://docs.conan.io/en/latest/reference/build_helpers/cmake.html#configure

jgsogo: I find it very confusing: _"skip: use keep instead. Activate SKIP"_, and in the next line: _"KEEP: do not adjust SKIP "_ 💥 . But it's just my preference.

PR #1128, opened 2019-03-20
Document general.error_on_override (CONAN_ERROR_ON_OVERRIDE)

jgsogo:
Docs for https://github.com/conan-io/conan/pull/4771

### DIFF
--- a/using_packages/conanfile_txt.rst
+++ b/using_packages/conanfile_txt.rst
@@ -98,7 +98,7 @@ requirement with one from another user or channel.
     OpenSSL/1.0.2p@conan/stable
     zlib/1.2.11@otheruser/alpha
 
-.. note:: text
+.. note::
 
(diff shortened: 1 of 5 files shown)


using_packages/conanfile_txt.rst
```diff
@@ -98,6 +98,12 @@ requirement with one from another user or channel.
     OpenSSL/1.0.2p@conan/stable
     zlib/1.2.11@otheruser/alpha
 
+.. note::
```
memsharded: I think this note here is a bit confusing, because the user is not aware of the ``override`` keyword, it can only be set in ``conanfile.py``, not ``conanfile.txt``. Maybe this section is too introductory for this concept.

danimtb: It is true that there is not an override keyword but you are talking about overriding a requirement here too, so it might be relevant

PR #1131, opened 2019-03-20
multi-config-install docs

memsharded:
Docs for https://github.com/conan-io/conan/pull/4709

### DIFF
--- a/reference/commands/consumer/config.rst
+++ b/reference/commands/consumer/config.rst
@@ -121,7 +121,7 @@ a Conan client configuration.
 .. note::
     During the installation, Conan skips any file with the name *README.md* or *LICENSE.txt*.
 
-The ``conan config install <item>`` calls are stored in a *config_install.json* file in the conan local cache. That allows to issue a ``conan config install`` command, without arguments, to iterate over the cached configurations, executing them again (updating).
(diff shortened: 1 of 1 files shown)


reference/commands/consumer/config.rst
```diff
@@ -108,7 +132,26 @@ need to specify them.
...
+
+  .. code-block:: bash
+
+      $ conan config install http://url/to/some/config.zip -sf=origin -tf=target
+      $ conan config install http://url/to/some/config.zip -sf=origin2 -tf=target2
+      $ conan config install http://other/url/to/other.zip -sf=hooks -tf=hooks
+      # Later on, execute again the previous configurations cached:
+      $ conan config install
+
+  It's not needed to specify any argument, it will iterate previously stored configurations in *config_install.json*, executing them again.
```
danimtb: Would it be interesting to document the file *config_install.json* in reference/configuration_files so anyone can create one and distribute the "bundle" Conan config?

memsharded: We agreed that for this release, we wanted *config_install.json* to be a cache only, and not be shared for "bundle" conan config. If we want this, we should discuss it, add it in a next release (because support is necessary, as if you get a new *config_install.json* you might need to recurse and call it again).

PR #1134, opened 2019-03-25
Clarify owner user name in the reference

### LINKED ISSUE #1132: [docs] simplify explanation of owner
The documentation explaining the `owner` part of package references was a source of confusion for me learning Conan.

> `pocoproject` is the owner of this package version. It is basically a namespace that allows different users to have their own packages for the same library with the same name, and interchange them. So, for example, you can upload a certain library under your own user name, and later the same packages can be uploaded, without modifications, to another official group or company username.

IMO place a full stop after "allows different users to have their own packages for the same library with the same name" and delete everything after. Especially words like "interchange" and "upload" being used, doesn't really add the the explanation and serves mostly to potentially confuse.

In fact, I opened an ill-fated issue (though I admit this may be a tendency of mine) about it: #4192.

danimtb:
closes #1132 (remember to close the issue when merged)

### DIFF
--- a/using_packages/conanfile_txt.rst
+++ b/using_packages/conanfile_txt.rst
@@ -50,8 +50,7 @@ Where:
     Packages can be overwritten, so it is also OK to have packages like "nightly" or "weekly", that
     are regenerated periodically.
   - ``pocoproject`` is the owner of this package. It is basically a namespace that allows different users to have their own packages for
-    the same library with the same name. For example, you can upload a certain library under your own user/organization name without
-    changing any content of the recipe.
+    the same library with the same name.
   - ``stable`` is the channel. Channels provide another way to have different variants of packages for the same library
     and use them interchangeably. They usually denote the maturity of the package as an arbitrary
     string such as "stable" or "testing", but they can be used for any purpose such as package revisions (e.g., the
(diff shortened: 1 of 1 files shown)


using_packages/conanfile_txt.rst
```diff
@@ -49,11 +49,9 @@ Where:
     string; it does not have to be a number, so, for example, it could indicate if this is a "develop" or "master" version.
     Packages can be overwritten, so it is also OK to have packages like "nightly" or "weekly", that
     are regenerated periodically.
-  - ``pocoproject`` is the owner of this package version. It is basically a namespace that allows
-    different users to have their own packages for the same library with the same name, and
-    interchange them. So, for example, you can upload a certain library under your own user
-    name, and later the same packages can be uploaded, without modifications, to another
-    official group or company username.
+  - ``pocoproject`` is the owner of this package. It is basically a namespace that allows different users to have their own packages for
+    the same library with the same name. For example, you can upload a certain library under your own user/organization name without
```
memsharded: I'd say it is not very clear. The interesting part of using your own username is that it allows to do changes to the packages.

Lawrencemm: I think that the audience for this documentation is likely to understand the significance of namespacing in this context, and we don't even need what comes after the first sentence.

danimtb: Removed last sentence then

PR #1144, opened 2019-03-28
Update remotes.rst

dslotter:
Adding file path for Windows.

### DIFF
--- a/uploading_packages/remotes.rst
+++ b/uploading_packages/remotes.rst
@@ -4,7 +4,7 @@ Remotes
 =======
 
 In the previous sections, we built several packages on our computer that were stored
-in the local cache, typically under *~/.conan/data*. Now, you might want to upload them to a Conan server
+in the local cache, typically under *~/.conan/data* or C:\conan_data. Now, you might want to upload them to a Conan server
 for later use on another machine, project, or for sharing purposes.
 
 Conan packages can be uploaded to different remotes previously configured with a name and a URL.


uploading_packages/remotes.rst
```diff
@@ -4,7 +4,7 @@ Remotes
 =======
 
 In the previous sections, we built several packages on our computer that were stored
-in the local cache, typically under *~/.conan/data*. Now, you might want to upload them to a Conan server
+in the local cache, typically under *~/.conan/data* or C:\conan_data. Now, you might want to upload them to a Conan server
```
memsharded: The C:\conan_data path is not typical at all. I'd say that <userhome>/.conan/data will be 99% of the cases good for the users reading the docs, why do you think it is necessary to add this?

dslotter: It's what my Windows system is using. I had to go looking for it. I installed and ran conan-win-64_1_4_5.exe

dslotter: I'm aware this is an older version of Conan -- Has this location changed in newer versions?

danimtb: I am not aware of this change but the folder is the one stated in the docs. I have Conan install via pip though, so I will have to check the installer. Maybe you have changed the conan storage path https://docs.conan.io/en/latest/reference/config_files/conan.conf.html#storage or the CONAN_USER_HOME

memsharded: Yes, ``conan_data`` does not appear in the codebase or in the docs, so the only way is that you defined the env-var at some point. Could you please check again?

dslotter: I did check again... I found this was configured independently. We can close this thread. Thanks.

danimtb:
Thanks for the update @dslotter. Please close this PR if it is not needed. Thanks! 😄

PR #1148, opened 2019-03-28
Update artifactory_ce.rst

dslotter:
Fix indenting in Python script. (Otherwise loop won't execute as intended.)

### DIFF
--- a/uploading_packages/artifactory_ce.rst
+++ b/uploading_packages/artifactory_ce.rst
@@ -75,4 +75,4 @@ This Python script might be helpful, given that it already defines the respectiv
         print("Downloading %s" % package)
         run("conan download %s -r=local" % package)
 
-        run("conan upload * --all --confirm -r=artifactory")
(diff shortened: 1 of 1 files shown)


uploading_packages/artifactory_ce.rst
```diff
@@ -72,7 +72,7 @@ This Python script might be helpful, given that it already defines the respectiv
...
 
     for package in packages:
-    print("Downloading %s" % package)
-    run("conan download %s -r=local" % package)
+        print("Downloading %s" % package)
+        run("conan download %s -r=local" % package)
 
-    run("conan upload * --all --confirm -r=artifactory")
\ No newline at end of file
+        run("conan upload * --all --confirm -r=artifactory")
```
memsharded: I think this for-loop was intended for the first 2 lines, bu the upload applies to *=all packages, so you don't need to iterate it.

danimtb:
Thanks for the fix!

dslotter:
You're welcome!

PR #1152, opened 2019-03-29
support [None, ANY]

thorsten-klein:
There can be used [None, ANY] within settings.yml file to point out that a setting can be set to any value or that it can be unset.


danimtb:
I think this is de intended behavior but will need the review of any of my colleagues

jgsogo:
These are the docs for https://github.com/conan-io/conan/pull/4765

PR #1166, opened 2019-03-29
Update revisions.rst

dslotter:
Fix capitalization. Fix tense of "requests". Fix awkward sentence. Modify phrasing. Add missing comma. Fix tense of "belong". Fix run-on sentence.


dslotter:
NOTE: The following sentence (at the end of the document) is non-sensical and needs to be totally re-written:

--

If you generate and upload N binary packages for a recipe revision, if you upload the revision you need to generate and upload again the N binaries if you want them to be used with the new recipe.

--

I did not change that sentence because I could not figure out what was originally meant. Please review and re-write that sentence in another revision to the documentation file.

memsharded:
Let me try, it would be something like:

If you generate and upload N binary packages for a recipe with a given revision, if you modify the recipe, and thus the recipe revision, you need to build and upload N new binaries matching that new recipe revision.

Does this make sense?

dslotter:
How about adding the word "then" to improve readability?

If you generate and upload N binary packages for a recipe with a given revision, then if you modify the recipe, and thus the recipe revision, you need to build and upload N new binaries matching that new recipe revision.

memsharded:
Yes, with "then" sounds better, thanks.

PR #1178, opened 2019-04-01
Review revisions section and set artifactory version support

danimtb:
Set Artifactory version to 6.9 supporting revisions


danimtb:
Merging with verbal approval 🤓

PR #1232, opened 2019-04-03
Purge 'export-pkg' reference docs

jgsogo:
I think that there is no need for so much explanation and examples in the `export-pkg` reference docs... I prefer to link to the _editables_ and _package existing binaries_ sections from here as I think those sections add more value to the user. 

Related to https://github.com/conan-io/conan/issues/4829 (will close it).


reference/commands/creator/export-pkg.rst
```diff
@@ -60,127 +62,37 @@ and or '--profile'.
...
+The :command:`export-pkg` command let you create a package from already existing files
+in your workspace, it can be useful if you are using a build process external to Conan
+and do not want to provide it with the recipe. Nevertheless, you should take into
+account that it will generate a package and Conan won't be able to guarantee its
+reproducibility or regenerate it again. This is **not** the normal or recommended flow
+for creating Conan packages.
 
-          def source(self):
-              self.run("git clone https://github.com/memsharded/hello.git")
+Execution of this command will result in several files copied to the ``package_id``
```
danimtb: It is actually the package folder... maybe just "the package folder in the cache"?

reference/commands/creator/export-pkg.rst
```diff
@@ -60,127 +62,37 @@ and or '--profile'.
...
-      $ conan install . --install-folder build_x86 -s arch=x86
-      $ conan build . --build-folder build_x86 --source-folder src
-      $ conan export-pkg . Hello/0.1@user/stable --build-folder build_x86 -s arch=x86
+ - You only have precompiled binaries available, then you can use the :command:`export-pkg`
+   to create the Conan package, or you can build a working recipe to donwload and
+   package them. These scenarios are described in the documentation section
+   :ref:`How to package existing binaries <existing_binaries>`.
 
 
 .. note::
```
danimtb: I am not sure if this note still applies as we have the *graph_info.json* now cc/ @memsharded

reference/commands/creator/export-pkg.rst
```diff
@@ -60,42 +62,57 @@ and or '--profile'.
                             be written
 
 
-:command:`conan export-pkg` executes the following methods of a *conanfile.py* whenever ``--package-folder`` is used:
+The :command:`export-pkg` command let you create a package from already existing files
+in your workspace, it can be useful if you are using a build process external to Conan
```
memsharded: Don't use "workspace" it can be confused with conan workspaces

reference/commands/creator/export-pkg.rst
```diff
@@ -123,69 +146,18 @@ with the ``export-pkg`` command. Check this :ref:`How to package existing binari
...
-      $ conan export-pkg . Hello/0.1@user/stable -pr=myprofile --source-folder=sources --build-folder=build
-
-  And such package will contain just the files:
-
-  .. code-block:: text
-
-      include/mylib.h
-      lib/mylib.lib
-
-- Building a conan package (for architecture x86) in a local directory and then send it to the local cache:
```
memsharded: I am ok with removing this part from this section, but I would like to see it as a more thorough summary in the https://docs.conan.io/en/latest/developing_packages/package_dev_flow.html section, in which the whole step-by-step commands of local development are shown, all together and using the working recipe from the repo (should be moved to examples), starting with a ``git clone examples``

reference/commands/creator/export-pkg.rst
memsharded: Yes, graph_info.json captures the profile already, so the command should be fine with it.

jgsogo: So... we leave the note as it is now, don't we?

reference/commands/creator/export-pkg.rst
jgsogo: Yes, I agree we need to revisit that section now that @danimtb has merged the local workflow example into the `examples` repo https://github.com/conan-io/examples/pull/12, but I would do it in a different issue.

reference/commands/creator/export-pkg.rst
danimtb: @memsharded it is not clear to me: Is the ``--profile`` argument still needed to cover some corner cases of the export-pkg flow?

jgsogo: IMHO, recommending the usage of `--profile` (or settings, options,...) is dangerous as you are going to package/copy existing files which have been created with the `settings/options` that should be in the `graph_info.json`, so if you pass other profile/settings/options you can package files that do not correspond to the profile used to generate them... (I think we are not doing any check here).

memsharded:
Please take into account the feedback in https://github.com/conan-io/conan/issues/3367#issuecomment-485038886 if possible.

reference/commands/creator/export-pkg.rst
```diff
@@ -60,42 +62,57 @@ and or '--profile'.
...
+different sources for the files:
 
-1. ``config_options()``
-2. ``configure()``
-3. ``requirements()``
-4. ``package_id()``
+ * If the argument ``--package-folder`` is provided, Conan will just copy all the
+   contents of that folder to the package one in the cache.
+ * If no ``--package-folder`` is given, Conan will execute the method ``package()``
+   of the recipe twice: first using the ``source_folder`` and then ``build_folder``
```
memsharded: This is no longer true, only 1 call to ``package()`` is done, and ``FileCopier`` is able to internally iterate the 2 folders (source and build)

jgsogo: I'll change it in another PR right now

jgsogo: Here: https://github.com/conan-io/docs/pull/1304

PR #1236, opened 2019-04-05
Added prerelease version matching comment

### LINKED ISSUE #1177: Document better include_prerelease
https://github.com/conan-io/conan/issues/4882

Put an explicit example and clarify that, first the condition (>=1.0) has to be satisfied, then, that using include_prerelease it will also take into account the prereleases.

monsdar:
Fixes #1177

### DIFF
--- a/mastering/version_ranges.rst
+++ b/mastering/version_ranges.rst
@@ -43,6 +43,8 @@ There are two options for the version range:
                                                    # E.g. version "1.2.3.4" would not be accepted. 
    [~1.2.3, loose=False, include_prerelease=True]  # Both options can be used for the same version range.
 
+Note that prereleases are handled with lower precedence than their release counterpart. The expression ``[>=1.0.0, include_prerelease=True]`` will not match the package reference ``Pkg/1.0.0@user/stable``. 1.0.0 is a full release, which is always considered "newer" than any prerelease of 1.0.0. To match release version 1.0.0 the following expression would work: ``[>0.9.9, include_prerelease=True]``.
+
 Version range expressions are evaluated at the time of building the dependency graph, from
 downstream to upstream dependencies. No joint-compatibility of the full graph is computed. Instead,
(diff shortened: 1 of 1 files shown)


mastering/version_ranges.rst
```diff
@@ -43,6 +43,8 @@ There are two options for the version range:
                                                    # E.g. version "1.2.3.4" would not be accepted. 
    [~1.2.3, loose=False, include_prerelease=True]  # Both options can be used for the same version range.
 
+Note that prereleases are handled with lower precedence than their release counterpart. The expression ``[>=1.0.0, include_prerelease=True]`` will not match the package reference ``Pkg/1.0.0@user/stable``. 1.0.0 is a full release, which is always considered "newer" than any prerelease of 1.0.0. To match release version 1.0.0 the following expression would work: ``[>0.9.9, include_prerelease=True]``.
+
```
jgsogo: This is not accurate, ``[>=1.0.0, include_prerelease=True]`` should match ``"1.0.0"``:

```
>>> from semver import SemVer, Range, max_satisfying
>>> v = SemVer("1.0.0", loose=True)
>>> r = Range(">=1.0.0", loose=True)
>>> max_satisfying([v], r, loose=True, include_prerelease=True)
<SemVer 1.0.0 >
```

I think there are a couple of typos in the paragraph:

> Note that prereleases are handled with lower precedence than their release counterpart. The expression ``[>=1.0.0, include_prerelease=True]`` will not match the package reference ``Pkg/1.0.0-23@user/stable``. 1.0.0 is a full release, which is always considered _newer_ than any prerelease of 1.0.0. To match this prerelease ``1.0.0-23`` the following expression would work: ``[>=1.0.0-0, include_prerelease=True]``.

jgsogo:
On https://github.com/conan-io/docs/pull/1411 text was changed and a comment about prerelease version matching was added. 

Anyway, thanks for the PR.

PR #1240, opened 2019-04-09
Update documentation of 'inspect' command with '--raw' attribute

jgsogo:
Docs for https://github.com/conan-io/conan/pull/4903

### DIFF
--- a/reference/commands/misc/inspect.rst
+++ b/reference/commands/misc/inspect.rst
@@ -10,7 +10,7 @@ conan inspect
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


solvingj:
I think the example should be: 
`$ conan inspect zlib/1.2.11@conan/stable -a settings --raw`

jgsogo:
No, `--raw` expects one argument. You can read about the reasoning [here](https://github.com/conan-io/conan/issues/3913#issuecomment-479076242). The problem was that some ``attributes`` expands over more than one line (``options``, ``default_options``) and it could be hard to know which lines correspond to which attributes. So, one attribute at a time is the easiest way to know that all lines correspond to that attribute.

PR #1247, opened 2019-04-16
Add note about editable package pitfall (#4749)

onnodb:
This is a tiny update to the documentation, pointing out the small pitfall with editable packages I encountered in conan-io/conan#4749.

I hope this helps! Please do feel free to reject if you feel this is not a good place and/or wording.

### DIFF
--- a/developing_packages/editable_packages.rst
+++ b/developing_packages/editable_packages.rst
@@ -248,6 +248,12 @@ time, without any Conan command.
 
     When a package is in editable mode, most of the commands will not work. It is not possible to :command:`conan upload`,
     :command:`conan export` or :command:`conan create` when a package is in editable mode.
+    
+.. note::
+
+   CMake-based consumer projects may not pick up on modifications made to header files in the editable package. In this 
+   case, a full rebuild is necessary to pick up on the changes in the package's source files. Using :ref:`workspaces`
+   may provide a better workflow in this specific case.
 
 Revert the editable mode
 ------------------------


developing_packages/editable_packages.rst
```diff
@@ -248,6 +248,12 @@ time, without any Conan command.
 
     When a package is in editable mode, most of the commands will not work. It is not possible to :command:`conan upload`,
     :command:`conan export` or :command:`conan create` when a package is in editable mode.
+    
+.. note::
+
+   CMake-based consumer projects may not pick up on modifications made to header files in the editable package. In this
```
danimtb: I think this behavior is not specific to CMake, but any build system and working with editable packages

danimtb: I mean, this is the normal (maybe not expected at first try) behavior

onnodb: That's entirely fair, it's just that I offered to add a note in the documentation in the linked Conan issue. 

I think it's enough of a potential trip-up that I figured it could save people troubleshooting time if one would point this out up-front. But feel free to close the PR if you don't think that's necessary.

danimtb: I think the note is fine, just make it more generic (it doesn't happen only with header libraries but any other library as well) and is not specific to CMake

czoido:
Hi! We are closing this PR as we are no longer merging changes for the Conan 1.x documentation. Thank you very much for your time and effort!

PR #1249, opened 2019-04-23
add docs for Emscripten

SSE4:
docs for https://github.com/conan-io/conan/pull/4346


jgsogo:
I need more docs here:

 * Maybe we should compose the profile to override only the required fields, would it work something like this?

    ```ini
    include(default)
    [settings]
    os=Emscripten
    arch=asm.js
    compiler=clang
    compiler.version=6.0
    compiler.libcxx=libc++
    [options]
    [build_requires]
    emsdk_installer/1.38.29@bincrafters/stable
    ```

    and then explain why do we need to override those settings

 * Bare minimum example, exporting a function from a library and calling it from the browser. It isn't so straightforward, is it?

SSE4:
okay, adding this

SSE4:
I need to add some new example, as poco-md5 has a dependency on OpenSSL, which doesn't build for emscripten

lasote:
Please @SSE4 we need this for the release.

SSE4:
yeah, I am working on adding an example

SSE4:
now depends on https://github.com/conan-io/examples/pull/13

SSE4:
@jgsogo 
profile now includes `default`
added a section on how to run code in the browser (requires https://github.com/conan-io/examples/pull/13)

integrations/emscripten.rst
```diff
@@ -0,0 +1,150 @@
...
+  arch=wasm
+  compiler=clang
+  compiler.version=6.0
+  compiler.libcxx=libc++
+  [options]
+  [build_requires]
+  emsdk_installer/1.38.29@bincrafters/stable
+  [env]
+
+These profile above are using the `emsdk_installer/1.38.29@bincrafters/stable <https://github.com/bincrafters/conan-emsdk_installer>`_ conan package.
```
jgsogo: Not yet in conan-center, maybe it is worth to remember link how to add the _bincrafters_ remote

SSE4: added (also submitted an inclusion request)

danimtb:
Merging the docs so they can be released. We will have to merge the emscripten example later

PR #1262, opened 2019-04-29
Docs for deploy generator and review of existing info

danimtb:
Docs for PR conan-io/conan#4972
Issue conan-io/conan#4847

- Added section for generator class reference
- Removed duplicated info in how-to about generator packages
- Updated custom integration in the integration section

### DIFF
--- a/howtos/custom_generators.rst
+++ b/howtos/custom_generators.rst
@@ -120,7 +120,7 @@ Now, let's create a folder for the application that will use Premake as build sy
     $ cd ..
     $ mkdir premake-project && cd premake-project
 
-Put the following files inside. Note the ``PremakeGen@0.1@memsharded/testing`` package reference in your *conanfile.txt*.
+Put the following files inside. Note the ``PremakeGen@0.1@myuser/testing`` package reference in your *conanfile.txt*.
 
 .. code-block:: text
    :caption: *conanfile.txt*
@@ -191,7 +191,7 @@ Now everything works, so you might want to share your generator:
 
 .. code-block:: bash
 
-    $ conan upload PremakeGen/0.1@memsharded/testing
+    $ conan upload PremakeGen/0.1@myuser/testing
 
 .. tip::
 
--- a/howtos/custom_generators.rst
+++ b/howtos/custom_generators.rst
@@ -89,7 +89,7 @@ This is a full working example. Note the ``PremakeDeps`` class as a helper. The 
 individual library separately, then also an aggregated information for all dependencies. This ``PremakeDeps`` wraps a single item of such
 information.
 
-Note the **name of the package** will be **PremakeGen/0.1@user/channel** as that is the name given to it, while the generator name is
+Note the **name of the package** will be **PremakeGen/0.1@<user>/<channel>** as that is the name given to it, while the generator name is
 **Premake** (the name of the class that inherits from ``Generator``). You can give the package any name you want, even the same as the
 generator's name if desired.
 
@@ -97,27 +97,37 @@ You ``export`` the package recipe to the local cache, so it can be used by other
 
 .. code-block:: bash
 
-   $ conan export . memsharded/testing
+   $ conan export . myuser/testing
 
 Using the generator
 -------------------
 
-Let's create a test project that uses this generator. We will use a simple "Hello World" program
-and also an existing library conan package, we will use the simple "Hello World" package we already created before:
+Let's create a test project that uses this generator. We will use a simple application that will use a "Hello World" library package as a
+requirement.
+
+First, let's create the "Hello World" library package:
 
 .. code-block:: bash
 
-   $ cd ..
-   $ mkdir premake-project && cd premake-project
+    $ mkdir conan-hello && cd conan-hello
+    $ conan new hello/0.1
+    $ conan create . myuser/testing
 
-Now put the following files inside. Note the ``PremakeGen@0.1@memsharded/testing`` package reference in your *conanfile.txt*.
+Now, let's create a folder for the application that will use Premake as build system:
+
+.. code-block:: bash
+
+    $ cd ..
+    $ mkdir premake-project && cd premake-project
+
+Put the following files inside. Note the ``PremakeGen@0.1@memsharded/testing`` package reference in your *conanfile.txt*.
 
 .. code-block:: text
    :caption: *conanfile.txt*
 
     [requires]
-    Hello/0.1@memsharded/testing
-    PremakeGen@0.1@memsharded/testing
+    hello/0.1@myuser/testing
+    PremakeGen@0.1@myuser/testing
     
     [generators]
     Premake
@@ -176,7 +186,7 @@ Now we are ready to build the project:
     $ make (or mingw32-make if in windows-mingw)
     $ ./MyApplication
     Hello World!
-   
+
 Now everything works, so you might want to share your generator:
 
 .. code-block:: bash
--- a/howtos/custom_generators.rst
+++ b/howtos/custom_generators.rst
@@ -3,97 +3,29 @@
 How to create and share a custom generator with generator packages
 ==================================================================
 
-There are several built-in generators, like ``cmake``, ``visual_studio``, ``xcode``...
-But what if your build system is not included? Or maybe the existing built-in generators
-doesn't satisfy your needs. There are several options:
+There are several built-in generators, like ``cmake``, ``visual_studio``, ``xcode``... But what if your build system is not included or the
+existing built-in ones doesn't satisfy your needs? This **how to** will show you how to create a generator for
+[premake](https://premake.github.io/) build system.
 
-- Use the ``txt`` generator, that generates a plain text file easy to parse, which you might
-  be able to use.
-- Use ``conanfile.py`` data, and for example in the ``build()`` method, access that information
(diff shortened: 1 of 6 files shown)


howtos/custom_generators.rst
```diff
@@ -3,98 +3,30 @@
...
-
-Creating a custom generator
----------------------------
-
-Basically a generator is a class that extends ``Generator`` and implements two properties: ``filename``,
-which will be the name of the file that will be generated, and ``content`` with the contents of
-that file. The **name of the generator** itself will be taken from the class name:
+There are several built-in generators, like ``cmake``, ``visual_studio``, ``xcode``... But what if your build system is not included or the
+existing built-in ones doesn't satisfy your needs? This **how to** will show you how to create a generator for
+[premake](https://premake.github.io/) build system.
```
jgsogo: This is not a valid hyperlink for `rst`.

howtos/custom_generators.rst
```diff
@@ -3,98 +3,30 @@
...
-
-Premake generator example
--------------------------
-
-Create a project (the best is a git repository):
+Create a folder with a new *conanfile.py* with the following contents::
 
 .. code-block:: bash
 
    $ mkdir conan-premake && cd conan-premake
```
jgsogo: This is not rendering ok:

image

howtos/custom_generators.rst
```diff
@@ -176,21 +101,18 @@ You ``export`` the package recipe to the local cache, so it can be used by other
 Using the generator
 -------------------
 
-Let's create a test project that uses this generator, and also an existing library conan package,
-we will use the simple "Hello World" package we already created before:
+Let's create a test project that uses this generator. We will use a simple "Hello World" program
+and also an existing library conan package, we will use the simple "Hello World" package we already created before:
```
jgsogo: Where is this "Hello World" package? If I enter this page of the documentation, I haven't created anything before.

I think that every _How to_ should be self-contained.

howtos/custom_generators.rst
```diff
@@ -151,21 +85,12 @@ Then, write in it the following **conanfile.py**:
...
 
-This is a full working example. Note the ``PremakeDeps`` class as a helper. The generator is
-creating premake information for each individual library separately, then also an aggregated
-information for all dependencies. This ``PremakeDeps`` wraps a single item of such information.
-
-Note the **name of the package** will be **PremakeGen/0.1@user/channel** as that is the name given
-to it, while the generator name is **Premake**. You can give the package any name you want, even
-matching the generator name if desired.
+Note the **name of the package** will be **PremakeGen/0.1@user/channel** as that is the name given to it, while the generator name is
+**Premake**. You can give the package any name you want, even matching the generator name if desired.
```
jgsogo: Is it case sensitive? Can I give the generator a custom name?

howtos/custom_generators.rst
```diff
@@ -199,83 +121,85 @@ reference in conanfile.txt.
...
-   $ conan install . -s compiler=gcc -s compiler.version=4.9 -s compiler.libcxx=libstdc++ --build
-   $ premake4 gmake
-   $ make (or mingw32-make if in windows-mingw)
-   $ ./MyApplication
-   Hello World!
+    $ conan install . -s compiler=gcc -s compiler.version=4.9 -s compiler.libcxx=libstdc++ --build
+    $ premake4 gmake
+    $ make (or mingw32-make if in windows-mingw)
+    $ ./MyApplication
+    Hello World!
```
jgsogo: I need to read here something like  _"after calling `conan install ...` there is anew file, `conanpremake.lua` (where the name comes from) that is created with...."_.

integrations/other.rst
```diff
@@ -7,54 +7,55 @@ Custom integrations
...
-  call your build system with that information, without requiring to generate a file.
-- Write and **create your own generator**. So you can upload it, version and reuse it, as well
-  as share it with your team or community. Check :ref:`generator packages <dyn_generators>` too.
-  
-  
-.. note:: 
-   
-   Need help integrating your build system? Tell us what you need. info@conan.io
+- First, search in Bintray for generator packages. Generators can be created and contributed by users as regular packages, so you can depend
+  on them as a normal requirement, use versioning and evolve faster without depending on the Conan releases. See :ref:`custom_generator`.
```
jgsogo:
```suggestion
  on them as a normal requirement, use versioning and evolve faster without depending on the Conan releases. If you can't find one that already fits your needs, you can write your own (:ref:`custom_generator`).
```

integrations/other.rst
```diff
@@ -7,54 +7,55 @@ Custom integrations
...
 
+- You can use the :ref:`text_generator` or :ref:``json_generator`` generators. They will generate a text file, simple to read that you can
```
jgsogo:
```suggestion
- You can use the :ref:`text_generator` or :ref:`json_generator` generators. They will generate a text file, simple to read that you can
```

integrations/other.rst
```diff
@@ -7,54 +7,55 @@ Custom integrations
...
+  easily parse with your tools to extract the required information.
+
+- Use the **conanfile data model** (ref:`deps_cpp_info_attributes_reference`, :ref:`deps_env_info_attributes_reference`) and access its
```
jgsogo:
```suggestion
- Use the **conanfile data model** (:ref:`deps_cpp_info_attributes_reference`, :ref:`deps_env_info_attributes_reference`) in your recipe to access its
```

integrations/other.rst
danimtb:
```suggestion
- You can use the :ref:`text_generator` or :ref:`json_generator` generators. They will generate a text file, simple to read that you can
```

integrations/other.rst
```diff
@@ -7,54 +7,55 @@ Custom integrations
...
+  properties and values, so you can directly call your build system with that information, without requiring to generate a file.
+
+- Write and **create your own generator**. So you can upload it, version and reuse it, as well as share it with your team or community.
+  Check :ref:`dyn_generators`.
```
jgsogo: If we move this entry list to the second place, we can remove from the first entry the comment about creating a _custom generator_.

reference/generators/custom.rst
```diff
@@ -0,0 +1,91 @@
+.. _custom_generator:
```
jgsogo: I wouldn't add this to the list of _available generators_... I would put a comment below that list ("if none of these generators fits your needs, you can write a custom generator"):

image

linking to the `howtos/custom_generators` or `integrations/other` and merge all this content there, otherwise we will have three sections of the documentation talking about the same: how to write a custom generator, and it is too much from my POV.

reference/generators/custom.rst
```diff
@@ -0,0 +1,91 @@
...
+=========================
+
+Don't see a generator that suit your needs? You can create a custom generator package and use it in your recipes!
+
+Basically a generator is a class that extends ``Generator`` and implements two properties:
+
+- ``filename`` (Required): Should return the name of the file that will be generated
+- ``content`` (Required): Should return the contents of the file with the desired format.
+
+The **name of the generator** itself will be taken from the class name. For example:
```
jgsogo: is the name case sensitive?

integrations/other.rst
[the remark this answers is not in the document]
memsharded: I still see the double quote.

reference/generators/custom.rst
danimtb: I understand what you mean, but we have different possibilities. The section at ``reference/generators/custom.rst`` is intended to be a reference of the generator class and the available attributes and properties. So I don't like the idea of moving it out of the reference but agree that making it part of the list is probably not ideal.

The integration part is a more generic one talking about the json generator, txt one, deps_cpp_info...

An finally the implementation of a generator package is covered in a how-to. All sections are properly cross-referenced and I think it makes

howtos/custom_generators.rst
danimtb: It is a package name, no further restriction applied. And the generator name is explained in the reference. I have added a clarification here too

reference/generators/custom.rst
danimtb: Added clarification

reference/generators.rst
```diff
@@ -43,3 +43,5 @@ Available generators:
    generators/json
    generators/premake
    generators/make
+   generators/deploy
+   generators/custom
```
jgsogo: I still think we should put this out of the list (in a line or in a tip/note):

```suggestion

If none of these generators fits your needs, you can create your own
:ref:`custom generator <custom_generator>`.
```

The only caveat is that we will need the `:orphan:` keyword inside `reference/generators/custom.rst` to avoid the rst error.

danimtb: Still don't like the idea of having the section so hidden, but I have taken your suggestion and we can change it in the future. Thanks for the helpful reviews

memsharded: I would also add an entry, linking to this, in ``Extending Conan``. This is clearly a point of extension and should be summarized there. Can be a short summary: purpose, goal, brief description and short recipe and link to main page

danimtb: I think in that case we should move the how-to to the ``Extending Conan`` section. Otherwise, there would be too many entries for generators with bits of information around the place. I will create an issue apart so this PR can be merged

PR #1266, opened 2019-04-30
Docs for compiler.cppstd subsetting

jgsogo:
Doc for https://github.com/conan-io/conan/pull/4917

### DIFF
--- a/howtos/manage_cpp_standard.rst
+++ b/howtos/manage_cpp_standard.rst
@@ -23,7 +23,8 @@ The consumer can specify it in a :ref:`profile<profiles>` or with the ``-s`` par
 
 
 As it is a subsetting, it can have different values for each compiler (also, take into account
-that depending on the version of the compiler the standard could have only partial support).
+that depending on the version of the compiler the standard could have only partial support
+and may change the ABI).
 
 Valid values for ``compiler=Visual Studio``:
 
--- a/howtos/manage_cpp_standard.rst
+++ b/howtos/manage_cpp_standard.rst
@@ -6,27 +6,24 @@ How to manage C++ standard [EXPERIMENTAL]
 .. warning::
 
     This is an **experimental** feature subject to breaking changes in future releases.
+    Previously, it was implemented as a first level setting ``cppstd``, we encourage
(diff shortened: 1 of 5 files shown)


howtos/manage_cpp_standard.rst
```diff
@@ -6,27 +6,25 @@ How to manage C++ standard [EXPERIMENTAL]
...
+    Previously, it was implemented as a first level setting ``cppstd``, we encourage
+    you to adopt the new subsetting and update your recipes if they were including the
+    deprecated one in its :ref:`settings_property` attribute.
 
-The setting representing the C++ standard is ``cppstd``.
-The detected default profile doesn't set any value for the ``cppstd`` setting.
+
+The setting representing the C++ standard is ``compiler.cppstd``.
+The detected default profile doesn't set any value for the ``compiler.cppstd`` setting, but
+Conan will use the default for each compiler if no value is provided.
```
lasote: Not accurate. Conan won't do anything if no value is specified. But the compiler will use, of course, the default value.

howtos/manage_cpp_standard.rst
```diff
@@ -84,11 +82,17 @@ When the ``cppstd`` setting is declared in the recipe and the consumer specifies
 Package compatibility
 ---------------------
 
-By default Conan will detect the default standard of your compiler to not generate different binary packages.
-For example, you already built some ``gcc > 6.1`` packages, where the default std is ``gnu14``.
-If you introduce the ``cppstd`` setting in your recipes and specify the ``gnu14`` value, Conan won't generate
+By default Conan will use the default standard for your compiler, so it won't generate different
```
lasote: Idem. Not accurate. You can say that when no value is specified, Conan will compute by default the same package_id corresponding to the default std of the compiler, so when you explicit the std value for the default of the compiler, the id will be the same as no specifying it.

howtos/manage_cpp_standard.rst
jgsogo: That's right, and very important!

howtos/manage_cpp_standard.rst
jgsogo: That's right, and very important

PR #1268, opened 2019-04-30
Updated year in license

danimtb:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2016 JFrog Ltd.
+Copyright (c) 2019 JFrog Ltd.
```
uilianries: Should we update to the current year, or a range with initial year, like `2016-2019` ?

danimtb:
I have just followed what we have at conan-io/conan, but that's why I assigned this to @memsharded

PR #1294, opened 2019-05-27
Add an example of removing system requirements via a wildcard

Minimonium:
Docs for: conan-io/conan#5176


danimtb:
Thanks a lot!

PR #1296, opened 2019-05-28
Feature/build mode cascade

lasote:


danimtb:
A usage example would be useful but we can add it later. I will open an issue

PR #1311, opened 2019-05-31
Add know pip installation issue

### LINKED ISSUE #1256: Issue realated to `pip install --user conan`
When you install a Python package into user workspace with `pip install --user conans` all the console scripts are created into `$HOME/.local/bin`. This path should be included into the `$HOME/.profile` file:

```sh
# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

...but it will be added to the `PATH` only if that directory exists. If Conan is the first application being installed this two lines will fail:

```bash
$> pip install --user conan
$> conan
bash: conan: command not found
```

The user needs to open the session again (start a new console) or reload the `.profile` file:

```bash
$> source ~/.profile
```

A note about this could be added to [Known installation issues with pip](https://docs.conan.io/en/latest/installation.html#known-installation-issues-with-pip), or to the _"Please read carefully"_ section.

danimtb:
closes #1256

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -66,9 +66,9 @@ Known installation issues with pip
       $ python3.4 get-pip.py
       $ pip install conan
 
-- When Conan being the first application installer via pip with :command:`pip install --user <username>`, usually the path is not updated in
-  the environment and the :command:`conan` command is not available in the current terminal session. This can be solved restarting the
-  session or running the following command:
+- When Conan is installed with :command:`pip install --user <username>`, usually a new directory is created for it. However, the directory
+  is not appended automatically to the `PATH` and the :command:`conan` commands do not work. This can be solved restarting the session of
+  the terminal or running the following command:
 
   .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


installation.rst
```diff
@@ -60,11 +60,19 @@ Known installation issues with pip
...
 
-    $ curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-    $ python3.4 get-pip.py
-    $ pip install conan
+      $ curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+      $ python3.4 get-pip.py
+      $ pip install conan
+
+- When Conan is installed with :command:`pip install --user <username>`, usually a new directory is created for it. However, the directory
+  is not appended automatically to the `PATH` and the :command:`conan` commands do not work. This can be solved restarting the session of
```
jgsogo: Just in case 😅 

```suggestion
  is not appended automatically to the `PATH` and the :command:`conan` commands do not work. This can usually be solved restarting the session of
```

PR #1322, opened 2019-06-12
Update multiconfig package approach N configs -> 1 package

### LINKED ISSUE #1325: Misleading documentation for packaging N configs -> 1 package (MSVC Alternative 2)
- [ *] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [ *] I've specified the Conan version, operating system version and any tool that can be relevant.
- [ *] I've explained the steps to reproduce the error or the motivation/use case of the question/suggestion.

conan 1.11.2, MSVC 2017
[Alternative #2](https://github.com/memsharded/hello_multi_config/blob/master/conanfile.py#L33) doesn't work because `cmake.command_line` is [read-only](https://docs.conan.io/en/latest/reference/build_helpers/cmake.html#command-line-read-only)

How to make `Alternative 2` approach work? Could you update the documentation and `hello_multi_config` repository?

P.S. I'm trying to package Google Protobuf, `find_package(Protobuf REQUIRED)` is used everywhere and I don't want to change it to `find_package(Protobuf REQUIRED CONFIG)` so I'm trying to pack Debug and Release libraries to a single package.

danimtb:
closes conan-io/conan#5341
closes #1325 

to be merged after https://github.com/conan-io/examples/pull/18

### DIFF
--- a/creating_packages/package_approaches.rst
+++ b/creating_packages/package_approaches.rst
@@ -142,7 +142,7 @@ Such a package can define its information for consumers as:
         self.cpp_info.release.libs = ["mylibrary"]
         self.cpp_info.debug.libs = ["mylibrary_d"]
 
-This will translate to the CMAake variables:
+This will translate to the CMake variables:
 
 .. code-block:: cmake
 
--- a/creating_packages/package_approaches.rst
+++ b/creating_packages/package_approaches.rst
@@ -74,17 +74,15 @@ mistake could be a very risky issue.
 
 Read more about this in :ref:`method_package_info`.
 
-
 .. _packaging_approach_N_1:
 
 N configs -> 1 package
 ----------------------
 
-You may want to package both debug and release artifacts in the same package,
-so it can be consumed from IDEs like Visual Studio. This will change the debug/release configuration from the
-IDE, without having to specify it in the command line. This type of package can contain different
-artifacts for different configurations, and can be used for example to include both the release and debug version of the "hello"
-library, in the same package.
+You may want to package both debug and release artifacts in the same package, so it can be consumed from IDEs like Visual Studio. This will
+change the debug/release configuration from the IDE, without having to specify it in the command line. This type of package can contain
+different artifacts for different configurations and can be used to include both the release and debug version of a library in the same
+package.
 
 .. image:: /images/multi_conf_packages.png
     :height: 300 px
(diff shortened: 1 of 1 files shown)


creating_packages/package_approaches.rst
```diff
@@ -154,7 +142,7 @@ Such a package can define its information for consumers as:
         self.cpp_info.release.libs = ["mylibrary"]
         self.cpp_info.debug.libs = ["mylibrary_d"]
 
-This will translate to the cmake variables:
+This will translate to the CMAake variables:
```
SSE4:
```suggestion
This will translate to the CMake variables:
```

creating_packages/package_approaches.rst
```diff
@@ -164,62 +152,54 @@ This will translate to the cmake variables:
...
-These specific config variables will not be automatically applied, but you can directly use them
-in your consumer CMake build script.
+These specific config variables will not be automatically applied, but you can directly use them in your consumer CMake build script.
 
 .. note::
 
-    The automatic conversion of multi-config variables to generators is currently only implemented
-    in the ``cmake`` and ``txt`` generators. If you want to have support for them in another
-    build system, please open a GitHub issue.
+    The automatic conversion of multi-config variables to generators is currently only implemented in the ``cmake`` and ``txt`` generators.
```
SSE4: I think `visual_studio` & `visual_studio_multi` generators also support it for a while

danimtb: You are right. However, I am not sure if adding the ``*_multi`` generators here would be confusing, as those are used only for consumption. Maybe @memsharded can give us a hint here

danimtb:
```suggestion
    The automatic conversion of multi-config variables to generators is currently only implemented in the ``cmake``, ``visual_studio`` and ``txt`` generators.
```

creating_packages/package_approaches.rst
```diff
@@ -93,54 +91,44 @@ library, in the same package.
...
-syntax:
+            cmake_release = CMake(self, build_type="Debug")
+            cmake_release.configure()
+            cmake_release.build()
+
+            cmake_debug = CMake(self, build_type="Release")
+            cmake_debug.configure()
+            cmake_debug.build()
+
+In this case, we assume that the binaries will be differentiated with a suffix in the CMake syntax:
```
jgsogo: I think it would be more clear if we provide more information:

```suggestion
In this example, the binaries will be differentiated with a suffix in the CMake syntax, so we have to add this information to the data provided to the consumers in the `package_info` function:
```

and in the code-block below add the full function:

```python
def package_info(self):
    self.cpp_info.release.libs = ["mylibrary"]
    self.cpp_info.debug.libs = ["mylibrary_d"]
```

creating_packages/package_approaches.rst
```diff
@@ -164,62 +152,54 @@ This will translate to the cmake variables:
     set(CONAN_LIBS_DEBUG mylibrary_d ${CONAN_LIBS_DEBUG})
     set(CONAN_LIBS_RELEASE mylibrary ${CONAN_LIBS_RELEASE})
 
-And these variables will be correctly applied to each configuration by ``conan_basic_setup()``
-helper.
+And these variables will be correctly applied to each configuration by ``conan_basic_setup()`` helper.
```
jgsogo: What about other generators? Maybe it is useful to mention the `cmake_find_package` at least as it is the one we are moving to. I suppose the target created by Conan in the `FindXXX.cmake` uses the proper library depending on the build_type, right?

creating_packages/package_approaches.rst
```diff
@@ -164,62 +152,54 @@ This will translate to the cmake variables:
...
 
-    All these steps are already coded in the repo https://github.com/memsharded/hello_multi_config and commented
-    out as **"Alternative 2"**
+    All these steps are already coded in the repo https://github.com/conan-io/examples/tree/master/features/multi_config and commented
+    out as **"Alternative 2"**.
 
-Also, you can use any custom configuration as they are not restricted. For example, if
-your package is a multi-library package, you could try doing something like:
+Also, you can use any custom configuration as they are not restricted. For example, if your package is a multi-library package, you could
+try doing something like:
```
jgsogo:
```suggestion
try to do something like:
```

creating_packages/package_approaches.rst
```diff
@@ -164,62 +152,54 @@ This will translate to the cmake variables:
...
 N configs (1 build) -> N packages
 ---------------------------------
 
 It’s possible that an existing build script is simultaneously building binaries for different
 configurations, like debug/release, or different architectures (32/64bits), or library types
-(shared/static). If such a build script is used in the previous “Single configuration packages”
-approach, it will definitely work without problems. However, we’ll be wasting precious build time, as
+(shared/static). If such a build script is used in the previous "Single configuration packages"
+approach, it will definitely work without problems. However, we'll be wasting precious build time, as
 we’ll be re-building the rebuilding project for each package, then extracting the relevant artifacts for
```
jgsogo:
```suggestion
we’ll be re-building the project for each package, then extracting the relevant artifacts for
```

creating_packages/package_approaches.rst
```diff
@@ -93,54 +91,44 @@ library, in the same package.
...
-Creating a multi-configuration Debug/Release package is simple, see the following example
-using CMake.
+Creating a multi-configuration debug/release package is simple
 
-The first step will be to remove ``build_type`` from the settings. It will not be an input setting, the
-generated package will always be the same, containing both Debug and Release artifacts.
-The Visual Studio runtime is different for debug and release (``MDd`` or ``MD``) and is set using the default runtime (MD/MDd). If this meets your needs, we recommend removing the ``runtime`` subsetting in the
-``configure()`` method:
+The first step will be to remove ``build_type`` from the settings. It will not be an input setting and the generated package will always
+contain both debug and release artifacts.
```
SSE4: may we add some clarification, what happens to other configurations besides Release/Debug (e.g. RelWithDebInfo), is it unsupported case or it will just work out of the box?

creating_packages/package_approaches.rst
```diff
@@ -93,54 +91,44 @@ library, in the same package.
...
-                    self.output.info("Building %s" % config)
-                    self.run('cmake "%s" %s -DCMAKE_BUILD_TYPE=%s'
-                            % (self.source_folder, cmake.command_line, config))
-                    self.run("cmake --build .")
-                    shutil.rmtree("CMakeFiles")
-                    os.remove("CMakeCache.txt")
-
-In this case, we assume that the binaries will be differentiated with a suffix in the Cmake
-syntax:
+            cmake_release = CMake(self, build_type="Debug")
```
SSE4: doesn't it require different build folders to be used? otherwise, I guess Release will just override Debug (or fail?)

creating_packages/package_approaches.rst
danimtb: Here I just fixed the outdated information to adequate it to the example, I didn't want to introduce in this PR new information.

I am sure the whole docs would benefit from a review regarding the cmake generators, but we still have to clear the path towards the full usability of the find_package generator.

I will try what you suggest and open a new issue accordingly.

SSE4: I think the only generators currently supported are `cmake` and `visual_studio` (plus their corresponding `multi` variants). we should mention that. 
besides that, what happens to `conan_basic_setup(TARGETS)`, is it supported as well?

creating_packages/package_approaches.rst
danimtb: name of the generated artifacts are different, check the CMakeLists of the example

creating_packages/package_approaches.rst
danimtb: probably it will just work, but I don't want to make the example more complex or make the explanation too long for the user

creating_packages/package_approaches.rst
jgsogo: I agree it deserves an investigation to know if it works or not, but it is not the aim of this PR. @danimtb , if you open the new issue related to this example, please add the concerns by @SSE4 so we have a look into it too.

creating_packages/package_approaches.rst
jgsogo: I think it is ok for this PR to keep the Debug/Release as it was before.

PR #1324, opened 2019-06-12
Update imports copy

uilianries:
I've omitted here, actually at line 139 could have `link_dll` listed:

```
package()
---------

The actual creation of the package, once that it is built, is done in the ``package()`` method. Using the ``self.copy()`` method, artifacts
are copied from the build folder to the package folder.

The syntax of ``self.copy`` inside ``package()`` is as follows:

.. code-block:: python

    self.copy(pattern, dst="", src="", keep_path=True, symlinks=None, excludes=None, ignore_case=False)
```

Why didn't I list it there? `link_dll` is a feature to be used with `imports`, but  both `package` and `imports` share the same code.

### DIFF
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -623,7 +623,7 @@ The ``self.copy()`` method inside ``imports()`` supports the following arguments
 
 .. code-block:: python
 
-    def copy(pattern, dst="", src="", root_package=None, folder=False, ignore_case=False, excludes=None, keep_path=True)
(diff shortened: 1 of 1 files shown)


danimtb:
Should we create a different copy interface to differentiate the import method from the package? I think this parameter makes no sense for package()

uilianries:
They have different interfaces, but both are using FileCopier.

memsharded:
Closing (check https://github.com/conan-io/conan/pull/5352#issuecomment-1396078635)

PR #1346, opened 2019-07-03
discourage private requirements

memsharded:
Biten again by some corner and weird case of private requirements abuse. I think at least we should discourage their use in the docs.


danimtb:
I understand the intention but still think this kind of requirements have to be supported. Please fix the indentation to make green the CI

danimtb:
Also, I think those corner cases of the general purpose and current status of private requirements should be documented as well so users understand what to expect

PR #1363, opened 2019-07-11
 Document CppInfo components and new attributes

danimtb:
Docs for conan-io/conan#6653
Issue conan-io/conan#5090


reference/conanfile/attributes.rst
```diff
@@ -905,10 +905,49 @@ This object should be filled in ``package_info()`` method.
...
+You can also model system dependencies for each component or just header files. With this information, generators will be able to retrieve
+the information about each component and generate different targets for each of them (Still not supported by any generator).
+
+Examples of the usage of components in generators would be:
+
+- Generate different targets with appropriate names in CMake to iterate better with ``find_package()`` functionality. This will allow more
+  granularity when linking with packages with more than one library packaged.
+- Generate files with appropriate names for CMake config and package config files like: *xxx-config.cmake* or *xxx.pc*
+
+The structure of the ``Component`` object is the same as the one used by the ``cpp_info`` object and has **the same default directories**.
```
memsharded: 2 important notes are needed here:

- That no global definition is possible if you are using components
- That the feature so far is not fully implemented, and at the moment is is only the definition in the conanfile, but is transparent and has no effect on consumers at all, and that is ongoing work.

reference/conanfile/attributes.rst
```diff
@@ -905,10 +905,49 @@ This object should be filled in ``package_info()`` method.
...
 +----------------------------------+---------------------------------------------------------------------------------------------------------+
+| self.cpp_info.components         | **[Experimental]** Dictionary with different components a package may have: libraries, executables...   |
++----------------------------------+---------------------------------------------------------------------------------------------------------+
 
 The paths of the directories in the directory variables indicated above are relative to the
 :ref:`self.package_folder<folders_attributes_reference>` directory.
 
+.. _cpp_info_components_attributes_reference:
+
+Components
```
memsharded: Actually I would prefer the reference the other way round:

- The attribute is only a placeholder, with a link
- The ``package_info()`` method is the section of the docs that contains all the detailed information about cpp_info and components.

Not to be changed now, but to have the idea in mind for the future.

reference/conanfile/attributes.rst
```diff
@@ -905,10 +905,49 @@ This object should be filled in ``package_info()`` method.
...
+        self.cpp_info.components["ssl"].libs = ["libssl"]
+
+You can also model system dependencies for each component or just header files. With this information, generators will be able to retrieve
+the information about each component and generate different targets for each of them (Still not supported by any generator).
+
+Examples of the usage of components in generators would be:
+
+- Generate different targets with appropriate names in CMake to iterate better with ``find_package()`` functionality. This will allow more
+  granularity when linking with packages with more than one library packaged.
+- Generate files with appropriate names for CMake config and package config files like: *xxx-config.cmake* or *xxx.pc*
```
jgsogo: I'm not sure this is something we should mention. Probably all components should go in the same file (is it not possible for any generator?). If the components generate their own file, we can have collisions and Conan won't be able to handle them.

And, as this is totally speculative, I would vote to remove this _example_

reference/conanfile/attributes.rst
danimtb: Ok, will take this into account for next refactors of the documentation

reference/conanfile/attributes.rst
danimtb: You are right. This was here when the feature was introduced together with the ``cpp_info.name`` attribute. But I think it is better to remove this from the reference right now and explain it better in a dedicated section in the future. Thanks!

memsharded: Yes, please remove that.

reference/conanfile/attributes.rst
```diff
@@ -970,6 +1016,10 @@ root folder of the package:
             # Get the sharedlinkflags property from OpenSSL package
             self.deps_cpp_info["openssl"].sharedlinkflags
 
+.. seealso::
+
+    Read more about :ref:`cpp_info_components_attributes_reference` in ``self.cpp_info``.
```
jgsogo: This is linking to this same section 🤔  Probably we can remove this `seealso` section, right?

danimtb: you are right

fulara:
These docs dont explain at all how to use these paths in the  consuming recipes,  they are very arbitraty.

The documentation states:
```
    def build(self):
        # Get the include directories of the SSL component of openssl package
        self.deps_cpp_info["openssl"].components["ssl"].include_paths
```
And it does not link to anything anywhere how to propagate these in cmake or  any generator.
I tried searching by lib_paths include_paths  but to my knowledge there is no such documentation.

Can we get some real example somewhere please?

comment really to the work itself:
On top of that even though the package offers consumers if in the clients package someone uses `CONAN_PKG::providing_pkg` 
they will depend on all of those

PR #1365, opened 2019-07-11
fix typo in def build(self)

ddocky:
I think existing_binaries.rst was missing a variable name in the HelloConan build() method.


memsharded:
Yes, you are right, thanks very much for fixing this!

PR #1373, opened 2019-07-23
- fix quotes

SSE4:


memsharded:
It seems it should go to master instead, not to develop?

SSE4:
changed to master

PR #1384, opened 2019-08-04
Correct binary build path in doc

raulh39:
The hello world project doesn't store the binary "greet" (or "greet.exe") in the build directory hello/bin. I think it stores it directly in the directory bin.
So in order for the recipe to work this change has to be made. After it, the console shows 
     Hello/0.1@user/testing package(): Packaged 1 file: greet
but before the fix that line it's not shown.


memsharded:
You are right, thanks for fixing this!

PR #1390, opened 2019-08-17
docs for renaming a file or folder with retrying.

yangcha:
Changelog: (Feature): os.rename() function frequently raises "Access is denied" exception on windows. This PR provides a function to rename file or folder with retrying to avoid the exception on windows.


czoido:
Hi! We are closing this PR as we are no longer merging changes for the Conan 1.x documentation. Thank you very much for your time and effort!

PR #1397, opened 2019-08-23
Remove redundant "by default" from virtualenv section

jamcodes:
The paragraph describing virtualenvs contained the following sentence: "By default, Conan does not modify by default the environment (...)". The second redundant "by default" was removed.


memsharded:
Thanks for fixing this @jamcodes!

PR #1415, opened 2019-09-13
Some recommendations to manage sanitizers

### LINKED ISSUE #1412: Document recommendations to manage sanitizers
Follow this thread:
https://github.com/conan-io/conan/issues/1346#issuecomment-530265900

We don't have a clear answer but we can document some recommendations/choices. I think that is the best we can do at the moment.

Should it be a how to? Anyway link to the section: https://docs.conan.io/en/latest/extending/custom_settings.html#adding-new-settings so people can understand how the settings-yml extension works.

czoido:
A brief guide on how sanitizers can be used in Conan. 
Closes #1412


extending.rst
```diff
@@ -11,3 +11,4 @@ This section provides an introduction to extension capabilities of Conan:
    extending/custom_settings
    extending/python_requires
    extending/hooks
+   extending/sanitizers
```
danimtb: I think this should go inside "custom settings"

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
+.. _sanitizers:
+
+Compiler sanitizers
+===================
+
+Sanitizers are tools that can detect bugs such as buffer overflows or accesses, dangling pointer or
+different types of undefined behavior.
+
+The two compilers that mainly support sanitizing options are gcc and clang. These options are
+passed to the compiler as flags and  depending on if you are using
```
danimtb:
```suggestion
passed to the compiler as flags and, depending on if you are using
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+`clang <https://clang.llvm.org/docs/UsersManual.html#controlling-code-generation>`_ or
+`gcc <https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html>`_ different sanitizers are
```
danimtb:
```suggestion
`gcc <https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html>`_, different sanitizers are
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+supported.
+
+Here we explain different options on how to model and use sanitizers with your Conan packages.
+
+Adding custom settings
+----------------------
+
+If you want to model the sanitizer options so that the package id is affected by them you have to
```
danimtb:
```suggestion
If you want to model the sanitizer options so that the package id is affected by them, you have to
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. code-block:: yaml
+    :caption: *settings.yml*
+
+    apple-clang:
+        version: ["5.0", "5.1", "6.0", "6.1", "7.0", "7.3", "8.0", "8.1", 
+                  "9.0", "9.1", "10.0", "11.0"]
+        libcxx: [libstdc++, libc++]
+        cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]
+        sanitizer: [None, Address, Thread, UndefinedBehavior, AddressUndefinedBehavior]
```
danimtb: please highlight this line in the code-block

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+
+Here you have modeled the use of ``-fsanitize=address``, ``-fsanitize=thread``,
+``-fsanitize=memory``, ``-fsanitize=undefined`` and the combination of ``-fsanitize=address`` and
+``-fsanitize=undefined``. Note that for example, for clang it is not possible to combine more than
+one of the ``-fsanitize=address``, ``-fsanitize=thread``, and ``-fsanitize=memory`` checkers in the
+same program.
+
+Adding thread sanitizer for a ``conan install``, in this case, could be done by calling ``conan
```
danimtb: Use `:command:` label for command line commands

```suggestion
Adding thread sanitizer for a :command:`conan install`, in this case, could be done by calling :command:`conan
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+install .. -s compiler.sanitizer=Thread``
```
danimtb:
```suggestion
install .. -s compiler.sanitizer=Thread`
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+    os_build=Macos
+    arch=x86_64
+    arch_build=x86_64
+    compiler=apple-clang
+    compiler.version=10.0
+    compiler.libcxx=libc++
+    build_type=Release
+    compiler.sanitizer=Address
+    [env]
+    CXXFLAGS=-fsanitize=address
```
danimtb: highlight the lines here

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+    CFLAGS=-fsanitize=address
+
+Then calling to ``conan create . -pr address_sanitizer_profile`` would inject ``-fsanitize=address``
```
danimtb:
```suggestion
Then calling to :command:`conan create . -pr address_sanitizer_profile` would inject :command:`-fsanitize=address`
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,167 @@
...
+to the build through the ``CXXFLAGS`` environment variable.
+
+Managing sanitizer settings with the build system
+#################################################
+
+Another option is to make use of the information that is propagated to the *conan generator*. For
+example, if we are using CMake we could use the information from the ``CMakeLists.txt`` to append
```
danimtb:
```suggestion
example, if we are using CMake we could use the information from the *CMakeLists.txt* to append
```

extending.rst
danimtb: Ok, now I understand why it is not there. Maybe this should not be in extending Conan but in the "how to" section

extending/sanitizers.rst
czoido: Done, thanks!

extending.rst
czoido: I think the "how to" section would be a good place for this information too.
Maybe we should move it...

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+----------------------
+
+If you want to model the sanitizer options so that the package id is affected by them, you have to
+introduce new settings in the *settings.yml* file (see :ref:`custom_settings` section for more
+information).
+
+Sanitizer options should be modeled as sub-settings of the compiler. Depending on how you want to
+combine the sanitizers you have two choices.
+
+Adding a list of commonly used options
```
lasote: Don't use "options" term here, could be confusing. 

```suggestion
Adding a list of commonly used values
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+######################################
+
+If you have a fixed set of sanitizers or combinations of them that are the ones you usually set for
+your builds you can add the sanitizers as a list of options. An example for *apple-clang* would be
```
lasote:
```suggestion
your builds you can add the sanitizers as a list of values. An example for *apple-clang* would be
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+Here you have modeled the use of ``-fsanitize=address``, ``-fsanitize=thread``,
+``-fsanitize=memory``, ``-fsanitize=undefined`` and the combination of ``-fsanitize=address`` and
+``-fsanitize=undefined``. Note that for example, for clang it is not possible to combine more than
+one of the ``-fsanitize=address``, ``-fsanitize=thread``, and ``-fsanitize=memory`` checkers in the
+same program.
+
+Adding thread sanitizer for a :command:`conan install`, in this case, could be done by calling
+:command:`conan install .. -s compiler.sanitizer=Thread`
+
+Adding different options to combine
```
lasote:
```suggestion
Adding different values to combine
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+###################################
+
+Another option would be to add the sanitizer options as multiple ``True`` or ``None`` fields so that
```
lasote:
```suggestion
Another option would be to add the sanitizer values as multiple ``True`` or ``None`` fields so that
```

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+.. code-block:: yaml
+    :caption: *settings.yml*
+    :emphasize-lines: 6,7,8
+
+    apple-clang:
+        version: ["5.0", "5.1", "6.0", "6.1", "7.0", "7.3", "8.0", 
+                  "8.1", "9.0", "9.1", "10.0", "11.0"]
+        libcxx: [libstdc++, libc++]
+        cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]
+        address_sanitizer: [None, True]
```
lasote: Why not False, True?

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+        thread_sanitizer: [None, True]
+        undefined_sanitizer: [None, True]
+
+Then, you can add different sanitizers calling, for example, to :command:`conan install ..
+-s compiler.address_sanitizer=True -s compiler.undefined_sanitizer=True`
```
lasote: A drawback of this approach is that not all the combinations will be valid or will make sense, but it is up to the consumer to use it correctly.

extending/sanitizers.rst
```diff
@@ -0,0 +1,170 @@
...
+
+Passing the information to the compiler or build system
+-------------------------------------------------------
+
+Here again, we have multiple choices to pass sanitizers information to the compiler or build system.
+
+Using from custom profiles
+##########################
+
+It is possible to have different custom profiles defining the compiler sanitizer option and
```
lasote:
```suggestion
It is possible to have different custom profiles defining the compiler sanitizer setting and
```

extending/sanitizers.rst
czoido: Thank you, I think that's worth mentioning

PR #1416, opened 2019-09-16
Add info --paths example

uilianries:
The command `info --paths` is not examplyfied in the docs.


reference/commands/consumer/info.rst
```diff
@@ -199,3 +199,29 @@ files in the root of the configuration folder:
 
 It is not necessary to modify the generated html file. Conan will automatically use the local paths to the cache files if
 present, or the internet ones if not.
+
+You can find where the package is installed in your cache by using the argument ``--paths``:
```
danimtb:
```suggestion
You can find where the package is installed in your cache by using the argument :command:`--paths`:
```

uilianries: Fix argument for conan info command

PR #1419, opened 2019-09-19
Update logging level names

uilianries:
Related PR: https://github.com/conan-io/conan/pull/5772


Minimonium:
It'd be cool to have some examples of what we can expect to get from these log levels.

jgsogo:
Yes, @Minimonium , it would be nice, but I think that the two files modified in this PR (up to now) aren't the place to add this information... those are sections inside already long files.

I would suggest improving the `Logging` one (https://docs.conan.io/en/latest/mastering/logging.html): there we talk about `CONAN_TRACE_FILE` but we say nothing about the logging level and it would probably make a lot of sense, wdyt @danimtb ?

danimtb:
Yes, makes sense to have that in the logging one

uilianries:
> Yes, @Minimonium , it would be nice, but I think that the two files modified in this PR (up to now) aren't the place to add this information... those are sections inside already long files.

I gonna open a new issue to introduce those examples.

PR #1429, opened 2019-10-01
Review extract version documentation

david-sinuela-pix4d:
The notes about returning None to force reading the version information
from the package metadata don't seem correct. From experience the
version always comes from the package metadata once it has been
exported.

Simple correction, please let me know if something more elaborated is desired.

### DIFF
--- a/howtos/capture_version.rst
+++ b/howtos/capture_version.rst
@@ -24,10 +24,8 @@ the *conanfile.py* recipe resides, and use it to define the version of the Conan
             ...
 
 In this example, the package created with :command:`conan create` will be called 
-``Hello/branch_commit@user/channel``. Note that ``get_version()`` returns ``None``
-if it is not able to get the Git data. This is necessary when the recipe is already in the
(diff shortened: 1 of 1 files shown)


memsharded:
Yes, probably it is poorly worded (language issue). 
But it is still very important that the ``get_version()`` functions do not raise when they are in the cache, and return ``None`` instead, because those functions are indeed evaluated when the recipe is parsed, as they are class attributes. Don't know how that could be explained, but I think it is worth trying to explicitly explain that somewhere in that doc page.

david-sinuela-pix4d:
Right, that is important, I'll iterate on the explanation.

The question comes from some internal discussion we are having about how to set the version of a recipe. We are considering different alternatives with varying flexibility: from an environment variable, a text file or git describe. The conversation would be simpler if we knew exactly when the version is evaluated (or rather saved).

david-sinuela-pix4d:
Second attempt up!

david-sinuela-pix4d:
In any case feel free to take over and adjust it to the writing style of the rest of the documentation.

memsharded:
Thanks!

PR #1457, opened 2019-10-25
Suggested improved description of bindir

solvingj:
Looks like it was erroneously copy/pasted from includedirs

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -862,7 +862,7 @@ This object should be filled in ``package_info()`` method.
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
 | self.cpp_info.resdirs          | Ordered list of resource (data) paths. Defaulted to ``["res"]``                                         |
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


reference/conanfile/attributes.rst
```diff
@@ -862,7 +862,7 @@ This object should be filled in ``package_info()`` method.
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
 | self.cpp_info.resdirs          | Ordered list of resource (data) paths. Defaulted to ``["res"]``                                         |
 +--------------------------------+---------------------------------------------------------------------------------------------------------+
-| self.cpp_info.bindirs          | Ordered list with include paths. Defaulted to ``["bin"]``                                               |
+| self.cpp_info.bindirs          | Ordered list with binary paths, such as executables and/or dynamic libraries on Windows. Defaulted to ``["bin"]``                                               |
```
jgsogo: The suggested line will match the table wide 😉 

```suggestion
| self.cpp_info.bindirs          | Ordered list with paths to binaries (executables, dynamic libraries,...). Defaulted to ``["bin"]``      |
```

PR #1461, opened 2019-10-27
remove * from the first search example

amalulla:
both
`conan search 'Poco*' --remote=conan-center`
and 
`conan search Poco --remote=conan-center` 
would work here, but the latter is more straight-forward and the page explains the Search command further later on


danimtb:
Thanks for the improvement! 🙌

PR #1465, opened 2019-10-28
Docs for cpp_info.build_modules

danimtb:
Docs for  conan-io/conan#5940
Issue conan-io/conan#5798


reference/conanfile/methods.rst
```diff
@@ -238,6 +239,8 @@ The ``cpp_info`` attribute has the following properties you can assign/append to
 - **srcdirs**: List of relative paths (starting from the package root) of directories in which to find sources (like
   .c, .cpp). By default it is empty. It might be used to store sources (for later debugging of packages, or to reuse those sources building
   them in other packages too).
+- **build_modules**: List of relative paths to build system related utility module files created by the package. Used by CMake generators to
```
lasote: I would explain here that currently only CMake is supported and that the files will be automatically included with the cmake generator when calling conan basic setup and when find_package is run for the others.

danimtb: Done!

PR #1468, opened 2019-10-29
compatible ids

memsharded:
Docs for https://github.com/conan-io/conan/pull/5837

### DIFF
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -216,7 +216,6 @@ step is processed after the whole dependency graph has been built, so it is not 
 based on this compatibility model, which only affects to binaries that can be replaced.
 
 
-
 .. _problem_of_dependencies:
 
 Dependency Issues
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -178,7 +178,7 @@ With ``CompatiblePackage`` we can define an ordered list of compatible packages,
 the package ID that our profile defines is not available. Let's see it with an example:
 
 Lets say that we are building with a profile of ``gcc 4.9``. But for a given package we want to
-fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7``, if we cannot find a binary built with ``gcc 4.9``.
+fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7`` if we cannot find a binary built with ``gcc 4.9``.
 That can be defined as:
 
 .. code-block:: python
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -205,7 +205,7 @@ It is the responsibility of the developer to guarantee that such binaries are in
                 self.compatible_packages.append(compatible_pkg)
 
 
-This defines that the binaries are compatible with binaries of itself built with a lower optimization value. In this example we can
+This recipe defines that the binaries are compatible with binaries of itself built with a lower optimization value. It can
 have up to 3 different binaries, one for each different value of ``optimized`` option. The ``package_id()`` defines that a binary 
 built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
 in their configuration. But a binary built with ``optimized=2``, will not be considered if the input is ``optimized=1``.
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -187,7 +187,7 @@ Note that if the input configuration is ``gcc 4.8``, it will not try to fallback
 condition is not met. Also, this approach is very different from the documented above, because it is possible to have
 distinct binaries for the different versions, while the above, which is mainly an erasure, maintains 1 binary for all of them.
 
-The ``CompatiblePackage`` class contains ``settings``, ``options`` and ``requires`` fields, as copies, so they can be modified.
+The ``CompatiblePackage()`` inherits the values of ``settings``, ``options`` and ``requires`` as attributes so they can be modified to model the compatibility.
 
 It is the responsibility of the developer to guarantee that such binaries are indeed compatible. For example in:
 
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -198,7 +198,7 @@ That can be defined as:
 Note that if the input configuration is ``gcc 4.8``, it will not try to fallback to binaries of ``gcc 4.7`` as the
 condition is not met.
 
-The ``CompatiblePackage()`` inherits the values of ``settings``, ``options`` and ``requires`` as attributes so they can be modified to model the compatibility.
+The ``CompatiblePackage()`` copies the values of ``settings``, ``options`` and ``requires`` from the current instance of the recipe so they can be modified to model the compatibility.
 
 It is the responsibility of the developer to guarantee that such binaries are indeed compatible. For example in:
 
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -221,7 +221,7 @@ have up to 3 different binaries, one for each different value of ``optimized`` o
 built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
 in their configuration. But a binary built with ``optimized=2`` will not be considered if the requested one is ``optimized=1``.
 
-The binary **should be interchangeable** at all effects. This also applies to other usages of that configuration. If this example used
+**The binary should be interchangeable at all effects**. This also applies to other usages of that configuration. If this example used
 the ``optimized`` option to conditionally require different dependencies, that will not be taken into account. The ``package_id()``
 step is processed after the whole dependency graph has been built, so it is not possible to define how dependencies are resolved
 based on this compatibility model, it only applies to use-cases where the binaries can be *interchanged*.
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -213,7 +213,7 @@ in their configuration. But a binary built with ``optimized=2``, will not be con
 The binary **should be interchangeable** at all effects. This also applies to other usages of that configuration. If this example used
 the ``optimized`` option to conditionally require different dependencies, that will not be taken into account. The ``package_id()``
 step is processed after the whole dependency graph has been built, so it is not possible to define how dependencies are resolved
-based on this compatibility model, which only affects to binaries that can be replaced.
+based on this compatibility model, which only applies to cases where the binaries that can be *interchanged*.
 
 
 .. _problem_of_dependencies:
--- a/creating_packages/define_abi_compatibility.rst
+++ b/creating_packages/define_abi_compatibility.rst
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
         - Recipes packaging **header only** libraries.
         - Adjusting **Visual Studio toolsets** compatibility.
 
+
+.. _compatible_packages:
+
+
+Compatible Packages
+-------------------
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+When a package ID is not found, it is possible to define an ordered list of compatible packages.
+Lets say that we are building with a profile of ``gcc 4.9``. But for a given package we want to
+fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7``, if we cannot find a binary built with ``gcc 4.9``.
+That can be defined as:
+
+.. code-block:: python
+
+    from conans import ConanFile, CompatiblePackage
+
+    class Pkg(ConanFile):
+        settings = "os", "compiler", "arch", "build_type"
+        def package_id(self):
+            if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+                for version in ("4.8", "4.7"):
+                    compatible_pkg = CompatiblePackage(self)
+                    compatible_pkg.settings.compiler.version = version
+                    self.compatible_packages.append(compatible_pkg)
+
+Note that if the input configuration is ``gcc 4.8``, it will not try to fallback to binaries of ``gcc 4.7`` as the
+condition is not true. Also, this approach is very different to the documented above, because it is possible to have
+distinct binaries for the different versions, while the above, which is mainly an erasure, maintains 1 binary for all of them.
+
+The ``CompatiblePackage`` class contains ``settings``, ``options`` and ``requires`` fields, as copies, so they can be modified.
+
+It is the responsibility of the developer to guarantee that such binaries are indeed compatible. For example in:
+
+.. code-block:: python
+
+    from conans import ConanFile, CompatiblePackage
+
+    class Pkg(ConanFile):
+        options = {"optimized": [1, 2, 3]}
+        default_options = {"optimized": 1}
+        def package_id(self):
+            for optimized in range(int(self.options.optimized), 0, -1):
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.options.optimized = optimized
+                self.compatible_packages.append(compatible_pkg)
+
+
+This defines that the binaries are compatible with binaries of itself built with a lower optimization value. In this example we can
+have up to 3 different binaries, one for each different value of ``optimized`` option. The ``package_id()`` defines that a binary 
+built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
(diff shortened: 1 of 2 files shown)


creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+Lets say that we are building with a profile of ``gcc 4.9``. But for a given package we want to
+fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7``, if we cannot find a binary built with ``gcc 4.9``.
+That can be defined as:
+
+.. code-block:: python
+
+    from conans import ConanFile, CompatiblePackage
+
+    class Pkg(ConanFile):
+        settings = "os", "compiler", "arch", "build_type"
```
danimtb: Add space for readability
```suggestion
        settings = "os", "compiler", "arch", "build_type"
        
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+        def package_id(self):
+            if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+                for version in ("4.8", "4.7"):
+                    compatible_pkg = CompatiblePackage(self)
+                    compatible_pkg.settings.compiler.version = version
+                    self.compatible_packages.append(compatible_pkg)
+
+Note that if the input configuration is ``gcc 4.8``, it will not try to fallback to binaries of ``gcc 4.7`` as the
+condition is not true. Also, this approach is very different to the documented above, because it is possible to have
```
danimtb:
```suggestion
condition is not met. Also, this approach is very different from the documented above, because it is possible to have
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+distinct binaries for the different versions, while the above, which is mainly an erasure, maintains 1 binary for all of them.
+
+The ``CompatiblePackage`` class contains ``settings``, ``options`` and ``requires`` fields, as copies, so they can be modified.
```
danimtb: - `CompatiblePackage()` ?
- ~fields~ --> attributes?
- as copies --> inherit?
```suggestion
The ``CompatiblePackage()`` inherits the values of ``settings``, ``options`` and ``requires`` as attributes so they can be modified to model the compatibility.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+        options = {"optimized": [1, 2, 3]}
+        default_options = {"optimized": 1}
+        def package_id(self):
+            for optimized in range(int(self.options.optimized), 0, -1):
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.options.optimized = optimized
+                self.compatible_packages.append(compatible_pkg)
+
+
+This defines that the binaries are compatible with binaries of itself built with a lower optimization value. In this example we can
```
danimtb:
```suggestion
This recipe defines that the binaries are compatible with binaries of itself built with a lower optimization value. It can
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+have up to 3 different binaries, one for each different value of ``optimized`` option. The ``package_id()`` defines that a binary 
+built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
+in their configuration. But a binary built with ``optimized=2``, will not be considered if the input is ``optimized=1``.
+
+The binary should be interchangeable at all effects. This also applies for other usage of that configuration. If this example used
```
danimtb: Highlight
```suggestion
The binary **should be interchangeable** at all effects. This also applies to other usages of that configuration. If this example used
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+the ``optimized`` option to conditionally require different dependencies, that will not be taken into account. The ``package_id()``
+step is processed after the whole dependency graph has been built, so it is not possible to define how dependencies are resolved
+based on this compatibility model, which only affects to binaries that can be replaced.
```
danimtb:
```suggestion
based on this compatibility model, which only applies to cases where the binaries that can be *interchanged*.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,69 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+
```
danimtb:
```suggestion
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,38 @@ even if it matches with the default of the compiler being used:
...
+
+If the binary of that package ID is not found, Conan allows to define an ordered list of compatible package IDs, of other configurations
+that are fully binary compatible and can be used as a fallback. The syntax to do this is:
+
+.. code-block:: python
+
+    from conans import ConanFile, CompatiblePackage
+
+    class Pkg(ConanFile):
+        settings = "os", "compiler", "arch", "build_type"
```
danimtb:
```suggestion
        settings = "os", "compiler", "arch", "build_type"

```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,38 @@ even if it matches with the default of the compiler being used:
...
+                self.compatible_packages.append(compatible_pkg)
+
+This will define that if we try to install this package with ``gcc 4.9`` and there is not binary for that configuration, it will check
+if there is an available binary built with ``gcc 4.8`` and use it if existing. But not the other way round.
+
+
+.. seealso::
+
+    For more information about :ref:`CompatiblePackage read this <compatible_package>`
+
```
danimtb:
```suggestion
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,38 @@ even if it matches with the default of the compiler being used:
...
+        settings = "os", "compiler", "arch", "build_type"
+        def package_id(self):
+            if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.settings.compiler.version = "4.8"
+                self.compatible_packages.append(compatible_pkg)
+
+This will define that if we try to install this package with ``gcc 4.9`` and there is not binary for that configuration, it will check
+if there is an available binary built with ``gcc 4.8`` and use it if existing. But not the other way round.
+
```
danimtb:
```suggestion
```

memsharded:
Merged your suggestions. 

The self.compatible_packages is not documented on purpose. Besides the documented way to do it, I don't want to define any further detail, if its a list or not. The only thing that can be done with it is the provided code, nothing else.

danimtb:
We already have similar attributes documented with these restrictions:

> cpp_info
>
> Important
> This attribute is only defined inside package_info() method being None elsewhere.

> user_info
> 
> This attribute is only defined inside package_info() method, being None elsewhere, so please use it only inside this method.

> info
> 
> Object used to control the unique ID for a package. Check the package_id() to see the details of the self.info object.

I think it is better to explain the intended usage than to hide it and let people experiment with it

jgsogo:
I would move the motivation to the top of the section in the "Define ABI Compatibility". I think that this sentence is the key one and should be at the very beginning (with a little rewording it should be a good introduction): _"this approach is very different from the documented above, because it is possible to have distinct binaries for the different versions, while the above, which is mainly an erasure, maintains 1 binary for all of them."_

memsharded:
You hit the point. ``conan info`` attribute doesn't document that it cannot be used in other methods, and that is a mistake. Yes, can be fixed in the docs, but that is the issue. If you only document it in the method that can be used there is no space for confusion.

We cannot even say that ``compatible_packages`` will be None elsewhere, because it is already initialize to a list to precisely avoid the linter errors that later users complain about. So the docs you are suggesting are: "compatible_packages: don't use it anywhere but in the package_id() method, go and look the package_id() method docs about how to use it"?  Should we add a experimental tag to this attribute too then, or rely on the ``package_id()`` section one? And if users search for it, and find this instead of the package_id() one, then they will need just another hop? Sure, can be done, but I see little value in it.

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+Using ``CompatiblePackage``, it is possible to define compatible binaries that have different package IDs. With
+this ``CompatiblePackage`` feature, it is possible to have a different binary for each ``gcc`` version, so the 
+``gcc 4.8`` package will be a different one with a different package ID than the ``gcc 4.9`` one, and still define
+that you can use the ``gcc 4.8`` package when building with ``gcc 4.9``.
+
+With ``CompatiblePackage`` we can define an ordered list of compatible packages, that will be checked in order if
+the package ID that our profile defines is not available. Let's see it with an example:
+
+Lets say that we are building with a profile of ``gcc 4.9``. But for a given package we want to
+fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7``, if we cannot find a binary built with ``gcc 4.9``.
```
jgsogo:
```suggestion
fallback to binaries built with ``gcc 4.8`` or ``gcc 4.7`` if we cannot find a binary built with ``gcc 4.9``.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+        
+        def package_id(self):
+            if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+                for version in ("4.8", "4.7"):
+                    compatible_pkg = CompatiblePackage(self)
+                    compatible_pkg.settings.compiler.version = version
+                    self.compatible_packages.append(compatible_pkg)
+
+Note that if the input configuration is ``gcc 4.8``, it will not try to fallback to binaries of ``gcc 4.7`` as the
+condition is not met. m.
```
jgsogo:
```suggestion
condition is not met.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+
+The ``CompatiblePackage()`` inherits the values of ``settings``, ``options`` and ``requires`` as attributes so they can be modified to model the compatibility.
```
jgsogo:
```suggestion
The ``CompatiblePackage()`` copies the values of ``settings``, ``options`` and ``requires`` from the current instance of the recipe so they can be modified to model the compatibility.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+            for optimized in range(int(self.options.optimized), 0, -1):
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.options.optimized = optimized
+                self.compatible_packages.append(compatible_pkg)
+
+
+This recipe defines that the binaries are compatible with binaries of itself built with a lower optimization value. It can
+have up to 3 different binaries, one for each different value of ``optimized`` option. The ``package_id()`` defines that a binary 
+built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
+in their configuration. But a binary built with ``optimized=2``, will not be considered if the input is ``optimized=1``.
```
jgsogo:
```suggestion
in their configuration. But a binary built with ``optimized=2`` will not be considered if the requested one is ``optimized=1``.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+
+The binary **should be interchangeable** at all effects. This also applies to other usages of that configuration. If this example used
+the ``optimized`` option to conditionally require different dependencies, that will not be taken into account. The ``package_id()``
+step is processed after the whole dependency graph has been built, so it is not possible to define how dependencies are resolved
+based on this compatibility model, which only applies to cases where the binaries that can be *interchanged*.
```
jgsogo:
```suggestion
based on this compatibility model, it only applies to use-cases where the binaries can be *interchanged*.
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -153,6 +153,80 @@ The same way we have adjusted the ``self.info.settings``, we could set the ``sel
...
+                compatible_pkg.options.optimized = optimized
+                self.compatible_packages.append(compatible_pkg)
+
+
+This recipe defines that the binaries are compatible with binaries of itself built with a lower optimization value. It can
+have up to 3 different binaries, one for each different value of ``optimized`` option. The ``package_id()`` defines that a binary 
+built with ``optimized=1`` can be perfectly linked and will run even if someone defines ``optimized=2``, or ``optimized=3``
+in their configuration. But a binary built with ``optimized=2``, will not be considered if the input is ``optimized=1``.
+
+The binary **should be interchangeable** at all effects. This also applies to other usages of that configuration. If this example used
```
jgsogo:
```suggestion
**The binary should be interchangeable at all effects**. This also applies to other usages of that configuration. If this example used
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,37 @@ even if it matches with the default of the compiler being used:
...
 Same behavior applies if you use the deprecated setting ``cppstd``.
 
+
+CompatiblePackage
+^^^^^^^^^^^^^^^^^
+The ``package_id()`` method serves to define the "canonical" binary package ID, the identifier of the binary that correspond to the
+input configuration of settins and options. This canonical binary package ID will be always computed, and Conan will check for its
+existence to be downloaded and installed.
+
+If the binary of that package ID is not found, Conan allows to define an ordered list of compatible package IDs, of other configurations
```
jgsogo:
```suggestion
If the binary of that package ID is not found, Conan lets the recipe writer define an ordered list of compatible package IDs, of other configurations
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,37 @@ even if it matches with the default of the compiler being used:
...
+that are fully binary compatible and can be used as a fallback. The syntax to do this is:
```
jgsogo:
```suggestion
that should be binary compatible and can be used as a fallback. The syntax to do this is:
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,37 @@ even if it matches with the default of the compiler being used:
...
+    class Pkg(ConanFile):
+        settings = "os", "compiler", "arch", "build_type"
+
+        def package_id(self):
+            if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.settings.compiler.version = "4.8"
+                self.compatible_packages.append(compatible_pkg)
+
+This will define that if we try to install this package with ``gcc 4.9`` and there is not binary for that configuration, it will check
```
jgsogo:
```suggestion
This will define that, if we try to install this package with ``gcc 4.9`` and there isn't a binary available for that configuration, Conan will check
```

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,37 @@ even if it matches with the default of the compiler being used:
...
+if there is an available binary built with ``gcc 4.8`` and use it if existing. But not the other way round.
```
jgsogo:
```suggestion
if there is one available built with ``gcc 4.8`` and use it. But not the other way round.
```

lasote:
fix this: `/home/travis/build/conan-io/docs/reference/conanfile/methods.rst:927:undefined label: compatible_package (if the link has no caption the label must precede a section header)`

reference/conanfile/methods.rst
```diff
@@ -894,6 +894,37 @@ even if it matches with the default of the compiler being used:
...
+                compatible_pkg = CompatiblePackage(self)
+                compatible_pkg.settings.compiler.version = "4.8"
+                self.compatible_packages.append(compatible_pkg)
+
+This will define that, if we try to install this package with ``gcc 4.9`` and there isn't a binary available for that configuration, Conan will check
+if there is one available built with ``gcc 4.8`` and use it. But not the other way round.
+
+.. seealso::
+
+    For more information about :ref:`CompatiblePackage read this <compatible_package>`
```
jgsogo:
```suggestion
    For more information about :ref:`CompatiblePackage read this <compatible_packages>`
```

PR #1475, opened 2019-11-03
Document the intent of conan_output_dirs_setup()

### LINKED ISSUE #1474: Document the intent of conan_output_dirs_setup()
As discussed in https://github.com/conan-io/conan/issues/6012, the intent of `conan_output_dirs_setup()` and `NO_OUTPUT_DIRS` is somewhat vague.

Currently the docs say:
> (Optional): Do not adjust the output directories. Deactivates the call to the macro conan_output_dirs_setup().

The fact that the output directories change in the first place is odd, so I think we should be more clear about (1) why this happens by default, (2) the side effects of turning it off, and (3) potentially recommend using `NO_OUTPUT_DIRS` for new projects.

peter-moran:
Resolves #1474

### DIFF
--- a/reference/generators/cmake.rst
+++ b/reference/generators/cmake.rst
@@ -141,7 +141,7 @@ variables. See the macros below for detailed information.
 Parameters:
     - ``TARGETS`` (Optional): Setup all the CMake variables by target (only CMake > 3.1.2). Activates the call to the macro
       ``conan_target_link_libraries()``.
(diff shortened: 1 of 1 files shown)


lasote:
Hi! Could you sign the CLA? Thanks for your help!

peter-moran:
Sorry about that, I've signed it now.

PR #1483, opened 2019-11-26
Change create command build policy documentation

czoido:
As explained in https://github.com/conan-io/conan/pull/6131 the --build policy help message was the same for the commands create, test, install and workspace but the behavior for the create command by default was not to behave as `--build=never` but as `--build=package name` which could be misleading.


danimtb:
`reference` should be clear enough for people to understand that it refers to the reference name or should it be `<reference>`?

Is that really the reference or the reference name?

czoido:
> `reference` should be clear enough for people to understand that it refers to the reference name or should it be `<reference>`?
> 
> Is that really the reference or the reference name?

Yes, I think that It will be more clear with package name, reference is confusing

Issue #1506, opened 2019-12-17
Troubleshooting: Case insensitive filesystem can't manage this


uilianries:
When a package with same name is installed, but using different name is installed we have the follow problem:

```
ERROR: Failed requirement 'OpenSSL/1.0.2o@conan/stable' from 'Poco/1.9.3@pocoproject/stable'
ERROR: Requested 'OpenSSL/1.0.2o@conan/stable' but found case incompatible 'openssl'
Case insensitive filesystem can't manage this
```

The recommendation is excluding the old package.

PR #1507, opened 2019-12-17
#1506 Add Troubleshooting for duplicated package

uilianries:
Add topic describing the situation when a different package has the same name.

closes https://github.com/conan-io/docs/issues/1506


uilianries:
Ops! there is an error:

/home/travis/build/conan-io/docs/faq/troubleshooting.rst:189:Could not lex literal_block as "bash". Highlighting skipped.

Let me fix it first

uilianries:
Done!

uilianries:
Up! @memsharded @jgsogo could you please merge this one?

uilianries:
The URL https://msbuildlog.com is down (403), which is causing an error on CI 

https://www.isitdownrightnow.com/msbuildlog.com.html

uilianries:
> 
> 
> **Really useful FAQ to add, thanks!**
> 
> I will ask you to change the perspective and imagine that the user is trying to install the new recipe and needs to remove the old one.
> 
> So first, `conan install poco/1.10.1@`, it should fail the other way around, and then you have to `conan remove "OpenSSL/*"`

Understood, a transitive package. Yes, it sounds good.

uilianries:
@jgsogo take a look, I think it follows your idea now

PR #1511, opened 2020-01-01
Add CppCon 2019 lightning talks that discuss Conan

ssrobins:


videos.rst
```diff
@@ -115,7 +115,25 @@ Videos and links
 
   .. raw:: html
 
-      <iframe width="560" height="315" src="https://www.youtube.com/watch?v=aqQT3J16OxI" frameborder="0" allowfullscreen>
+      <iframe width="560" height="315" src="https://www.youtube.com/embed/aqQT3J16OxI" frameborder="0" allowfullscreen>
+      </iframe>
+
+      <br/><br/>
+
+- CppCon 2019: "Why Conan? 5 reasons to use Conan package manager" by Diego Rodriguez-Losada
```
memsharded: I would exclude this one. The reason is that this is the Conan documentation. Examples, use cases, technical things, etc., are ok in the docs. This is more like a "motivational" (lets say it is more marketing than technical docs) talk, so I don't think this belongs here.

ssrobins: Fair point, fixed!

memsharded:
Thanks, merged! Will be published together with the next release.

PR #1514, opened 2020-01-08
Update cross_building.rst

haraldF:
Formula that have some generators that run on the host (e.g. protoc) require `arch_build` and `os_build` to be set in the profile. Thus, include the default profile that sets them.

### DIFF
--- a/systems_cross_building/cross_building.rst
+++ b/systems_cross_building/cross_building.rst
@@ -302,6 +302,8 @@ RPI one:
 
 .. code-block:: text
 
+    include(default)
+
     standalone_toolchain=/myfolder/arm_21_toolchain # Adjust this path
     target_host=arm-linux-androideabi
     cc_compiler=clang
@@ -338,6 +340,8 @@ match the gcc toolchain compiler:
 
 
 .. code-block:: text
+
+    include(default)
 
     standalone_toolchain=/myfolder/arm_21_toolchain
     target_host=arm-linux-androideabi


haraldF:
If there's a better way to set the `arch_build` and `os_build`, please let me know :)

systems_cross_building/cross_building.rst
```diff
@@ -302,6 +302,8 @@ RPI one:
 
 .. code-block:: text
 
+    include(default)
```
memsharded: The thing with this approach is that it relies on the profile to be correctly overwriting and not appending most of the default profile. I think I would add the ``os_build`` and ``arch_build`` explicitly in the profile, not including(default). At least in my experience it is very unusual to have a cross-building profile that is capable of cross-building with the same profile in different OSs, as the cross-compiler paths, etc are typically different in any case.

haraldF: I chose this approach as the file was indeed overwriting everything from the default profile. A nicer fix would be to allow importing only certain settings from a profile, would that be possible?

e.g.:

```
import(default).settings.os_build
# or
import settings.os_build from default
```

Another approach would be to generate a `default` and a `default_host` profile. `default_host`  could be imported in all cross-compilation profiles?

As for hard-coding, the docs would look sub-optimal because the topic is called "Linux/Windows/macOS to Android" - we'd have to split the chapter in three, or add an ugly place-holder :/

PR #1515, opened 2020-01-09
Update getting started

czoido:
This PR updates the getting started chapter in the documentation:

- Use poco package without user and channel from the conan center index.
- Update images to remove user and channel and to show the graph with the updated dependencies.
- Notes about the user and channel searching for a specific package or when returning packages from the remote.
- Update the command line output.
- Minor changes in the example.


getting_started.rst
```diff
@@ -3,11 +3,14 @@
...
 
-We'll use CMake as build system in this case but keep in mind that Conan **works with any build system** and is not limited to using CMake.
+We'll use CMake as build system in this case but keep in mind that Conan **works with any build
+system** and is not limited to using CMake.
 
-Make sure you are running the latest Conan version. Read the :ref:`Conan update<conan_update>` section get more information.
+Make sure you are running the latest Conan version. Read the :ref:`Conan update<conan_update>`
+section get more information.
 
 An MD5 Encrypter using the Poco Libraries
```
jgsogo: This should say "hash calculator" too, but we need to preserve the previous link (add it to the `conf.py` file)

czoido: We'll leave the conf.py as the link is an anchor and won'tfail, but I've changed the title in the other pages where it is referenced.

PR #1528, opened 2020-01-23
Update Conan cache folder description

uilianries:
Based on follow commands:

```
$ conan --version
Conan version 1.21.1
$ export CONAN_USER_HOME=/tmp/foo
$ conan profile new default --detect
Found gcc 9.2
Found clang 9.0
gcc>=5, using the major as version

Profile created with detected settings: /tmp/foo/.conan/profiles/default
```

As you can see, `CONAN_USER_HOME` determines the base folder for conan cache folder, not the conan cache itself.

I didn't check on Windows, but I presume the result will be the same.

Related on issue https://github.com/conan-io/conan/issues/6407

### DIFF
--- a/mastering/custom_cache.rst
+++ b/mastering/custom_cache.rst
@@ -36,7 +36,7 @@ Windows users:
 .. code-block:: bash
 
    $ SET CONAN_USER_HOME=c:\data
-   $ conan install . # call conan normally, config & data will be in c:\data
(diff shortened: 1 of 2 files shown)


danimtb:
Related to conan-io/conan#5295 as well. Thanks a lot for the PR @uilianries!

PR #1542, opened 2020-02-04
Put back graph clean-modified subcommand

theodelrieu:
This reverts commit e5f252712b67d4d37d8c0c512b97810f4c08a4f0.

### DIFF
--- a/versioning/lockfiles.rst
+++ b/versioning/lockfiles.rst
@@ -102,6 +102,15 @@ There are 2 main entry points for lockfile information in conan commands:
         generate a lockfile, but the problem with :command:`conan info -if=.` is that it does not allow to 
         specify a profile or settings.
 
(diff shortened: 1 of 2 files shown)


versioning/lockfiles.rst
```diff
@@ -220,21 +229,27 @@ We can now proceed iteratively with the following procedure:
         $ cp release/conan.lock build_server_folder/release
         $ cd build_server_folder
 
-4. build the package
+4. clean "modified" nodes from the lockfile
```
memsharded: Not necessary here. You might want to keep the history of modifications to this lockfile at this point, as this node should probably not rebuild things that were already built.

The clean is needed at the end of the iteration, when the whole graph has been built and we want to start a new iteration.

PR #1553, opened 2020-02-11
Fix virtualenv activate/deactivate filename

FernandoVelcic:
I found a mistake in the documentation. The virtualenv generator files are "activate.sh", "deactivate.sh", "activate.bat" and "deactivate.bat"

### DIFF
--- a/devtools/running_packages.rst
+++ b/devtools/running_packages.rst
@@ -63,10 +63,10 @@ This will generate a few files that can be called to activate and deactivate the
 
 .. code-block:: bash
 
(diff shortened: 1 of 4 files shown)


danimtb:
Hi @FernandoVelcic,

The documentation example is talking about the `virtualrunenv` generator, not `virtualenv`. So, for this case, the files that are generated are correctly specified as *activate_run.sh* & *deactivate_run.sh*

FernandoVelcic:
You are right! Thanks.

PR #1558, opened 2020-02-13
Add quotes to wildcards in examples to prevent shell expansion

DavidSpickett:
Without this local file names may be used:
```
$ touch foo
$ conan search *
There are no packages matching the 'foo' pattern
```

With quotes this does not happen:
```
$ conan search '*'
Existing package recipes:
<...>
```

This is to to prevent confusion as in conan-io/conan#5710


memsharded:
Hi @DavidSpickett 

This is great, thank you!

As we would like to release this change asap, could you please target the "master" branch instead?

Also, please check the above CLA, sometimes it is needed to use the same email as Github when doing the commits, so Github recognizes it.

howtos/generic_ci_artifactory.rst
```diff
@@ -91,7 +91,7 @@ is calling to the ``start`` subcommand of ``conan_build_info``. This will set th
     $ cd .. && cd consumer
     $ conan install . # Consumes mypackage, generates a lockfile
     $ conan create . consumer/1.0@user/stable --lockfile conan.lock
-    $ conan upload * -c -r local # Upload all packages to local remotes
+    $ conan upload '*' -c -r local # Upload all packages to local remotes
```
memsharded: I am afraid this is broken in Windows terminal. We need to use double quotes for compatibility

memsharded:
```
(conanpip37) C:\Users\memsharded>conan search "*"
Existing package recipes:

openssl/1.0.2t
poco/1.9.4
zlib/1.2.11

(conanpip37) C:\Users\memsharded>conan search '*'
There are no packages matching the ''*'' pattern

(conanpip37) C:\Users\Diego>```

DavidSpickett: Done. (I think, not very familiar with the pull request UI)

memsharded:
Thanks very much!

uilianries:
Thanks @DavidSpickett !

Issue #1564, opened 2020-02-18
Use alphabetical order for references


uilianries:
Sometimes is faster looking directly to the reference list rather using the search bar. However, the reference list is unordered, which causes some confusion when looking for a specific method:

![Screenshot_2020-02-18 Tools — conan 1 22 2 documentation](https://user-images.githubusercontent.com/4870173/74732646-764b3b00-5229-11ea-80ec-34bf0fcfb548.png)

PR #1565, opened 2020-02-18
#1564 Sorted Tools reference

uilianries:
Related issue: #1564 

I've made a [small script](https://gist.github.com/uilianries/5737f5c76b31b30bc624defe365b1039) which is able to sort the method block for tools reference. Unfortunately, I didn't find any plugin which is able to do it for readthedocs.

Good luck reviewing the entire reference/tools.rst  :sweat_smile: 

Also, I found some broken references there in tools reference. The build command didn't report because those links are not being used.

NOTE: I want to update other parts of the documentation, but it can result in conflicts due other PRs, so I'll do it by parts.

### DIFF
--- a/howtos/manage_cpp_standard.rst
+++ b/howtos/manage_cpp_standard.rst
@@ -100,4 +100,4 @@ Required version
 ----------------
 
 When the package to be built requires a minimal C++ standard support (e.g. 17), it can be done by
-comparing the ``cppstd``. For such condition, there is the helper :ref:`check_min_cppstd <tools.check_min_cppstd>`.
+comparing the ``cppstd``. For such condition, there is the helper :ref:`check_min_cppstd <tools_check_min_cppstd>`.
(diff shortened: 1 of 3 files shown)


danimtb:
We have the glossary section sorted this way https://github.com/conan-io/docs/blob/develop/glossary.rst but I think the `:sorted:` thing only works under a parent section

uilianries:
Oh boy! This is much better! I'll use it instead. Thanks a lot!

danimtb:
let me know if that works, I am not sure but it worths a try

uilianries:
Actually not, reading the [glossary](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary) section on Sphinx, only there the directive **:sorted:** works. I spent some time trying to find a similar way, but the closest approach was using the python config script to sort all indexes. Also, I opened a question on [stackoverflow](https://stackoverflow.com/questions/60320880/how-to-sort-items-on-sphinx-rst), let's see.

uilianries:
According the answer there in StackOverflow, there is no tool for sorting. Thus, I can say this an impossible mission. I'm closing this issue.

PR #1567, opened 2020-02-19
Update editable packages using an example

uilianries:
Required PR first: https://github.com/conan-io/examples/pull/47


developing_packages/editable_packages.rst
```diff
@@ -179,9 +187,9 @@ Layout files are specified in the :command:`conan editable add` command, as an e
 
 .. code-block:: bash
 
-    $ conan editable add . cool/version@user/dev --layout=win_layout
+    $ conan editable add . say/0.1@user/channel --layout=layout_vs
```
jgsogo:
```suggestion
    $ cd examples/features/editable/cmake/say
    $ conan editable add . say/0.1@user/channel --layout=layout_vs
```

developing_packages/editable_packages.rst
```diff
@@ -24,32 +24,40 @@ to copy locally built artifacts into the conan cache and make them available to
...
 
-Let's see this feature over an example where a developer is creating a ``CoolApp`` but at the same
-time they want to work on ``cool/version@user/dev`` library which is tightly coupled to the app.
+Let's see this feature over an practical example; the code can be found in the conan examples repository:
 
-The package ``cool/version@user/dev`` is already working, the developer has the sources in a
+.. code-block:: bash
+
+    $ git clone https://github.com/conan-io/examples.git
+    $ cd features/editable/cmake
```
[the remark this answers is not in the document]
uilianries: Yeah, but I followed the same idea from other examples in docs. I don't think putting `cd` in the description is problem, because it works for any OS.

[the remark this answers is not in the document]
jgsogo: No, but if we assume that the user is already inside the directory `features/editable/cmake`, then all the other commands in this piece of docs wouldn't be so clear (IMO): instead of

```
conan editable add examples/features/editable/cmake/say say/0.1@user/channel
```

it should be

```
conan editable add say say/0.1@user/channel
```

That's my only point to avoid that `cd`

developing_packages/editable_packages.rst
```diff
@@ -153,14 +161,19 @@ This file can use the package reference to customize logic for a specific packag
...
+       [source_folder]
+       src
+
        [includedirs]
-       src/include
+       src
 
-       [cool/version@user/dev:includedirs]
-       src/core/include
+       [libdirs]
```
jgsogo: We need to show in the docs the syntax with the `[reference:one-directory]`. What I meant by https://github.com/conan-io/examples/pull/47#discussion_r387561269 is that we should change the file in the example, not here in the docs. This syntax is quite confusing and having the example running with it will be useful

developing_packages/editable_packages.rst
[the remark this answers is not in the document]
uilianries: I've removed cd from the docs

developing_packages/editable_packages.rst
uilianries: Done, I've added to the jinja layout file

jgsogo: But, before entering this snippet, the text says _"This file can use the package reference to customize logic for a specific package:"_, so it has to be here, or move the comment above.

jgsogo: But, before entering this snippet, the text says "This file can use the package reference to customize logic for a specific package:", so it has to be here, or move the comment above.

uilianries: Indeed! Okay, it's updated now.

PR #1575, opened 2020-03-03
docs for full_transitive_package_id

memsharded:
Docs for https://github.com/conan-io/conan/pull/6451


creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
             self.compatible_packages.append(p)
 
 
+.. _full_transitive_package_id:
+
+Enabling full transitivity in package_id modes
+++++++++++++++++++++++++++++++++++++++++++++++
+
+When a package declares in its ``package_id()`` method that it is not affected by its dependencies, that will propagate down
+to the indirect consumers of that header only package. If the header only package dependency is a static library, for example, then downstream
```
jgsogo: As this affects only to a graph with packages declared as `header_only`, we should start this section with a big warning to make clear that difference, the intention and the rationale: it works this way for all the packages but for header-only ones. From that starting point we can start writing

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+consumers of the header only library that uses a package mode different from the default, should be also affected by the upstream
+transitivity dependency. Lets say that we have the following scenario:
+
+- ``pkgA/1.0``
+- ``pkgB/1.0`` depends on ``pkgA/1.0``, and defines ``self.info.header_only()`` in its ``package_id()``
+- ``pkgC/1.0`` depends only on ``pkgB/1.0``
+- ``App/1.0`` depends on ``pkgC/1.0`` and ``pkgA/1.0``
+- We are using ``full_version_mode``
+- Now we create a new ``pkgA/2.0`` that has some changes in its header, that would require to rebuild ``pkgC/1.0`` against it.
+- ``App/1.0`` now depends on ``pkgC/1.0`` and ``pkgA/2.0``
```
jgsogo: We need a graph here, and we need to see the `conanfile.py` sources. This way it is really hard to understand for someone reading about package IDs

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+
+    $ conan config set general.full_transitive_package_id=1
+
+Or change:
+
+.. code-block:: text
+   :caption: *conan.conf* configuration file
+
+   [general]
+   full_transitive_package_id=1
```
jgsogo: Activating and deactivating should be clear enough in the section about Conan configuration, here we can refer to that section, but we shouldn't write it here.

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+
+
+If we do this, then ``pkgC/1.0`` will compute 2 different package-id, one for ``pkgA/1.0`` and the other to link with ``pkgA/2.0``
+
+This will become the default behavior in the future (Conan 2.0)
```
jgsogo: This warning/notice should be at the top of the section, and probably we should recommend activating it as it will be the default behavior and it fixes what is considered a wrong behavior.

reference/config_files/conan.conf.rst
```diff
@@ -134,6 +135,10 @@ as it may result in corrupted packages.
 The ``default_package_id_mode`` changes the way package IDs are computed. By default, if not specified
 it will be ``semver_direct_mode``, but can change to any value defined in :ref:`package_id_mode`.
 
+The ``full_transitive_package_id`` changes the way package IDs are computed regarding transitive dependencies.
```
jgsogo: This changes the way if, and only if, there are header-only packages in the graph. I'd add that note.

jgsogo:
I'm asking @czoido to review it too. It is needed that someone that hasn't been involved in the PRs understands this piece of docs

jgsogo:
@memsharded , have a look at the review and merge `develop` to fix the broken CI

@czoido , please, review the docs, how a newcomer to Conan will read about this (with/without this flag activated)?

creating_packages/define_abi_compatibility.rst
memsharded: No, it can also affect users calling ``self.info.requires.clear()``, ``self.info.requires.remove["dep"]``, ``self.info.requires.unrelated_mode()``. It is not only for header-only libraries. I will add these cases to the docs.

memsharded:
Added a infographic and improved the text. I think it should be more understandable now.

reference/config_files/conan.conf.rst
memsharded: As said above, not only for header-only

creating_packages/define_abi_compatibility.rst
memsharded: Done.

creating_packages/define_abi_compatibility.rst
memsharded: Done

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+    This will become the default behavior in the future (Conan 2.0). It is recommended to activate it when possible (it might require rebuilding some packages,
+    as their package IDs will change)
+
+
+When a package declares in its ``package_id()`` method that it is not affected by its dependencies, that will propagate down
+to the indirect consumers of that package. There are several ways this can be done, ``self.info.header_only()``, ``self.info.requires.clear()``,
+``self.info.requires.remove["dep"]`` and ``self.info.requires.unrelated_mode()``, for example.
+
+Let's assume for the discussion that it is a header only library, using the ``self.info.header_only()`` helper. This header only package has
+a single dependency, which is a static library. hen downstream
```
danimtb:
```suggestion
a single dependency, which is a static library. Then, downstream
```

czoido:
The image preview looks like this for me... 🤔Is it just my screen? (Macbook retina)

image

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,49 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+- ``PkgC/1.0`` depends only on ``PkgB/1.0``
+- ``PkgB/1.0`` depends on ``PkgA/1.0``, and defines ``self.info.header_only()`` in its ``package_id()``
+- We are using ``full_version_mode``
+- Now we create a new ``PkgA/2.0`` that has some changes in its header, that would require to rebuild ``PkgC/1.0`` against it.
+- ``App/1.0`` now depends on ``PkgC/1.0`` and ``PkgA/2.0``
+
+.. image:: /images/conan-full_transitive_package_id.png
+    :height: 310 px
+    :width: 200 px
+    :align: center
```
czoido:
```suggestion
    :width: 100 %
    :align: center
```

creating_packages/define_abi_compatibility.rst
```diff
@@ -701,6 +701,48 @@ generated with the ``recipe_revision_mode`` can be resolved if no package for th
...
+
+Let's assume for the discussion that it is a header only library, using the ``self.info.header_only()`` helper. This header only package has
+a single dependency, which is a static library. Then, downstream
+consumers of the header only library that uses a package mode different from the default, should be also affected by the upstream
+transitivity dependency. Lets say that we have the following scenario:
+
+- ``App/1.0`` depends on ``PkgC/1.0`` and ``PkgA/1.0``
+- ``PkgC/1.0`` depends only on ``PkgB/1.0``
+- ``PkgB/1.0`` depends on ``PkgA/1.0``, and defines ``self.info.header_only()`` in its ``package_id()``
+- We are using ``full_version_mode``
```
czoido: I would move this to the line 722: "Lets say that we have the following scenario where we are using full_version_mode:"

czoido:
In my humble opinion, if the point of the section is to explain the propagation of the packageID and what `full_transitive_package_id` does I would put an example where App does not depend on PkgA directly to explain the`full_transitive_package_id` and then, after this, putting explaining what could happen if it depended from PkgA directly as well. I felt that a little bit confusing at first.

PR #1580, opened 2020-03-05
Warning for conan-docker-tools images

danimtb:
We are using and maintaining the images in Conan Docker Tools for open source usage and we'd like to clearly state that in the documentation and warn people to not use them in production or do it at their own risk.

Please provide feedback about the text and I will open a PR later to the readme on the conan-docker-tools repo.


jgsogo:
@danimtb , check if you can add an empty commit or something (close this and open another PR) to get the notification from the CLA.

PR #1593, opened 2020-03-12
Reword how revisions are fetched.

### LINKED ISSUE #1592: Ambiguous paragraph in how conan client handles revisions
Docs @ https://docs.conan.io/en/latest/versioning/revisions.html mention following:

```If a client requests a reference like lib/1.0@conan/stable, Conan will automatically retrieve the latest revision. In the client cache there is only one revision installed simultaneously.```

This could be interpreted as  it will *always* check for newer revisions and downloads them but that's not the case if one already has `lib/1.0@conan/stable` in local cache.

### PR DESCRIPTION
**Pull request by rasjani**
Fixes #1592

### CODE REVIEW

File: versioning/revisions.rst [hunk 1]
```diff
@@ -36,8 +36,10 @@ How it works
   The same package ID (for example for Linux/GCC5/Debug), can have multiple revisions (PREVs) that belong
   to a concrete RREV.
 
-If a client requests a reference like ``lib/1.0@conan/stable``, Conan will automatically retrieve the latest revision.
-In the client cache there is **only one revision installed simultaneously**.
+If a client requests a reference like ``lib/1.0@conan/stable``, Conan will automatically retrieve the latest revision in case
+local cache doesn't have any revisions at that point. If client needs to update the existing revision, it has to ask for updates
+with with with -u, --update argument to :command:`conan install` command… In the client cache there is
```
Minimonium: `with with with` :slightly_smiling_face:

Minimonium:
```suggestion
with -u, --update argument to :command:`conan install` command. In the client cache there is
```

versioning/revisions.rst
```diff
@@ -36,8 +36,10 @@ How it works
   The same package ID (for example for Linux/GCC5/Debug), can have multiple revisions (PREVs) that belong
   to a concrete RREV.
 
-If a client requests a reference like ``lib/1.0@conan/stable``, Conan will automatically retrieve the latest revision.
-In the client cache there is **only one revision installed simultaneously**.
+If a client requests a reference like ``lib/1.0@conan/stable``, Conan will automatically retrieve the latest revision in case
+local cache doesn't have any revisions at that point. If client needs to update the existing revision, it has to ask for updates
```
Minimonium:
```suggestion
the local cache doesn't contain any revisions already. If a client needs to update an existing revision, they have to ask for updates explicitly
```

memsharded:
Targeting "master" branch, as this can be directly published, no need to wait for the next release.

PR #1596, opened 2020-03-15
redesign

nadavc-jfrog:
MAR-681


memsharded:
Can you please capture some screen shots of the output? Here it seems the TOC is in read, it is non-readable.

You also need to sign the CLA properly, please check.

PR #1604, opened 2020-03-17
intro

memsharded:


memsharded:
The CI is failing @jgsogo Should we lower the checker limits even further? What is happening?

jgsogo:
Yes, let's lower the number of clients. Some time ago (2 weeks) we increased the number of threads checking those links, from 5 (default) to 30, it was working, but something has changed (GitHub?) and now it is failing _always_. The downside is that this check will take longer... and it takes some time.

I'm thinking about a different approach: skip all the checks for all the `https://github.com/conan-io/conan/pull/XXXX` links, they won't fail and if they do we will probably realize about it without the need of this automatic linkcheck. wdyt?

---

Meanwhile, I'll take care of this PR and lowering the workers.

memsharded:
> I'm thinking about a different approach: skip all the checks for all the https://github.com/conan-io/conan/pull/XXXX links, they won't fail and if they do we will probably realize about it without the need of this automatic linkcheck. wdyt?

Yes, I am totally fine with skipping those specific checks.

But I would go even further. Have an extremely lightweight PR checks. For me, it is more annoying that bringing very high value. The number of true positives of these checkers has been extremely low, if any, but the false positives are too high. Not publishing the docs from master branch because of a random failure in the CI is problematic, especially in master. I'd rather have a nightly build for us to check things.

jgsogo:
(I've moved your comment to https://github.com/conan-io/docs/issues/1607#issuecomment-600536453)

PR #1605, opened 2020-03-17
Use the build folder to apply any patch

uilianries:
Related to https://github.com/conan-io/conan-center-index/pull/1107#discussion_r392179080


reference/tools.rst
```diff
@@ -591,7 +591,7 @@ tools.patch()
     def patch(base_path=None, patch_file=None, patch_string=None, strip=0, output=None, fuzz=False)
 
 Applies a patch from a file or from a string into the given path. The patch should be in diff (unified diff) format. To be used mainly in
-the ``source()`` method.
+the ``build()`` method.
```
jgsogo: I'd try to add a reason here, very short, but something:

_"Use this tool in the ``source()`` method only for those patches that don't depend on settings, otherwise move usage to ``build()`` method"_

uilianries: I would say, only when is not copied to build_folder, `no_copy_sorce=True`. Otherwise, use build_folder.

memsharded: Still no, even if you use ``no_copy_source=True``, you don't want to try to apply patches that depends on settings or options.

uilianries: For recipe development, it's best applying to `build()` only, even when it doesn't depends on settings.
- When you need to check the authors source code to review, the source folder is the clear, you can check before any patch there.
- We need to download the sources only once, if for some reason the patch is broken/wrong, you will need to download all sources again until fixing it, and believe, it happens.
- build folder is re-created for every new build, applying patch there is not problem  because it's "temporary"
- IHMO source folder should be untouched. we can use for review as the single source of truth

The only case that I can't use the build folder, is for `no_copy_source=True`. Because the package folder needs to collect from source folder only. Header-only and Boost are good examples.

jgsogo: Ok, then prefer the `build()` over the `source()`: _"Use it preferably in the `build()` method"_ 👍 

The `source()` method won't have `settings` and `options` available ([docs about v2](https://docs.conan.io/en/latest/reference/conan_v2_mode.html#changes-in-recipes)), so those recipes with `no_copy_sorce=True` won't be able to use it in order to apply conditionally some patches.

jgsogo:
Please, @uilianries , merge changes from `master`, they should fix the CI and all those `429 Client Error` errors.

reference/tools.rst
uilianries: Done, I just added your new statement. It sounds good.

uilianries:
@jgsogo done, I've updated with master branch.

uilianries:
@jgsogo Interesting, even merging from master, we have a lot of broken links ... I gonna open a new PR with a fix.

uilianries:
@jgsogo only 2 links has been redirected.

However, Gnome is down:
https://www.isitdownrightnow.com/gnome.org.html

GCC org is up, so I think it was a temporary error:
https://www.isitdownrightnow.com/gcc.gnu.org.html

reference/tools.rst
```diff
@@ -590,8 +590,7 @@ tools.patch()
 
     def patch(base_path=None, patch_file=None, patch_string=None, strip=0, output=None, fuzz=False)
 
-Applies a patch from a file or from a string into the given path. The patch should be in diff (unified diff) format. To be used mainly in
-the ``source()`` method.
+Applies a patch from a file or from a string into the given path. The patch should be in diff (unified diff) format. Use it preferably in the build() method.
```
jgsogo:
```suggestion
Applies a patch from a file or from a string into the given path. The patch should be in diff (unified diff) format. Use it preferably in the ``build()`` method.
```

danimtb:
Should we add the recommendations in https://github.com/conan-io/conan-center-index/pull/1107#discussion_r392179080 to complete the explanation or do you think it is a bit excessive?

uilianries:
@danimtb It's a good idea. At least, users will have a reference when looking for an official reference.

uilianries:
Gnome is up again, let's restart the CI.

uilianries:
Well ...

> (line  128) broken    https://msbuildlog.com - 403 Client Error: Site Disabled for url: https://msbuildlog.com/

danimtb:
Don't worry about the links. We are going to remove those checks in master so PRs affected can be merged

PR #1613, opened 2020-03-19
Add docs for CONAN_SYSREQUIRES_MODE default_mode

czoido:
Docs for https://github.com/conan-io/conan/pull/6677

### DIFF
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -623,6 +623,8 @@ On Windows, there is no standard package manager, however **choco** can be invok
             installer = SystemPackageTool(tool=ChocolateyTool()) # Invoke choco package manager to install the package
             installer.install(pack_name)
 
+.. _systempackagetool:
+
 SystemPackageTool
 +++++++++++++++++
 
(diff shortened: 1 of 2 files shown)


danimtb:
I'd suggest cross-referencing both sections for easier navigation

reference/env_vars.rst
```diff
@@ -461,6 +461,8 @@ the :ref:`tools.msvc_build_command() <tools_msvc_build_command>`
 and the :ref:`MSBuild() <msbuild>` build helper, will not call ``devenv`` command to upgrade the ``sln`` project, irrespective of
 the ``upgrade_project`` parameter value.
 
+.. _env_vars_conan_sysrequires_mode:
+
```
jgsogo: The default was changed in the PR, we need to update these docs.

reference/env_vars.rst
```diff
@@ -480,6 +482,11 @@ See values behavior:
       doesn't install any package in your system. Useful if you want to keep manual control of these dependencies,
       for example in your development environment.
 
+Although this value is defaulted to ``enabled``, the default value can be configured with the
+``default_mode`` parameter in the :ref:`SystemPackageTool<systempackagetool>` constructor. If this parameter is passed,
+``CONAN_SYSREQUIRES_MODE`` will take this value if the environment variable has not been previously
+defined.
```
jgsogo: I would remove this paragraph, the value of envvar `CONAN_SYSREQUIRES_MODE` won't take the value of the argument `default_mode`.

reference/conanfile/methods.rst
```diff
@@ -651,6 +653,16 @@ When the environment variable ``CONAN_SYSREQUIRES_SUDO`` is not defined, Conan w
...
 
+Also, when the environment variable :ref:`CONAN_SYSREQUIRES_MODE <env_vars_conan_sysrequires_mode>` is not
+defined, Conan will set its value to ``enabled``. In this case, it's also possible to set the default
+value of the variable to other valid values (``verify`` or ``disabled``). This can be useful in the
+case that, for example, we want to install missing system requirements in the docker images of a CI
+server so that they are able to generate binaries (this images would run with the environment
+variable ``CONAN_SYSREQUIRES_MODE`` defined as ``enabled``), but passing ``default_mode="verify"`` as
+a parameter in the constructor of ``SystemPackageTool`` in the recipe, we will not force the
+installation of those requirements when the recipe is installed and ``CONAN_SYSREQUIRES_SUDO`` is not
+set.
```
jgsogo: I don't understand this paragraph/example 😓 

If the docker images run with `CONAN_SYSREQUIRES_MODE=enabled` that value will override the `default_mode="verify"`...

czoido: Yes, that's the idea, if you set the `CONAN_SYSREQUIRES_MODE` environment variable, that value will always be used, it takes preference from the argument in the `SystemPackageTool` constructor.
That's how you can install the requirements in the Docker images but not force the installation if you don't have that variable set. I'll try to rephrase this a little bit as it may be confusing.

reference/env_vars.rst
czoido: The default will still be the same as before unless you define the `default_mode` parameter in the constructor of `SystemPackageTool` and you don't define the variable.

jgsogo: No, now the default is `None`. We don't know if someone is reading this environment variable to perform any other task 🍔

reference/env_vars.rst
```diff
@@ -461,10 +461,12 @@ the :ref:`tools.msvc_build_command() <tools_msvc_build_command>`
 and the :ref:`MSBuild() <msbuild>` build helper, will not call ``devenv`` command to upgrade the ``sln`` project, irrespective of
 the ``upgrade_project`` parameter value.
 
+.. _env_vars_conan_sysrequires_mode:
+
 CONAN_SYSREQUIRES_MODE
 ----------------------
 
-**Defaulted to**: ``enabled`` allowed values ``enabled``/``verify``/``disabled``
+**Defaulted to**: ``None`` allowed values ``enabled``/``verify``/``disabled``
```
jgsogo: Others use "Not defined":

```suggestion
**Defaulted to**: Not defined (allowed values ``enabled``/``verify``/``disabled``)
```

PR #1625, opened 2020-03-24
Remove redundant text in visualstudio.rst

syntroniks:


danimtb:
I tried to retarget this to master as it is not related to the next release of Conan but did not work. It is ok as is, it will be published on next Conan release

syntroniks:
@danimtb I will fix this PR
Thanks for your speedy review

PR #1586, opened 2020-03-10
Draft for cross-building

jgsogo:
Draft for crossbuilding: https://github.com/conan-io/conan/pull/5594

Use with this repo: https://github.com/jgsogo/conan-xbuild

PR #1629, opened 2020-03-25
Docs about 'profile:host' and 'profile:build'

jgsogo:
Docs for https://github.com/conan-io/conan/pull/5594

Closes https://github.com/conan-io/docs/pull/1586


devtools/create_installer_packages.rst
```diff
@@ -3,199 +3,190 @@
 Creating conan packages to install dev tools
 ============================================
 
-Conan 1.0 introduced two new settings, ``os_build`` and ``arch_build``. These settings represent the machine where Conan is running, and are
-important settings when we are packaging tools.
+One of the most useful features of Conan is to package executables like compilers or build tools and
+distribute them in a controlled way to the team of developers. This way Conan helps not only with the 
+graph of dependencies of the application itself, but also with all the ecosystem needed to generate the
+project, making it really easy to control everythin involved in the deployed application.
```
czoido:
```suggestion
project, making it really easy to control everything involved in the deployed application.
```

devtools/create_installer_packages.rst
```diff
@@ -3,199 +3,190 @@
...
+that are in the ``build`` context. In other words: in this example we are running a Windows machine, so 
+and we need a version of ``nasm`` compatible with this machine, so we are providing a ``windows`` profile
+for the ``build`` context, and we are generating the library for the ``host`` platform which is declared
+in the ``profile_host`` profile (read more about :ref:`build requires context <build_requires_context>`).
 
-            def build(self):
-                cmake = CMake(self)
-                ...
+Using two profiles forces Conan to make this distinction and it has several advantages over the previous
+approach:
```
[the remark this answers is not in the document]
jgsogo: I'm modifying the line, but I want to keep the reference to the previous paragraph.

devtools/create_installer_packages.rst
```diff
@@ -3,199 +3,195 @@
...
 
        def package(self):
-           self.copy("*", dst="", keep_path=True)
+           self.copy("*", src=self.nasm_folder_name, dst="bin", keep_path=True)
            self.copy("license*", dst="", src=self.nasm_folder_name, keep_path=False, ignore_case=True)
 
        def package_info(self):
-           self.output.info("Using %s version" % self.nasm_folder_name)
-           self.env_info.path.append(os.path.join(self.package_folder, self.nasm_folder_name))
+           self.env_info.PATH.append(os.path.join(self.package_folder, "bin"))
```
ytimenkov, line 69: @jgsogo it should be `cpp_info.bindir`, no?

ytimenkov: I mean Iirc all helpers should use `run_environment` or virtualrunenv, so having `bindirs` should be enough. Did this change?

jgsogo: Depending on where you use this recipe. If you use it with the new model with the two profiles, you just need to set `cpp_info.bindirs` to the `bin` directory (it is added by default). But if you want to use this same recipe using only one profile in the command line, you need to add the  `bin` folder to the PATH.

My intention here was to write an example that works with the two scenarios, even if this is talking about the two profiles... do you think it is misleading? I can imagine people copy/pasting an example and IMO it is better if it works for the two scenarios (if it requires almost no extra code)

ytimenkov: I mean Conan handled well ld_library_path on Linux, path on windows (for dynamic libraries) as well as macos wieredness with environment jamming in some cases.

And it felt that use of environment variables for path/dll path was discouraged or obsolete, things just worked (with `run_environment` though).

jgsogo: 🤔 You are right, `RunEnvironment` is already adding bin directories to `PATH`, so it works with one profile (this is what `run_envrionment=True` does) and with two profiles (logic that propagates environment from _build_ to _host_ does it already).

So this line is useless.

jgsogo: Needed for the `virtualenv` generator (using one profile)

ytimenkov: Yes, that's why I merged them into one (in our company): as soon as you want to run a tool you need runenv. And this is 109% times since we have compiler and python as Conan packages.

ytimenkov: I think`run_environment=True` could be a default for Conan 2.0. and virtualenv should behave as if running from recipe (thus all virtual* merged).
 
I don't see why it is a harmful default. And we get defaults wrong: explicit, const, noexcept? 😂

jgsogo: Yes, Conan 2.0 will reconsiders the defaults for the `self.run(...)` and the environment that is populated before entering recipe functions like `build()`. We are also aware that all `virtualxxxenv` generators might be merged into one.

It's been many years of 1.x and our commitment to stability, we really are looking forward for the next major version and break some of the conventions.

PR #1633, opened 2020-03-26
Remove linkcheck from master jobs

danimtb:
Related to failing jobs at #1507 and #1605


uilianries:
Can they be changed to warning level instead? At least we will be able to see when building locally.

danimtb:
@uilianries I rarely or never use the deploy_gh_pages.py script, nor the make linkcheck locally. Only ``make html``.

uilianries:
Understood!

Issue #1545, opened 2020-02-04
Remove docs about obsolete 'config_file's (registry.txt)


jgsogo:
This section is obsolete. The file is now remotes.json, and it doesn't contain references or package_references anymore.

Leave these changes here, and do a PR to fix this. Thanks!

_Originally posted by @memsharded in https://github.com/conan-io/docs/pull/1537_

jgsogo:
Mentioned here too: https://github.com/conan-io/conan/pull/4724#issuecomment-481150908

jgsogo:
Fixed in https://github.com/conan-io/docs/pull/1640

PR #1640, opened 2020-03-31
Updates docs related to 'remotes.json' file

jgsogo:
Closes https://github.com/conan-io/docs/issues/1545

Removes `config_files/registry.txt` entry in `reference/config_files` and remove one mention to it inside the text.


reference/config_files.rst
```diff
@@ -9,12 +9,13 @@ These are the most important configuration files, used to customize conan.
...
    config_files/default_profile
-   config_files/settings.yml
-   config_files/registry.txt
-   config_files/client_certificates
-   config_files/artifacts.properties
    config_files/editable_layout
-   config_files/conandata.yml
+   config_files/metadata.json
+   config_files/remotes.json
+   config_files/settings.yml
```
jgsogo: Changing the order to follow some order, now they use lexicographic ordering.

reference/config_files/metadata.json.rst
```diff
@@ -0,0 +1,69 @@
+.. _metadata.json:
+
+metadata.json
```
memsharded: I don't think we should document this file at all.
This is internal implementation detail, subject to change at anytime

jgsogo: I agree, it is the same for `registry.json` file. Are you ok removing both of them? I'd prefer to remove them, for sure.

memsharded: Yes, I am totally fine with that, I didn't realize, that shouldn't be documented either.

Maybe just the opposite: a note somewhere saying that these files are internal implementation details, subject to change at anytime, and "dont use them", and point to the remotes.txt in ``conan config install`` for the remote case.

reference/config_files.rst
jgsogo: I've removed `config_files/registry.txt` entry

PR #1645, opened 2020-03-31
Set Travis Timeout to 20 min

uilianries:
The last deploy failed due Timeout. It's possible to update the output timeout:

https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

/cc @czoido @danimtb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ cache:
 install:
   - pip install -r requirements.txt
 script:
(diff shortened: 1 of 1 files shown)


jgsogo:
I'm running the `make latexpdf` locally and it stops at some point _forever_, so, unfortunately, it is not a matter of waiting more or less time. 😞

PR #1655, opened 2020-04-06
Improvement/cmake helper

aleksa2808:
* sorting the order of listed definitions in the CMake helper
* added information on CMAKE_MODULE_PATH / CMAKE_PREFIX_PATH updating (banged my head on that one untill I peeked into the code!)

The sorting splits some things that could perhaps go together, but it all ended up being quite messy with CMAKE_\*, CONAN_\* and plain variables being all mixed up.


danimtb:
Thanks @aleksa2808 for this PR. Indeed there was some missing information

PR #1656, opened 2020-04-06
Add command for c++ new abi

uilianries:
It's a common mistake the incompatible C++ ABI when you are newbie in Conan. The important information about that is separated by "Getting Started - Step 5" and "How to manage the GCC >= 5 ABI" sections.

This PR brings the magic command which switches to libstdc++11 for the default profile in the section "How to manage the GCC >= 5 ABI" to make easier to refer when explaining about that problem. Otherwise, we need to send getting started link and point the step, like in conan-io/conan#6808

### DIFF
--- a/howtos/manage_gcc_abi.rst
+++ b/howtos/manage_gcc_abi.rst
@@ -17,7 +17,12 @@ When Conan creates the default profile the first time it runs, it adjusts the ``
 compatibility. However, if you are using GCC >= 5 your compiler is likely to be using the new CXX11 ABI by default (libstdc++11).
 
 If you want Conan to use the new ABI, edit the default profile at ``~/.conan/profiles/default`` adjusting ``compiler.libcxx=libstdc++11``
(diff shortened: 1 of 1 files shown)


memsharded:
I  like to point to the Getting Started when people are facing this problem. This is not something that is hidden in a howto, but in the very first Getting Started doc page, highlighted in green as an important note.

uilianries:
So is it better distributing 2 links instead of only one?

memsharded:
> So is it better distributing 2 links instead of only one?

For this case, one link that points to Getting Started is enough. The information is there, clearly visible, but it is not being read. It is not because it is hidden in a howto, and this should be said upfront. We can link the how-to section from the Getting Started.

uilianries:
> We can link the how-to section from the Getting Started.

The link is already there. Well, let's keep it as is. The real problem is not the docs, but who is not reading the docs.

PR #1669, opened 2020-04-18
Fix example flags for new options/settings model

solvingj:


jgsogo:
I've created a PR to Conan fixing this typo.  We need to update here the output for all the commands affected https://github.com/conan-io/conan/pull/6937

danimtb:
@solvingj there is a script that updates the commands help pages here https://github.com/conan-io/docs/blob/develop/commands_help_update.py (the code is horrible but should do the job). You will need to have Conan installed from sources in your environment with the changes you want to update the documentation with.

jgsogo:
> @solvingj there is a script that updates the commands help pages here https://github.com/conan-io/docs/blob/develop/commands_help_update.py (the code is horrible but should do the job). You will need to have Conan installed from sources in your environment with the changes you want to update the documentation with.

Why was I not aware of that script? I've modified those lines several times by hand... 😭

solvingj:
Will use the script to re-submit a PR with more complete update/fix of all such inaccuracies after 1.25 release.

solvingj:
@danimtb can we create a PR template which mentions this script so that users see it any time a user creates a PR ?

danimtb:
yes, I think we can add that to the this repo

PR #1682, opened 2020-04-29
Add component requires information

danimtb:
Docs for conan-io/conan#6871

### DIFF
--- a/creating_packages/package_information.rst
+++ b/creating_packages/package_information.rst
@@ -26,6 +26,13 @@ like the location of the header files, library names, defines, flags...
 The package information is done using the attributes of the :ref:`cpp_info_attributes_reference` object. This information will be aggregated
 by Conan and exposed via ``self.deps_cpp_info`` to consumers and generators.
 
+.. important::
+
+    This information is important as it describes the package contents in a generic way with a pretty straightforward syntax that can later
+    be translated to a suitable format. The advantage of having this information here, is that the package could be consumed from a
+    different build system that the one used to compile the library. For example, a library that builds using Autotools can be consumed
+    later in CMake with this information using any of the CMake generators.
+
 .. _package_information_components:
 
 Using Components
@@ -60,6 +67,32 @@ executable will be one component inside ``cpp_info`` like this:
 
 You can define dependencies among different components using the ``requires`` attribute and the name of the component. The dependency graph
 for components will be calculated and values will be aggregated in the correct order for each field.
+
+. code-block:: python
+
+    def package_info(self):
+        self.cpp_info.components["LibA"].libs = ["liba"]      # Name of the library for the 'LibA' component
+        self.cpp_info.components["LibA"].requires = ["LibB"]  # Requires point to the name of the component
+
+        self.cpp_info.components["LibB"].libs = ["libb"]
+
+        self.cpp_info.components["LibC"].libs = ["libc"]
+        self.cpp_info.components["LibC"].requires = ["LibA"]
+
+        self.cpp_info.components["LibD"].libs = ["libD"]
+        self.cpp_info.components["LibD"].requires = ["LibA"]
+
+        self.cpp_info.components["LibE"].libs = ["libe"]
+        self.cpp_info.components["LibE"].requires = ["LibB"]
+
+        self.cpp_info.components["LibF"].libs = ["libf"]
+        self.cpp_info.components["LibF"].requires = ["LibD", "LibE"]
+
+For consumers and generators, the order of the libraries from this components graph will be:
+
+.. code-block:: python
+
+        self.deps_cpp_info.libs == ["libf", "libe", "libd", "libc", "liba", "libb"]
 
 Declaration of requires from other packages is also allowed:
 
--- a/creating_packages/package_information.rst
+++ b/creating_packages/package_information.rst
@@ -33,6 +33,10 @@ by Conan and exposed via ``self.deps_cpp_info`` to consumers and generators.
     different build system that the one used to compile the library. For example, a library that builds using Autotools can be consumed
     later in CMake with this information using any of the CMake generators.
 
+.. seealso::
+
+    Read :ref:`method_package_info` to learn more about this method.
+
 .. _package_information_components:
 
 Using Components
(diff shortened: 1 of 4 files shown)


reference/conanfile/attributes.rst
```diff
@@ -965,7 +998,23 @@ You can also model system dependencies for each component or just header files.
...
+    class MyConan(ConanFile):
+        ...
+        requires = "zlib/1.2.11", "openssl/1.1.1g"
+
+    def package_info(self):
+        self.cpp_info.components["comp1"].requires("zlib::zlib")
+        self.cpp_info.components["comp2"].requires("comp1", "openssl::ssl")
+
+Note that component requires should **explicitly** define requirements to other package (``zlib::zlib``) or other package components
+(``openssl::ssl``).
```
jgsogo, line 1017: I would convert this into a "note", "important", or similar, so it is highlighted in the documentation. Even though Conan will raise an error, for me it is very important, it the only difference between the top-level `cpp_info` and the components' one

creating_packages/package_information.rst
```diff
@@ -0,0 +1,81 @@
+.. _package_information:
+
+Define the package information
+==============================
+
+When creating a recipe to package a library, it is important to define the information about the package so consumers can get the
+information correctly. Conan achieves this by decoupling the information of the package from the format needed using
+:ref:`generators_reference`, that translate the generic information into the appropriate format file.
+
+This generic information is define inside the recipe, using the :ref:`method_package_info` method. There you can declare package information
```
madebr, line 10:
```suggestion
This generic information is defined inside the recipe, using the :ref:`method_package_info` method. There you can declare package information
```

reference/conanfile/attributes.rst
```diff
@@ -933,30 +933,58 @@ Components
 
     This is a **experimental** feature subject to breaking changes in future releases.
```
jgsogo, line 935: IMO, I'd remove all this "Components" section and substitute it by a paragraph (if you feel like, move more things to the other section):

_"More fine grained control over individual libraries available in a single Conan package can be achieved using `components` (LINK). Components allow you define a `cpp_info` object per each of those libraries and also requirements between them and to components of other packages:"_

>>>> Python with example. 

_"Note that components should declare explicitly that they require other packages...."_

creating_packages/package_information.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. code-block:: python
+
+    class MyConan(ConanFile):
+        ...
+        requires = "zlib/1.2.11", "openssl/1.1.1g"
+
+    def package_info(self):
+        self.cpp_info.components["comp1"].requires("zlib::zlib")             # Depends on all components in zlib package
+        self.cpp_info.components["comp2"].requires("comp1", "openssl::ssl")  # Depends on ssl component in openssl package
```
jgsogo, line 72: Does the syntax `requires("....")` work?  I can't remember any example/test using them in the PR.

creating_packages/package_information.rst
```diff
@@ -0,0 +1,81 @@
...
+
+.. important::
+
+    Note that component requires should **explicitly** define requirements to other package (``zlib::zlib``) or other package components
+    (``openssl::ssl``).
```
jgsogo, line 77: I think this is important and it is not clear enough: _"By default components won't link against any other package required by the recipe, the `requires` list has to be populated explicitly with the list of components from other packages to use: it can be a single component (example) or the full requirement (example)."_

creating_packages/package_information.rst
```diff
@@ -0,0 +1,81 @@
...
+
```
jgsogo, line 78: We probably need here other note about built-in generators, they don't allow yet to consume this information.

creating_packages/package_information.rst
danimtb: you are right, this is an error

reference/conanfile/attributes.rst
memsharded: Kind of agree. This section still reads a bit like tutorial, not reference. Maybe moving some parts to the other doc, and leaving here just the components fields reference.

creating_packages/package_information.rst
```diff
@@ -0,0 +1,89 @@
...
+    def package_info(self):
+        self.cpp_info.name = "OpenSSL"
+        self.cpp_info.components["crypto"].names["cmake_find_package"] = "Crypto"
+        self.cpp_info.components["crypto"].libs = ["libcrypto"]
+        self.cpp_info.components["crypto"].defines = ["DEFINE_CRYPTO=1"]
+        self.cpp_info.components["ssl"].names["cmake"] = "SSL"
+        self.cpp_info.components["ssl"].includedirs = ["include/headers_ssl"]
+        self.cpp_info.components["ssl"].libs = ["libssl"]
+        self.cpp_info.components["ssl"].requires = ["crypto"]
+
```
memsharded, line 60: I think at this moment it would be very good to highlight as "important" note, that yes, this is a bit of repeated information of what the package might have in its build scripts, but this is better:
- Not a big deal, typically define it once, not that terrible repetition, and it is nicely explicit and relatively intuitive
- The most important thing: This is generic and can be consumed by other build systems, different ones from the used one.

danimtb: I added the "important" box at the top as a general message. Please check it and I can move it down here if you want. thanks

reference/conanfile/attributes.rst
```diff
@@ -924,52 +924,49 @@ This object should be filled in ``package_info()`` method.
...
 
     Components information is still not available from the consumer side (``self.deps_cpp_info`` doesn't provide the ``components``
-    dictionary). We are planning to complete this feature in next releases.
-
-    The information of components is not lost but aggregated to the *global* scope and the usage of components should be transparent right
-    now to consumers and generators.
-
-The structure of the ``Component`` object is the same as the one used by the ``cpp_info`` object and has **the same default directories**.
+    dictionary). The information of components is not lost but aggregated to the *global* scope and the usage of components should be
+    transparent right now to consumers and generators.
```
jgsogo, line 965: I'd remove this `important` and add a link to the `package_information_components`. Take into account the `seealso` note below, probable both of them can be joined.

PR #1695, opened 2020-05-12
Master

memsharded:


reference/tools.rst
```diff
@@ -67,7 +67,7 @@ If **arch** or **compiler_version** is specified, it will ignore the settings an
 for these parameters.
 
 Parameters:
-    - **settings** (Required): Conanfile settings. Use ``self.settings``.
+    - **conanfile** (Required): Conanfile object. Use ``self`` in a ``conanfile.py``.
```
danimtb, line 70: Should this indicate that this is also backward compatible with a ``settings`` argument?

danimtb: Should this change be done in `vcvars_dict()` and `vcvars()` as well?

reference/tools.rst
```diff
@@ -810,19 +810,31 @@ tools.cross_building()
...
 
     from conans import tools
 
-    if tools.cross_building(self.settings):
+    if tools.cross_building(self):
         # Some special action
 
 Parameters:
-    - **settings** (Required): Conanfile settings. Use ``self.settings``.
+    - **conanfile** (Required): Conanfile object. Use ``self`` in a ``conanfile.py``.
```
danimtb, line 837: Is this backward compatible with a `settings` argument as well?

memsharded: No, we don't want them to be passing settings anymore. The argument in latest Conan is a conanfile. Yes, it might not break, but that is a different story.

reference/tools.rst
danimtb: I have opened an issue about this here https://github.com/conan-io/conan/issues/7009

Issue #1694, opened 2020-05-12
document python_requires_extend attribute


SSE4:
from https://github.com/conan-io/hooks/pull/124#issuecomment-627374534
it appears that attribute `python_requires_extend` is not documented on the [attributes](https://docs.conan.io/en/latest/reference/conanfile/attributes.html) reference section. meanwhile, it's mentioned at least twice in the docs:

- https://docs.conan.io/en/latest/reference/conanfile/methods.html?highlight=python_requires_extend
- https://docs.conan.io/en/latest/extending/python_requires.html?highlight=python_requires_extend

the documentation block suggested by @jgsogo :
```
python_requires_extend is the way to inherit from a base class inside a python requires. 
Conan injects this class as a base class to the conanfile in runtime. 
Why? We need to be able to load the python file with the recipe and 
the python requires information before resolving the classes/methods from 
python_requires (if we use Python inheritance we need to retrieve 
python_requires first and the interpreter will use the base class). This let Conan 
control the python_requires.
```

SSE4:
block has been added

PR #1709, opened 2020-05-22
Add 'python_requires_extend' to the reference. Minor changes to pyreqs docs

jgsogo:
close https://github.com/conan-io/docs/issues/1694


reference/conanfile/attributes.rst
```diff
@@ -1348,6 +1348,32 @@ Its basic syntax is:
...
+
+python_requires_extend
+----------------------
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+
+This class attribute defines one or more classes that will be injected in runtime as base classes of
```
SSE4, line 1360: it would be nice to have an example of syntax of multiple classes specified in `python_requires_extend`, e.g. should I specify tuple, array, list or something else?

reference/conanfile/attributes.rst
```diff
@@ -1348,6 +1348,32 @@ Its basic syntax is:
...
+the recipe class. Syntax for each of these classes should be a string like ``pyreq.MyConanfileBase``
```
SSE4, line 1361: also, it's not obvious what's the effect of injection.
e.g. will `class Pkg(ConanFile):` transform into `class Pkg(pyreq.MyConanfileBase):`? or will it transform into `class Pkg(pyreq.MyConanfileBase, ConanFile)` (multiple inheritance)? or something else?

jgsogo: I've added it to the `python_requires` section (the linked one)

reference/conanfile/attributes.rst
jgsogo: Now the example uses several python requires.

PR #1715, opened 2020-05-25
- docs for tools.msvs_toolset

SSE4:
docs for https://github.com/conan-io/conan/issues/7061

### DIFF
--- a/reference/tools.rst
+++ b/reference/tools.rst
@@ -1756,14 +1756,14 @@ tools.msvs_toolset()
 
 .. code-block:: python
 
-    def msvs_toolset(settings)
+    def msvs_toolset(conanfile)
 
-Returns the corresponding Visual Studio platform toolset based on the settings. For instance, it may return ``v142``
+Returns the corresponding Visual Studio platform toolset based on the settings of the given ``conanfile``. For instance, it may return ``v142``
 for ``compiler=Visual Studio`` with ``compiler.version=16``. If ``compiler.toolset`` was set in settings, it has a
 priority and always returned.
 
 Parameters:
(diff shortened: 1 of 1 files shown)


reference/tools.rst
```diff
@@ -1747,3 +1747,20 @@ for ``compiler.cppstd=17``, and so on.
 
 Parameters:
     - **settings** (Required): Conanfile settings. Use ``self.settings``.
+
+
+.. _tools.msvs_toolset:
+
+tools.msvs_toolset():
+---------------------
```
danimtb, line 1755:
```suggestion
tools.msvs_toolset()
--------------------
```

reference/tools.rst
```diff
@@ -1747,3 +1747,20 @@ for ``compiler.cppstd=17``, and so on.
...
+
+.. code-block:: python
+
+    def msvs_toolset(settings)
+
+Returns the corresponding Visual Studio platform toolset based on the settings. For instance, it may return ``v142``
+for ``compiler=Visual Studio`` with ``compiler.version=16``. If ``compiler.toolset`` was set in settings, it has a
+priority and always returned.
```
danimtb, line 1763: this means that the tool has no effect and it just returns the ``compiler.toolset`` value, right?

SSE4: correct, the only effect is returning `compiler.toolset` value in this case

jgsogo:
Oh! There are conflicting files

SSE4:
how come? I already solved conflicts yesterday. I've checked about 1 hour ago, there were no conflicts!

PR #1722, opened 2020-06-02
Documentation for components and cmake_find_package generator

danimtb:
Docs for https://github.com/conan-io/conan/pull/7108

### DIFF
--- a/reference/generators/cmake_find_package.rst
+++ b/reference/generators/cmake_find_package.rst
@@ -55,8 +55,56 @@ utilities exported by the package available for consumers just by setting `find_
 Moreover, this also adjusts `CMAKE_MODULE_PATH` and `CMAKE_PREFIX_PATH` to the values declared by the package in ``cpp_info.buildirs``, so
 modules in those directories can be found.
 
-Target in Find<PKG-NAME>.cmake
-------------------------------
+Components
+++++++++++
+
+If a recipe defines different components using ``self.cpp_info.components``, the variables declared in *Find<PKG-NAME>.cmake* will be:
+
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| NAME                               | VALUE                                                                                               |
++====================================+=====================================================================================================+
+| <PKG-NAME>_FOUND                   | Set to 1                                                                                            |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <PKG-NAME>_VERSION                 | Package version                                                                                     |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <PKG-NAME>_COMPONENTS              | List of components of the package                                                                   |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_INCLUDE_DIRS           | Containing all the include directories of the component                                             |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_INCLUDE_DIR            | Containing the first include directory declared by the component                                    |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_INCLUDES               | Same as ``<COMP-NAME>_INCLUDE_DIRS``                                                                |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_LIB_DIRS               | Containing all the library directories of the component                                             |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_RES_DIRS               | Containing all the resources directories of the component                                           |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_DEFINITIONS            | Definitions of the component                                                                        |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_COMPILE_DEFINITIONS    | Definitions of the component                                                                        |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


reference/generators/cmake_find_package.rst
```diff
@@ -55,8 +55,56 @@ utilities exported by the package available for consumers just by setting `find_
 Moreover, this also adjusts `CMAKE_MODULE_PATH` and `CMAKE_PREFIX_PATH` to the values declared by the package in ``cpp_info.buildirs``, so
 modules in those directories can be found.
 
-Target in Find<PKG-NAME>.cmake
-------------------------------
+Components
+++++++++++
```
jgsogo, line 59: I would expect previous variables (`<PKG-NAME>_INCLUDE_DIRS`, `<PKG-NAME>_SYSTEM_LIBS`,...) to be available too, so _"if using components these variables will be added to the ones listed before:"_:

reference/generators/cmake_find_package.rst
```diff
@@ -55,8 +55,56 @@ utilities exported by the package available for consumers just by setting `find_
...
++====================================+=====================================================================================================+
+| <PKG-NAME>_FOUND                   | Set to 1                                                                                            |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <PKG-NAME>_VERSION                 | Package version                                                                                     |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <PKG-NAME>_COMPONENTS              | List of components of the package                                                                   |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_INCLUDE_DIRS           | Containing all the include directories of the component                                             |
```
jgsogo, line 72: I'm not sure if these variables should have the package as prefix: `<PKG-NAME>_<COMP-NAME>_INCLUDE_DIRS`, otherwise there can be collisions between packages with components with the same name.

Maybe we should have a look to existing CMake's modules and see if (and how) they expose information about components.

reference/generators/cmake_find_package.rst
```diff
@@ -55,8 +55,56 @@ utilities exported by the package available for consumers just by setting `find_
...
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_FRAMEWORKS             | Framework names to do a `find_library()`                                                            |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_FRAMEWORK_DIRS         | Framework directories to perform the `find_library()` of <PKG-NAME>_FRAMEWORKS                      |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+| <COMP-NAME>_DEPENDENCIES           | List of requirements                                                                                |
++------------------------------------+-----------------------------------------------------------------------------------------------------+
+
+Being ``<COMP-NAME>`` the component name used as key in the components dictionary (by default) or the one declared in ``cpp_info.name`` or
+in ``cpp_info.names["cmake_find_package"]`` if specified
```
jgsogo, line 104: Here it should be  `self.cpp_info.components["cmp"].name` and `self.cpp_info.components["cmp"].names["cmake_find_package "]`. Probably too long but... 🤷

MGritsenko:
Trying new components feature, seems like I miss some detail.  
I've created a project which has two targets, core.dll and runner.exe. Core.dll depends on Qt5::Core. Runner.exe depends on core.dll.

Also I've created a conan recipe conanfile.py.

**I defined qt in requirements:**

```
def requirements(self):
        self.requires("qt/e5.12.4@mikhail/testing", private=False)
```
**In package_info I use components:**

```
def package_info(self):
    self.cpp_info.components["core"].libs = ["core"]
    self.cpp_info.components["core"].requires = ["qt"]
    self.cpp_info.components["runner"].requires = ["core"]
```

**On build I have error:**

> File "C:\Users\user\AppData\Roaming\Python\Python37\site-packages\conans\client\installer.py", line 571, in _call_package_info raise ConanException("%s package_info(): %s" % (str(conanfile), e)) conans.errors.ConanException: lexer/1.0.0@mikhail/testing package_info(): Package require 'qt' not used in components requires

I assume I miss some detail. I would appreciate any advice or reference to read about.

danimtb:
There is an example here https://docs.conan.io/en/latest/reference/conanfile/attributes.html#cpp-info

```
self.cpp_info.components["crypto"].requires = ["zlib::zlib"]  # Depends on all components in zlib package
```

In your case, if you want a component to depend on the whole `qt` package requirement you will have to specify:

```
self.cpp_info.components["core"].requires = ["qt::qt"]
```

Please try that and let me know if it works.  Thanks for trying this new feature!

MGritsenko:
> There is an example here https://docs.conan.io/en/latest/reference/conanfile/attributes.html#cpp-info
> 
> ```
> self.cpp_info.components["crypto"].requires = ["zlib::zlib"]  # Depends on all components in zlib package
> ```
> 
> In your case, if you want a component to depend on the whole `qt` package requirement you will have to specify:
> 
> ```
> self.cpp_info.components["core"].requires = ["qt::qt"]
> ```
> 
> Please try that and let me know if it works. Thanks for trying this new feature!

Thank you for your reply. Unfortunately it does not work. Continue to investigate..

MGritsenko:
> There is an example here https://docs.conan.io/en/latest/reference/conanfile/attributes.html#cpp-info
> 
> ```
> self.cpp_info.components["crypto"].requires = ["zlib::zlib"]  # Depends on all components in zlib package
> ```
> 
> In your case, if you want a component to depend on the whole `qt` package requirement you will have to specify:
> 
> ```
> self.cpp_info.components["core"].requires = ["qt::qt"]
> ```
> 
> Please try that and let me know if it works. Thanks for trying this new feature!

Thank you! It did help!

PR #1749, opened 2020-06-26
Fix exception in capture package version example

wdobbe:
The example to capture the package version from a CMakeLists.txt contains a wrong string suffix.
Executing the example will result in `TypeError: cannot use a bytes pattern on a string-like object`.
Replace with 'r' prefix.


danimtb:
Thank you!

PR #1752, opened 2020-06-29
Updated docs for conan new --template

TheQwertiest:
Docs for the following feature: https://github.com/conan-io/conan/pull/7215
Note: command line `--help` output was not updated, because it appears that it's heavily outdated for a lot of files anyway (i.e. `commands_help_update.py` changes a lot of files).

### DIFF
--- a/reference/commands/creator/new.rst
+++ b/reference/commands/creator/new.rst
@@ -109,7 +109,7 @@ Creates a new package recipe template with a 'conanfile.py' and optionally,
   - ``name`` and ``version``: defined from the command line.
   - ``package_name``: a *CamelCase* variant of the name. Any valid Conan package name like ``package_name``, ``package+name``, ``package.name`` or
     ``package-name`` will be converted into a suitable name for a Python class, ``PackageName``.
-  - ``conan_version``: a string containing current conan version, e.g. ``1.24.0``.
+  - ``conan_version``: an object that renders as the current Conan version, e.g. ``1.24.0``.
 
 - Create files from a custom, predefined user template directory:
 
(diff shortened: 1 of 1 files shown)


reference/commands/creator/new.rst
```diff
@@ -87,22 +87,51 @@ Creates a new package recipe template with a 'conanfile.py' and optionally,
...
-  Where ``name`` and ``version`` placeholders are defined from the command line, and ``package_name`` is a *CamelCase*
-  variant of the name: any valid Conan package name like ``package_name``, ``package+name``, ``package.name`` or
-  ``package-name`` will be converted into a suitable name for a Python class, ``PackageName``.
\ No newline at end of file
+  Supported place-holders:
+  
+  - ``name`` and ``version``: defined from the command line.
+  - ``package_name``: a *CamelCase* variant of the name. Any valid Conan package name like ``package_name``, ``package+name``, ``package.name`` or
+    ``package-name`` will be converted into a suitable name for a Python class, ``PackageName``.
+  - ``conan_version``: a string containing current conan version, e.g. ``1.24.0``.
```
jgsogo, line 112:
```suggestion
  - ``conan_version``: an object that renders as the current Conan version, e.g. ``1.24.0``.
```

In the future this probably will be an actual `Version` object, people will want to check for the version in order to add some features (or not) to the generated files.

```python
class Recipe(ConanFile):
   {% if conan_version >= "1.26" %}
   toolchain = "cmake"
   {% endif %}
   ...
```

TheQwertiest:
@jgsogo , but that is not consistent with the current implementation?

TheQwertiest:
I.e. documentation should reflect the current state and not the state of the future release. It will just confuse users otherwise.

jgsogo:
IMO, documentation should reflect **how you are allowed to use something**, in this case the important bit is that it renders as a version string (this is the only thing you should do with it right now), saying it is a string is a very strong commitment. 
If we say it is a string, someone could write:

```
{% set major = conan_version.split('.', 1)[0] %}
{% if major > 1 %}
   This is Conan v2!
{% endif %}
```

and that will block us from passing any other object type (it would be a regression). If we say it renders as a string, the previous snippet should use the `str`-cast and it won't break if we change the actual string for a `Version` object.

PR #1755, opened 2020-06-30
Add components reference to cmake_find_package_multi generator

danimtb:
Docs for https://github.com/conan-io/conan/pull/7259

### DIFF
--- a/reference/generators/cmake_find_package_multi.rst
+++ b/reference/generators/cmake_find_package_multi.rst
@@ -86,7 +86,7 @@ Components
 ++++++++++
 
 If a recipe uses :ref:`components <package_information_components>`, the targets generated will be ``<PKG-NAME>::<COMP-NAME>`` with the following properties adjusted. Being
-``<COMP-NAME>`` the dictionary key used to declare the component or the alternative name declared in ``cpp_info.components["comp_name"].name`` or in 
+``<COMP-NAME>`` the dictionary key used to declare the component or the alternative name declared in
 ``cpp_info.components["comp_name"].names["cmake_find_package_multi"]`` if specified:
 
 - ``INTERFACE_INCLUDE_DIRECTORIES``: Containing all the include directories of the component.
@@ -96,4 +96,4 @@ If a recipe uses :ref:`components <package_information_components>`, the targets
 
 Moreover, a global target ``<PKG-NAME>::<PKG-NAME>`` will be declared with the following properties adjusted:
 
-- ``INTERFACE_LINK_LIBRARIES``: Containing all the component targets to link the global target to (includes package's components only).
+- ``INTERFACE_LINK_LIBRARIES``: Containing the list of targets for all the components in the package.
--- a/reference/generators/cmake_find_package_multi.rst
+++ b/reference/generators/cmake_find_package_multi.rst
@@ -86,7 +86,7 @@ Components
 ++++++++++
 
 If a recipe uses :ref:`components <package_information_components>`, the targets generated will be ``<PKG-NAME>::<COMP-NAME>`` with the following properties adjusted. Being
-``<COMP-NAME>`` the dictionary key used to declare the component or the alternative name declared in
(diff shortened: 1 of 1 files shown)


reference/generators/cmake_find_package_multi.rst
```diff
@@ -46,6 +46,7 @@ A target named ``<PKG-NAME>::<PKG-NAME>`` target is generated with the following
 - ``INTERFACE_INCLUDE_DIRECTORIES``: Containing all the include directories of the package.
 - ``INTERFACE_LINK_LIBRARIES``: Library paths to link.
 - ``INTERFACE_COMPILE_DEFINITIONS``: Definitions of the library.
+- ``INTERFACE_COMPILE_OPTIONS``: Compile options of the library.
```
danimtb, line 49: this was missing from the documentation although it is taken as an example down below

reference/generators/cmake_find_package_multi.rst
```diff
@@ -81,3 +82,18 @@ the *CMakeLists.txt*.
 Moreover, this also adjusts `CMAKE_MODULE_PATH` and `CMAKE_PREFIX_PATH` to the values declared by the package in ``cpp_info.buildirs``, so
 modules in those directories can be found.
 
+Components
+++++++++++
+
+If a recipe uses :ref:`components <package_information_components>`, the targets generated will be ``<PKG-NAME>::<COMP-NAME>`` with the following properties adjusted. Being
+``<COMP-NAME>`` the package name used in the reference (by default) or the one declared in ``cpp_info.components["comp_name"].name`` or in
```
jgsogo, line 89:
```suggestion
``<COMP-NAME>`` the dictionary key used to declare the component or the alternative name declared in ``cpp_info.components["comp_name"].name`` or in 
```

memsharded, line 89: Do not document the ``cpp_info.components["comp_name"].name`` if we are planning to leave only the ``names`` one.

reference/generators/cmake_find_package_multi.rst
```diff
@@ -81,3 +82,18 @@ the *CMakeLists.txt*.
...
+``cpp_info.components["comp_name"].names["cmake_find_package_multi"]`` if specified:
+
+- ``INTERFACE_INCLUDE_DIRECTORIES``: Containing all the include directories of the component.
+- ``INTERFACE_LINK_LIBRARIES``: Containing the targets to link the component to (includes component's libraries and dependencies).
+- ``INTERFACE_COMPILE_DEFINITIONS``: Containing the definitions of the component.
+- ``INTERFACE_COMPILE_OPTIONS``: Containing the compile options of the component.
+
+Moreover, a global target ``<PKG-NAME>::<PKG-NAME>`` will be declared with the following properties adjusted:
+
+- ``INTERFACE_LINK_LIBRARIES``: Containing all the component targets to link the global target to (includes package's components only).
```
jgsogo, line 99:
```suggestion
- ``INTERFACE_LINK_LIBRARIES``: Containing the list of targets for all the components in the package.
```

reference/generators/cmake_find_package_multi.rst
jgsogo: Are we going to remove the `name`? The `name` assign the same value to all the generators, I think it is useful.... what we want to remove is **reading the `name` value**, it makes no sense to retrieve a _name_ if not for a concrete generator

memsharded: Not what I understood, I thought that we wanted to simplify that, and make everything explicit per-generator. But I am fine with it, just lets make sure that we are all aligned.

jgsogo: We need to maintain the setter for `name`. It will help to change the name of a package (i.e.: [pro-mdnsd](https://github.com/conan-io/conan-center-index/blob/master/recipes/pro-mdnsd/all/conanfile.py), it is not using this approach, but it would be much better):

```python
class Recipe
    name = "pro-mdnsd"

    def package_info:
        self.cpp_info.name = "mdnsd"
```

You really want all the generators to use name "mdnsd", but you cannot list all the generators... you don't even know all the generators.

PR #1759, opened 2020-07-01
Remove experimental warning

czoido:
This PR removes from experimental two conan features:

- Revisions
- The use of scm mode but only when used with `scm_to_conandata`

### DIFF
--- a/creating_packages/package_repo.rst
+++ b/creating_packages/package_repo.rst
@@ -101,8 +101,9 @@ Capturing the Remote and Commit: ``scm``
 
 .. warning::
 
-    Depending on the use, this feature may be considered **experimental** and subject to breaking
-    changes in future releases. See :ref:`note <scm_to_conandata_note>` below.
+    This is an **experimental** feature subject to breaking changes in future releases. Depending on
+    the use, this is feature may be considered **experimental**. See :ref:`note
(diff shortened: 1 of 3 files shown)


creating_packages/package_repo.rst
```diff
@@ -101,7 +101,8 @@ Capturing the Remote and Commit: ``scm``
 
 .. warning::
 
-    This is an **experimental** feature subject to breaking changes in future releases.
+    Depending on the use, this feature may be considered **experimental** and subject to breaking
+    changes in future releases. See :ref:`note <scm_to_conandata_note>` below.
```
jgsogo, line 105: I'd add everything here, not referring to another note below

reference/conanfile/attributes.rst
```diff
@@ -1257,7 +1257,8 @@ scm
 
 .. warning::
 
-    This is an **experimental** feature subject to breaking changes in future releases.
+    Depending on the use, this is feature may be considered **experimental** and subject to breaking
+    changes in future releases. See :ref:`note <scm_to_conandata_note>` below.
```
jgsogo, line 1261: Same here, I would keep the "experimental warning/note" everything inside a single callout.

PR #1766, opened 2020-07-03
Update tools.rst

wdobbe:
Indicate that tools.remove+files_by_mask() will remove files also in sub directories.


danimtb:
thank you!

PR #1788, opened 2020-07-30
Document required_conan_version for recipes

czoido:
Docs for https://github.com/conan-io/conan/pull/7360


czoido:
> I am not sure about the section where this should be placed, but I have no better suggestion right now. Contents are fine 👍

Yes, I wasn't sure as well, let's leave it there and move it in the future if we think that there's a better place.

memsharded:
Add it to: https://docs.conan.io/en/latest/configuration.html too

The idea of this section is that it is more "problem driven" than reference:

- I want to configure conan so it raises if I am using and old Conan version
- I want to configure faster downloads
etc

PR #1799, opened 2020-08-05
Docs for creating custom build helpers

czoido:
Docs on how to create a custom build helper in Conan using Python requires,


extending/custom_build_helper.rst
```diff
@@ -0,0 +1,119 @@
+.. _custom_build_helper:
+
+Creating a custom build helper for Conan
+----------------------------------------
+
+If Conan has not a build helper for the build tool you are using, you can create a custom build helper
```
lasote, line 6: It that "has not" correct?

```suggestion
If Conan doesn't have a build helper for the build tool you are using, you can create a custom build helper
```

extending/custom_build_helper.rst
```diff
@@ -0,0 +1,119 @@
...
+with the help of :ref:`python_requires`. You can create a package defining the build helper for that
```
lasote, line 7: A lot of "help" here.

```suggestion
with the :ref:`python_requires`. You can create a package defining the build helper for that
```

PR #1846, opened 2020-09-18
Add meta date to CLion page

uilianries:
/cc @memsharded @BraverO

I've added the metadata in the rst file, Sphinx should generate the same for HTML:

Screenshot at 2020-09-18 15-35-38


BraverO:
Thank you! Would it be better simple to name the page "Clion Plugin" if we cannot edit the rest of the meta title?
And in the description - should we use CLion throughout the Meta description?

So basically like this:
Meta title: 
CLion Plugin — Conan 1.29.0 documentation - Conan docs
Description: 
An official Jetbrains Conan plugin for CLion. CLion uses CMake as the build system of projects, so you can use the CMake generator to manage your requirements

uilianries:
@BraverO I made a new change, now the title and description follow your suggestion. Please, take a look.

Screenshot_2020-09-21 CLion Plugin — conan 1 29 1 documentation

uilianries:
@BraverO Done. The correct format is [CLion](https://www.jetbrains.com/clion/).

PR #1852, opened 2020-09-22
Add metadata for CLion plugin

uilianries:
First PR was not totally clear according @BraverO suggestion.

This one uses HTML metadata only for CLion Plugin and its description.

The HTML pages doesn't show the metadata content, it is only displayed when reading the HTML source.

/cc @BraverO

Current result:

Untitled


BraverO:
Looks good to go to me!

PR #1864, opened 2020-10-01
Document 'cpp_info.requires'

jgsogo:
Docs for https://github.com/conan-io/conan/pull/7644


solvingj (approved):
I'm not very extremely familiar with the implications of the feature yet, but the explanations and grammar seem correct. I'll continue reading about it and hope to try to use the feature and report back.

Issue #1831, opened 2020-09-10
[question] More details about overriding upstream options from downstream recipes


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

I still don't completely understand how and when upstream options are overridden, and which recipe takes precedence if multiple downstream recipes override options of a common upstream recipe.

https://docs.conan.io/en/latest/mastering/conditional.html claims that upstream recipes options should be overridden in `configure` method when it is absolutely necessary.

- If multiple recipes override the same upstream option, which recipe wins?
  Let's say for these two cases:
  - two direct parents override the same option of their common child.
  - one direct parent and root override the same option.

- some recipes like https://github.com/conan-community/conan-opencv conditionally override several upstream options in `requirements` method instead of `configure` method. Can we expect the same behaviour as if they would have been overridden in `configure` method (in case of conflicts with other recipes, does overriding this option in `requirements` take precedence over overriding in`configure` method, without respect of rules in first point)? Is is the same to override after `self.requires("upstream")` or before, in `requirements` method? Is it bad practice to override in `requirements` method?

PR #1867, opened 2020-10-01
Rework docs related to options

jgsogo:
close https://github.com/conan-io/docs/issues/1831

After @SpaceIm I really think it is important to clarify the documentation around the options. In this PR I try to improve current docs, although I think we should highlight a little bit some details, these sections in the reference are too long.

Suggestions welcome


reference/conanfile/attributes.rst
```diff
@@ -320,77 +320,98 @@ The ``get_safe()`` method will return ``None`` if that setting or subsetting doe
...
+        class MyPkg(ConanFile):
+
+            options = {
+                "shared": [True, False],
+                "option1": ["value1", "value2"],
+                "option2": "ANY",
+            }
+
+            def configure(self):
+                self.options.shared = False
```
memsharded, line 389: This reads like a confusing use case. Why would you have an option that results always, hardcoded, same value. That is not an option at all?

reference/conanfile/attributes.rst
```diff
@@ -320,77 +320,98 @@ The ``get_safe()`` method will return ``None`` if that setting or subsetting doe
...
     - You can inspect available package options reading the package recipe, which can be
       done with the command :command:`conan inspect mypkg/0.1@user/channel`.
+
+.. tip::
+
     - Options ``"shared": [True, False]`` and ``"fPIC": [True, False]`` are automatically managed in :ref:`cmake_reference` and
       :ref:`autotools_reference` build helpers.
 
+
+**Define the value of an option**
```
memsharded, line 367: I think this section needs to explain alternatives in order:

- First, the recommended one in the general case, ``default_options``. Specially for values of the current package (there is no example for this at the moment in the docs).
- In the configure() method. Use an example of forcing the value of a requirement option, not a self.option
- In the conanfile.txt
- In the command line

reference/conanfile/attributes.rst
```diff
@@ -458,48 +504,38 @@ consistent implementation take into account these considerations:
...
+                   "option2": "ANY"}
+        default_options = {"shared": True,
+                           "option1": "value1",
+                           "option2": 42}
 
-        def requirements(self):
-            if self.settings.os != "Windows":
-                self.requires("pkg/0.1@user/channel")
+        def build(self):
+            shared = "-DBUILD_SHARED_LIBS=ON" if self.options.shared else "-DBUILD_SHARED_LIBS=OFF"
```
memsharded, line 530: This BUILD_SHARED_LIBS is probably not a good example, as this is already done by ``CMake`` helper. Chose a generic ``cmake.definition`` based on ``option1`` or something like that.

reference/conanfile/attributes.rst
```diff
@@ -458,48 +504,38 @@ consistent implementation take into account these considerations:
...
+            self.run("cmake . %s %s" % (cmake.command_line, shared))
+            ...
 
-For this example running in Windows, the `default_options` for the `pkg/0.1@user/channel` will be ignored, they will only be used on every
-other OS.
+Remember that you can also assign default values for options of your requirements as we've seen in
+the attribute :ref:`conanfile_options`.
 
-You can also set the options conditionally to a final value with ``config_options()`` instead of using ``default_options``:
+You can also set the options conditionally to a final value with ``configure()`` instead of using ``default_options``:
```
memsharded, line 538: It seems that the intent of this has been changed. If you change from config_options => configure you are hardcoding the options values. The existing example with config_options was defining a conditional default value, that could be overriden from downstream.

memsharded (changes requested):

reference/conanfile/attributes.rst
jgsogo: The list is ordered according to precedence as said in the text.

There are examples for `default_options` in the reference entry for that attribute (for options in the same package) and here, for options declared in required packages.

reference/conanfile/attributes.rst
jgsogo: From this sentence in the docs _"`config_options()` is used to configure or constraint the available options in a package, before they are given a value."_, people understand that assigning a value in `config_options()` is forbidden (https://github.com/conan-io/conan/issues/7786#issuecomment-702945471), that the method is intended only to restrict the possible values of an option or to remove the option.

SpaceIm:
Here is a POC of conditional default option (in `config_options()`):

https://github.com/conan-io/conan-center-index/pull/3341#issuecomment-717958354

Doc could be update with an example of confitional default option.

reference/conanfile/attributes.rst
```diff
@@ -320,77 +320,99 @@ The ``get_safe()`` method will return ``None`` if that setting or subsetting doe
...
-
-        def build(self):
-            shared = "-DBUILD_SHARED_LIBS=ON" if self.options.shared else ""
-            cmake = CMake(self)
-            self.run("cmake . %s %s" % (cmake.command_line, shared))
-            ...
+        options = {
+            "shared": [True, False],
+            "option1": ["value1", "value2"],
+            "option2": "ANY",
```
Minimonium, line 347: We could enumerate some other interesting option declarations
```suggestion
            "option2": "ANY",
            "option3": [None, "value1", "value2"],
            "option4": [True, False, "value"],
```

reference/conanfile/attributes.rst
Minimonium: But it also seems that the CCI relies on the current behavior of `config_options` to allow overridable from downstream default_options: https://github.com/conan-io/conan-center-index/pull/3341#issuecomment-717958354
What's the intended behaviour in here?

reference/conanfile/attributes.rst
```diff
@@ -320,77 +320,99 @@ The ``get_safe()`` method will return ``None`` if that setting or subsetting doe
...
-.. code-block:: python
 
-    class MyPkg(ConanFile):
-        ...
-        options = {"shared": [True, False],
-                   "option1": ["value1", "value2"],}
+A very common one is the option ``shared`` with allowed values of ``[True, False]`` that many recipes declare and use to configure the
+build system and link static or dynamically the dependencies.
 
 Values for each option can be typed or plain strings, and there is a special value, ``ANY``, for
```
jgsogo, line 337:
```suggestion
Values for each option can be typed or plain strings (``"value"``, ``True``, ``None``,...) and there is a special value, ``"ANY"``, for
```

reference/conanfile/attributes.rst
jgsogo: ATM I think we should describe the current behavior. 

In this section I prefer to use ``configure()`` over ``config_options()`` because it is talking about _"assigning a final value"_ and I understand it means a value that cannot be overridden.

reference/conanfile/attributes.rst
```diff
@@ -405,32 +429,57 @@ go over all of them for the example recipe ``mypkg`` defined above:
...
-- Last way of defining values for options, with the highest priority over them all, is to pass
-  these values using the command argument :command:`-o` in the command line:
+- Last way of defining values for options is to pass these values using the command argument :command:`-o,--option` in the command line:
 
   .. code-block:: bash
 
-    $ conan install . -o MyPkg:shared=True -o OtherPkg:option=value
+    $ conan install . -o mypkg:shared=True
+
+Regarding the precedence of all these ways of assigning a value to an option, it works like any othe configuration element in Conan:
```
danimtb, line 444:
```suggestion
Regarding the precedence of all these ways of assigning a value to an option, it works like any other configuration element in Conan:
```

danimtb (approved):

jgsogo:
Here we need to say a word about lockfiles and their precedence

solvingj (approved):

danimtb:
Let's get this merged 🙏

mastering/conditional.rst
```diff
@@ -48,6 +49,12 @@ Here is an example of what we could do in our **configure method**:
           else:
              self.requires("openssl/1.0.2u")
 
+      def build(self):
+          # We can check the final values of options of our requirements
+          if self.options['poco'].that_option != "bar":
```
memsharded, line 54: Now in the ``validate()`` method, but it is good here for this PR

reference/conanfile/attributes.rst
```diff
@@ -320,77 +320,104 @@ The ``get_safe()`` method will return ``None`` if that setting or subsetting doe
...
+    options for some configurations.
 
-.. code-block:: python
 
-    class MyPkg(ConanFile):
-        ...
-        options = {"shared": [True, False],
-                   "option1": ["value1", "value2"],}
+A very common one is the option ``shared`` with allowed values of ``[True, False]`` that many recipes declare and use to configure the
+build system and link static or dynamically the dependencies.
```
memsharded, line 335:
```suggestion
build system to produce a static library or a shared library.
```

reference/conanfile/attributes.rst
```diff
@@ -405,32 +432,57 @@ go over all of them for the example recipe ``mypkg`` defined above:
...
-:ref:`corresponding section<method_configure_config_options>` has examples documenting these
-use cases. However, conditionally assigning values to options can have it drawbacks as it is
-explained in the :ref:`mastering section<conditional_settings_options_requirements>`.
+.. code-block:: python
+
+    class OtherPkg(ConanFile):
+        requires = "mypkg/0.1@user/channel"
+
+        def build(self):
+            if self.options['mypkg'].shared:
```
memsharded, line 466: Move to ``validate()`` (in another PR)

memsharded (approved):

PR #1883, opened 2020-10-13
Add Clang 11

uilianries:
Related to https://github.com/conan-io/conan/pull/7871


memsharded:
There is the problem that patch releases cannot go to ``develop`` branch, otherwise it is difficult to release the docs without pulling things that might be in develop only for 1.31

memsharded:
This can wait until publication to 1.31, no big issue

PR #1895, opened 2020-10-26
detailed intro

memsharded:


index.rst
```diff
@@ -1,12 +1,30 @@
...
+Conan can manage any number of different binaries for different configurations, including different
+architectures, compilers, compiler versions, runtimes, C++ standard library, etc. When binaries are not
+available for one configuration, they can be built from sources. Conan can create,
+upload and download binaries with the same commands and flows in every platform, saving lots of time
+in development and continuous integration. The binary compatibility can be configured and customized.
+
+Conan has  a very large and active community, especially in `Github repositories <https://github.com/conan-io/conan>`_
+and `Slack #conan channel <https://cpplang-inviter.cppalliance.org/>`_. This community also creates and maintains
+packages in ConanCenter. Conan is used in production by thousands of companies, and consequently, it has a
+commitment to stability, not breaking while moving forward in Conan 1.X versions.
```
solvingj, line 26:
```suggestion
commitment to stability, with no breaking changes across all Conan 1.X versions.
```

solvingj (commented):
I think your original wording is fine, but consider the suggested change as well. The tricky part are the terms portable and universal, and even my own wording feels a bit vague.

index.rst
memsharded: I hope this means that the no breaking changes only apply forward, and nobody can expect that recipes using latest Conan 1.31 will work in Conan 1.0. I'd prefer to make this explicit if not fully clear from the current proposal.

solvingj:
Once you merge it, probably should put it through auto-format again to fix. The per-line suggestion system in github is nice, but not ideal for re-wording sentences.

PR #1925, opened 2020-11-20
describing workaround to make build_requires affect the package_id

memsharded:


devtools/build_requires.rst
```diff
@@ -18,6 +18,14 @@ In short:
 - Some of them might not even be taken into account when a package like zlib is created, such as cross-compiling it to Android (in which
   the Android toolchain would be a build requirement too).
 
+
+.. important::
+
+    ``build_requires`` are designed for packaging tools, utilities that only run at build-time, but are no part of the final binary code.
```
danimtb, line 24:
```suggestion
    ``build_requires`` are designed for packaging tools, utilities that only run at build-time, but are not part of the final binary code.
```

danimtb (approved):

PR #1946, opened 2020-12-01
Add note about shared global target in CMake components

danimtb:
Docs for https://github.com/conan-io/conan/pull/8114

### DIFF
--- a/reference/generators/cmake_find_package.rst
+++ b/reference/generators/cmake_find_package.rst
@@ -85,5 +85,6 @@ Moreover, a global target ``<PKG-NAME>::<PKG-NAME>`` will be declared with the f
 
 .. important::
 
-    If there is a global target shared by different packages, then the global target will be linked to all the components targets **across
-    the different packages**. This means that the global target will include different components targets across different packages.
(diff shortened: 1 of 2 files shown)


reference/generators/cmake_find_package.rst
```diff
@@ -82,3 +82,8 @@ If a recipe uses components, the targets generated will be ``<PKG-NAME>::<COMP-N
 Moreover, a global target ``<PKG-NAME>::<PKG-NAME>`` will be declared with the following properties adjusted:
 
 - ``INTERFACE_LINK_LIBRARIES``: Containing all the component targets to link the global target to (includes package's components only).
+
+.. important::
+
+    If there is a global target shared by different packages, then the global target will be linked to all the components targets **across
+    the different packages**. This means that the global target will include different components targets across different packages.
```
jgsogo, line 89: "**Name conflicts.-** If the name of the global target is the same for different packages, Conan will aggregate into this global target all the components from all those different packages. This means that this global target will contain information coming from different packages. For the components themselves, a name conflict will result in one of them being inaccessible without further notice."

jgsogo (approved):

PR #1947, opened 2020-12-02
validate docs

memsharded:
Docs for https://github.com/conan-io/conan/pull/8053

### DIFF
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -286,9 +286,9 @@ The :ref:`cpp_info_attributes_reference` attribute has the following properties 
   package may have: libraries, executables... Read more about this feature at :ref:`package_information_components`.
 - **requires**: **[Experimental]** List of components from the requirements this package (and its consumers) should link with. It will
   be used by generators that add support for components features (:ref:`package_information_components`).
-   
 
-If your recipe has requirements, you can access to the information stored in the ``cpp_info`` of your requirements 
+
+If your recipe has requirements, you can access to the information stored in the ``cpp_info`` of your requirements
 using the ``deps_cpp_info`` object:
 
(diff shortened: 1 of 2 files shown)


czoido (approved):

reference/conanfile/methods.rst
```diff
@@ -519,6 +519,106 @@ it is an example of a recipe for a library that doesn't support Windows operatin
...
+
+    class Pkg(ConanFile):
+        requires = "dep/0.1"
+
+        def validate(self):
+            if self.options["dep"].myoption == 2:
+                raise ConanInvalidConfiguration("Option 2 of 'dep' not supported")
+
+
+If a package uses ``compatible_packages`` feature, it should not add to those compatible packages configurations that will not be valid,
```
jgsogo, line 596: In this section, I think it is more important to show how a compatible package (a valid one) will be used to substitute an invalid package. Something with the `compiler.cppstd` is very evident: the package is invalid (will fail with `conan create`) if using C++11, but a `conan install` will use a compatible one that was built using C++14,17,20.

These comments about what packages to add to `self.compatible_packages` belong to the documentation of the compatible_packages feature.

memsharded: But this is also a valid use case, in which you can fallback to release artifacts in Linux if you don't have the debug one, which usually works in non-windows platforms. I am not sure if I understand the issue.

jgsogo: Not an issue, I think that the other use case is much more interesting regarding `validate + compatible-packages`. Falling back to Release builds I think it is more related to compatible-packages than validate, and this is the documentation for the `validate` function. One example is enough, the other is better IMO.

jgsogo (approved):

PR #1952, opened 2020-12-03
conan.tools.microsoft docs

memsharded:
Docs for: https://github.com/conan-io/conan/pull/8014


czoido:
I think we should point this one to the release/1.32.0 branch so that after the 1.33 is launched we don't lose this in the docs generated for 1.32

czoido (approved):

PR #1958, opened 2020-12-05
tools: Add description for flatten_directory()

MaciejKucia:


MaciejKucia:
This is now resolved through new option in tools.unzip

PR #1963, opened 2020-12-11
Fix conanfile.py in make.rst

zabereer:
There is no "Make" in scope, this is also consistent with docs/creating_packages/toolchains/make.rst.


memsharded:
This was fixed in https://github.com/conan-io/docs/pull/1961, it seems for some reason, it has not been updated. I need to check.

Thanks for contributing this.

PR #1971, opened 2020-12-22
fix docs about running from sources

memsharded:
Comes from https://github.com/conan-io/conan/issues/8252

Running from sources docs is outdated, ``pip install -e .`` is much better

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -132,29 +132,16 @@ Install from source
 You can run Conan directly from source code. First, you need to install Python and
 pip.
 
-Clone (or download and unzip) the git repository and install its requirements:
+Clone (or download and unzip) the git repository and install it with:
 
 .. code-block:: bash
 
-    $ git clone https://github.com/conan-io/conan.git
-    $ cd conan
-    $ pip install -r conans/requirements.txt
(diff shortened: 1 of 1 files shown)


installation.rst
```diff
@@ -132,29 +132,16 @@ Install from source
...
 
 .. code-block:: bash
 
-    $ git clone https://github.com/conan-io/conan.git
-    $ cd conan
-    $ pip install -r conans/requirements.txt
+    # clone folder name matters, to avoid imports issues
+    $ git clone https://github.com/conan-io/conan.git conan_src
+    $ cd conan_src
+    $ pip install -e .
```
danimtb, line 142: does this also install the *requirements.txt*?

SSE4: it does
(it doesn't install `requirements_dev.txt`, but it does install `requirements.txt`)

SSE4, line 142:
```suggestion
    $ python -m pip install -e .
```
I'd always recommend to use `pythom -m pip`, sometimes pip.exe belongs to the different interpreter, so it's better to use explicit

czoido (approved):

Issue #1976, opened 2021-01-07
conan_set_find_library_paths mentions CMAKE_INCLUDE_PATH twice


gerben90:
The documentation under [conan_set_find_library_paths](https://docs.conan.io/en/1.32/reference/generators/cmake.html?highlight=conan_set_find_library_paths#conan-set-find-library-paths) mentions CMAKE_INCLUDE_PATH twice:
> Sets CMAKE_INCLUDE_PATH and CMAKE_INCLUDE_PATH.

One of them should be CMAKE_LIBRARY_PATH.

gerben90:
Created a PR to fix this issue:
https://github.com/conan-io/docs/pull/1977

memsharded:
Solved by https://github.com/conan-io/docs/pull/1977, thanks!

PR #1977, opened 2021-01-07
Change duplicate entry in cmake.rst (#1976)

gerben90:


gerben90:
Created this PR to solve #1976

memsharded:
Thanks for fixing this!

This will be published when next release is out.

PR #2004, opened 2021-01-21
Make CI fail when the linkcheck of PR's fail

czoido:
We only fail the CI job if the linkcheck fails when the commit is merged to master. That avoids some spurious fails when the PR is made but is making that sometimes we do not detect broken links in the PR and after merging to master the job fails and the docs are not published. 
Temporary we could disable that check so that the PR's fail (I think it's better to fail the PR even if it's caused by spurious fails). 
IMHO in case that after the merge to master the job fails because we have spurious fails, or a link changed in between the PR was checked and the PR was merged, we should let the documentation be published and create an active alert system (slack, dashboard...) to fix the problem as soon as possible. We could also let the job fail and don't publish the docs, then through the alert system fix the problem, but my personal preference is publishing even if we have one or two links broken for a couple of hours. We should discuss the best strategy.

This PR should fail first. Then after the broken links are fixed, it should pass.

Please @uilianries, be aware that I had to put conan-community buildroot link back because the buildroot example is not merged yet on the new repo. This should be fixed when it's merged.

### DIFF
--- a/.ci/test.jenkins
+++ b/.ci/test.jenkins
@@ -32,7 +32,7 @@ node('Linux') {
         linkcheck: {
             image.inside {
                 int output = sh(script: 'make linkcheck', returnStatus: true)
-                if (output != 0 && isMaster) {
+                if (output != 0) {
                     currentBuild.result = 'FAILURE'
                     error('Stop here, linkcheck failed!')
                 }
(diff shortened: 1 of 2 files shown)


jgsogo (approved):
I'm a BIG fan of failing in the PRs. So far the Jenkins infrastructure is more robust (I haven't seen spurious failures) than the one we used to have in Travis.

PR #2006, opened 2021-01-30
[getting started] add reference to cmake section

simogasp:
Just a minor improvement.
I think it's useful to give the link to the section describing the cmake integration when mentioned in the getting started


memsharded:
Thanks!

PR #2024, opened 2021-02-18
add draft of note about revisions and line endings

solvingj:
It seems to be a daily occurrence now that users come in reporting the problem of binary missing, which turns out to be due to having multiple revisions, which turns out to be caused by different line endings on windows, caused by GIT.  

Since usually this happens after people have just read about and enabled the revisions feature, putting the note in the revisions documentation seems to be the best way to inform people about the situation BEFORE it breaks their CI flow.

### DIFF
--- a/versioning/revisions.rst
+++ b/versioning/revisions.rst
@@ -93,17 +93,22 @@ files with @CRLF character. The default line ending outside of Windows is @LF.
 
 **Problem**
 
-As a result, when Conan does export of an otherwise identical GIT repository and commit, the Conan revision will be different. 
-This has a very unfortunate consequence when the same recipe and package are uploaded from Windows and Non-Windows build machines in
-a continuous integration environment. The net consequence is that `conan install` on the "latest" revision will only see the binaries
-for one platform or the other. 
+As a result, when Conan does export of an otherwise identical GIT repository/commit on a Windows and a Non-Windows build machine,
+the Conan revisions will be different between the two. This has a very unfortunate consequence when two such packages are then 
+uploaded to a Conan repository (as is very common in continuous integration workflows). The net consequence of two such uploads, 
+is that the next time `conan install` is run for that package, the "latest" revision will only have the binaries for one platform
+or the other (whichever CI job finished and uploaded last). If `conan install` is run for the other platform, it will receive
+the following error:
+    
+    ERROR: Missing prebuilt package for <package_ref>
 
 **Workaround**
 
-The problem is unfortunately something that is external to Conan, so we cannot provide a general-purpose fix. The most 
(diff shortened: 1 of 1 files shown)


versioning/revisions.rst
```diff
@@ -76,6 +76,58 @@ Take into account that it changes the default Conan behavior. e.g:
...
+
+**Workaround**
+
+The problem is unfortunately something that is external to Conan, so we cannot provide a general-purpose fix. We
+can only explain the situation as we have done here and suggest a few ways to address the problem.  The most 
+straightforward and reasonable solution is to prevent GIT for windows from replacing the line-endings on the Windows 
+build machines. This avoids the problem completely by ensuring that the Conan Revisions are the same on Windows and 
+Non-Windows machines. There are several ways to achieve the goal. 
+
+For users who can add files at the repository level, a file named `.gitconfig` can be added which contains the following:
```
memsharded, line 113: I am not sure this works: https://git-scm.com/docs/git-config => $GIT_DIR/config

The usual workaround is ``.gitattributes`` file.

versioning/revisions.rst
```diff
@@ -76,6 +76,58 @@ Take into account that it changes the default Conan behavior. e.g:
...
+ - Using Continuous Integration services
+ - Building on multiple platforms
+ - One of the target platforms is Windows
+
+When cloning a repository on a Windows machine, the GIT client will replace all of the line endings on all of the 
+files with @CRLF character. The default line ending outside of Windows is @LF.  
+
+**Problem**
+
+As a result, when Conan does export of an otherwise identical GIT repository/commit on a Windows and a Non-Windows build machine,
```
memsharded, line 96: I think this section (and the docs in general) can be highly reduced and be as concise as possible and to the point (without losing information). Same as the workaround section. Something like:

Problem
Git will (by default) checkout files in Windows systems using CRLF line endings, effectively producing different files. As files are different, the Conan revisions will be different from the revisions computed in other platforms as Linux, resulting in missing the respective binaries in the other revision.

Solution
It is necessary to instruct Git to do the checkout with the same line endings. This can be done with a ``.gitattributes`` file, .....


At this point is more important to highlight with a ``warning`` or some other decorator about this potential problem so users don't skip reading this.

memsharded (commented):
Thanks for contributing this, a couple of quick notes:
- For docs, try to be a bit more concise
- Make sure of the .gitconfig vs .gitattributes solution

versioning/revisions.rst
solvingj: Resolved in https://github.com/conan-io/docs/pull/2028

solvingj:
Resubmitted in https://github.com/conan-io/docs/pull/2028

PR #2028, opened 2021-02-23
add note about revisions and line endings

solvingj:
Re-submitting from my own fork, incorporating feedback from @memsharded : https://github.com/conan-io/docs/pull/2024

### DIFF
--- a/versioning/revisions.rst
+++ b/versioning/revisions.rst
@@ -76,6 +76,24 @@ Take into account that it changes the default Conan behavior. e.g:
     - If you generate and upload N binary packages for a recipe with a given revision, then if you modify the recipe, and thus the recipe
       revision, you need to build and upload N new binaries matching that new recipe revision.
 
+GIT and Line Endings on Windows
+-------------------------------
+
+.. warning::
+
+  **Problem**
+
+  Git will (by default) checkout files in Windows systems using CRLF line endings, effectively producing different files. As files are different, the Conan revisions will be different from the revisions computed in other platforms such as Linux, resulting in missing the respective binaries in the other revision. 
+
+**Solution**
+
+It is necessary to instruct Git to do the checkout with the same line endings. This can be done several ways, for example, by adding a .gitattributes file:
+
+.. code-block:: ini
+
+  [auto]
+    crlf = false
+    
 Server support
 --------------
 


danimtb (approved):

versioning/revisions.rst
```diff
@@ -76,6 +76,24 @@ Take into account that it changes the default Conan behavior. e.g:
...
       revision, you need to build and upload N new binaries matching that new recipe revision.
 
+GIT and Line Endings on Windows
+-------------------------------
+
+.. warning::
+
+  **Problem**
+
+  Git will (by default) checkout files in Windows systems using CRLF line endings, effectively producing different files. As files are different, the Conan revisions will be different from the revisions computed in other platforms such as Linux, resulting in missing the respective binaries in the other revision.
```
SSE4, line 86: 1. I don't think it's unique to Git. e.g. subversion as exactly the same problem (`svn:eol-style`)
2. the problem is more broad, it's not just CRLF vs LF, but generic SCM config difference. there are other config settings which may lead into the exactly same problem (e.g. `core.symlinks` is another configuration causing similar problems)

versioning/revisions.rst
```diff
@@ -76,6 +76,24 @@ Take into account that it changes the default Conan behavior. e.g:
...
+  Git will (by default) checkout files in Windows systems using CRLF line endings, effectively producing different files. As files are different, the Conan revisions will be different from the revisions computed in other platforms such as Linux, resulting in missing the respective binaries in the other revision. 
+
+**Solution**
+
+It is necessary to instruct Git to do the checkout with the same line endings. This can be done several ways, for example, by adding a .gitattributes file:
+
+.. code-block:: ini
+
+  [auto]
+    crlf = false
```
SSE4, line 95: I'd never recommend changing it that way (for all files). why?
- some files (e.g. `.rc` resource files) should always have CRLF
- some files (e.g. `Makefile` makefiles) should always have LF
applying global `clrf=false` or `clrf=true` will cause breakage, of one sort or another (and errors aren't very descriptive and easy to debug, e.g. `common.mak:139: *** missing separator.  Stop.`)
I'd reference another good example of `.gitattributes` file here, e.g. [this one](https://gist.github.com/dpalomar/ebc64d6f56cfc25a3e5c20fe7cdfbd4f).

versioning/revisions.rst
SSE4, line 86: is it really true? as I understand ([thread](https://stackoverflow.com/questions/39408793/git-core-autocrlf-line-ending-default-setting/48230871)), the default value of `core.autocrlf` is `false`, it means git will not modify the line endings of the file, which should effectively result into the same revision for all systems.

versioning/revisions.rst
solvingj: Yeah, I thought it was a simple thing. It's much more complicated than I realized.  Not sure embedding or linking to that gist is a good idea. Can you propose a cut down subset which you think would be a good suggestion?

versioning/revisions.rst
solvingj: Because the windows installer GUI default selection is true, it's ends up getting set to "true" incidentally for some major number of people without them realizing it. Also, silent install ends up the same, and lots of people use silent install (`choco install git` etc) .

>- the default value is "false"
>- Windows installer let you choose the desired behavior but by default (if you install without changing proposed settings) it sets it to "true". This is not the software default, the installer sets the core.autocrlf system setting.

versioning/revisions.rst
SSE4: this was just a first example of .gitattributes I've found in google :)
I guess we may safely put minimal example, doing changes only for `*.py` (or even just for `conanfile.py` for minimal breakage)

SSE4: this should be enough IMO:
```
conanfile.py            text eol=lf
```
(maybe same for exported sources, if any, like `CMakeLists.txt`)

solvingj: wait, the original suggestion should work:
```
  [auto]
    crlf = false
```
it means just keep line endings unchanged.

SSE4 (approved):

uilianries (approved):

danimtb:
Now that I realize, this PR should point to master, right?

danimtb (approved):

PR #2038, opened 2021-03-01
Document CONAN_SYSTEM_INCLUDES

### LINKED ISSUE #601: Document CONAN_SYSTEM_INCLUDES
In CMake using `set(CONAN_SYSTEM_INCLUDES ON)` makes Conan add library headers as "system headers". Thanks to that the compiler doesn't check for warnings in them.

I only learned about that variable through https://github.com/conan-io/conan/issues/2629.

Note that I don't know if it is usable in other build systems, but I think it may be.

danimtb:
closes #601

### DIFF
--- a/reference/generators/cmake.rst
+++ b/reference/generators/cmake.rst
@@ -211,6 +211,7 @@ conan_global_flags()
 ++++++++++++++++++++
 
 Sets the corresponding variables to CMake's ``include_directories()`` and ``link_directories()``.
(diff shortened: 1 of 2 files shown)


solvingj (approved):

helmesjo:
How would I enable this using the `cmake_find_package` generator?

danimtb:
@helmesjo this variable is for the `cmake` generator only. There is no such option for `cmake_find_package/_multi`

helmesjo:
@danimtb Ok, thanks for the reply!

PR #2048, opened 2021-03-22
List required Conan version for features

### LINKED ISSUE #2046: Add feature's version for new methods
Hi!

As `required_conan_version` requires a specific version related to which features we want to use in recipe, we need to know Conan releases too.

Usually I need to check the changelog to discover when a feature was released, but not all are clear. For instance, `validate()` can not be found when searching for "validate" on changelog.

My proposal is adding which version corresponds for that feature. For example:

Screenshot_2021-03-19 Methods — conan 1 34 0 documentation(1)

Where 1.33.0 link points to [1.33.0 changelog](https://docs.conan.io/en/latest/changelog.html#jan-2021).

uilianries:
Add Conan version required for some popular features used on CCI.

My code editor hates extra spaces.

closes #2046

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -321,11 +321,11 @@ options
 -------
 
 Conan provides this attribute to declare traits which will affect only one reference, unlike the settings that are typically the
-same for all the recipes in a Conan graph. Options are declared per recipe, this attribute consist on a dictionary where the key is the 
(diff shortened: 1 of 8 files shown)


danimtb (approved):
I've been always a bit hesitant about adding those marks to the whole documentation as it will pollute everything with version marks. Switching from different versions of the documentation seems cleaner to me but less practical so I am ok with this at least for those features affecting the recipe that are widely adopted in CCI.

memsharded (approved):

PR #2092, opened 2021-04-29
Layouts: Docs

lasote:


solvingj:
Great work!

Letting the users encode the "layout" invariants in the recipe in a single place in a much more declarative way is a huge improvement in our abstraction.  The fact that it improves the recipe in a general way, but also enables us to solve the previous challenges to `editable` feature is a double-win. Very excited for this. 

The most important piece of feedback I have right now surrounds the use of `-if my_install` in the examples. The install folder will always contain files specific to one build configuration, just like the generators folder. In our example, we go through the whole process of Conan providing binaries to CLion, and showing how the binaries built by CLion will be available to consumers via `editable` which is awesome.  However, we only do one configuration, so a single `my_install` folder looks fine.  However, as we know, in real-world user scenarios, users work with multiple configurations. 

I know you've added this note and explained why the build folder must be passed:

> Maybe you are wondering why the install folder is not parametrized and has to be specified with the -if argument....

But I think this hides from us and the users just how big much complexity this one required argument adds to the user. The situation poses effectively the same logistical challenge for users which was posed by the `conan install` command since the beginning. We've solved the problem for the "generators" output with `self.folders.generators`, and in most cases, ensuring a unique and predictable generators folder involves some non-trivial python logic such as the one shown in the example:
```
        self.folders.build = "cmake-build-{}".format(str(conanfile.settings.build_type).lower())
        self.folders.generators = os.path.join(self.folders.build, "generators")
```
The user will now have to ensure a similarly unique and predictable install folder gets passed to Conan, and this means they'll need wrapper scripts which employ some similar logic to calculate the `--install-folder` prior to calling the Conan command. This is really really undesirable accidental complexity. 

I understand that the root cause is an unfortunate implementation detail of how Conan caches the configuration and that it will likely be possible to eliminate in 2.0.  However, I suggest that it's worth considerable time and complexity and creativity to provide SOME way for the user to be able to avoid the manual management of a unique install-folder-per-profile in Conan 1.0. I can think of many possible ways to approach it, as can everyone else I'm sure.  But, before I suggest possibilities, I want to see if others agree that this is a problem.

solvingj (changes requested):
Great work!

Letting the users encode the "layout" invariants in the recipe in a single place in a much more declarative way is a huge improvement in our abstraction.  The fact that it improves the recipe in a general way, but also enables us to solve the previous challenges to `editable` feature is a double-win. Very excited for this. 

The most important piece of feedback I have right now surrounds the use of `-if my_install` in the examples. The install folder will always contain files specific to one build configuration, just like the generators folder. In our example, we go through the whole process of Conan providing binaries to CLion, and showing how the binaries built by CLion will be available to consumers via `editable` which is awesome.  However, we only do one configuration, so a single `my_install` folder looks fine.  However, as we know, in real-world user scenarios, users work with multiple configurations. 

I know you've added this note and explained why the install folder must be passed:

> Maybe you are wondering why the install folder is not parametrized and has to be specified with the -if argument....

But I think this note fails to explain just how big much complexity this one required argument adds to the user in real world workflows. The situation poses effectively the same logistical challenge for users which was posed by the `conan install` command since the beginning. We've solved the problem for the "generators" output with `self.folders.generators`, and in most cases, ensuring a unique and predictable generators folder involves some non-trivial python logic such as the one shown in the example:
```
        self.folders.build = "cmake-build-{}".format(str(conanfile.settings.build_type).lower())
        self.folders.generators = os.path.join(self.folders.build, "generators")
```
However, we couldn't solve it for the `graph_info.json` and `conanbuildinfo.txt` in the exact same way due to order-of-operations. As a result, the user will now have to ensure a similarly unique and predictable install folder gets passed to Conan, and this means they'll need wrapper scripts which employ some similar logic to calculate the `--install-folder` prior to calling the Conan command. This is really really undesirable accidental complexity. 

I understand that the root cause is an unfortunate implementation detail of how Conan caches the configuration and that it will likely be possible to eliminate in 2.0.  And I understand we can't solve it within the `layout()` method the same way we did with the `generator` folder.  However, I suggest that it's worth considerable time and complexity and creativity to provide SOME way for the user to be able to avoid the manual management of a unique install-folder-per-profile in Conan 1.0. I can think of many possible ways to approach it, as can everyone else I'm sure.  But, before I suggest possibilities, I want to see if others agree that this is a problem.

solvingj:
I spoke with @memsharded who explained that we will not spend any additional time working around Conan 1.x behaviors which make this feature less elegant and more difficult to use.  Since the change I requested is only relevant for 1.x, we can resolve the comment as "won't do".

solvingj (approved):

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
+.. _package_layout:
+
+Package layout
+==============
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.37.0 <https://github.com/conan-io/conan/releases/tag/1.37.0>`_
```
czoido, line 10:
```suggestion
Available since: `1.37.0 <https://github.com/conan-io/conan/releases>`_
```
That link is making the ci fail, maybe we can point to the general releases link?

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+      | - src
+          | - CMakeLists.txt
+          | - hello.cpp
+          | - hello.h
+
+
+.. code:: python
+
+    import os
+    from conans import ConanFile
```
czoido, line 73:
```suggestion
    from conans import ConanFile, CMake
```

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+            self.folders.imports = self.folders.build
+            self.folders.source = "src"
+
+        def source(self):
+            # In the source method, the current directory == self.source_folder
+            assert self.source_folder == os.getcwd()
+
+        def build(self):
+            # We are at a folder like "myproject/cmake-build-debug"
+            cmake = CMake()
```
czoido, line 94:
```suggestion
            cmake = CMake(self)
```

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+
+        class Pkg(ConanFile):
+
+            def layout(self):
+                # ###### FOLDERS
+                # The sources can be found in the root dir
+                self.folders.source = ["."]
+
+                # The build folder is created with the CLion way
+                self.folders.build = "cmake-build-{}".format(str(conanfile.settings.build_type).lower())
```
czoido, line 283:
```suggestion
                self.folders.build = "cmake-build-{}".format(str(self.settings.build_type).lower())
```

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+
+                # We want to have the toolchains in the build folder so we can always pass
+                # `-DCMAKE_TOOLCHAIN_FILE=generators/conan_toolchain.cmake` to CMake
+                self.folders.generators = os.path.join(self.folders.build, "generators")
+
+                # In case we use "conan package" we declare an output directory
+                self.folders.package = "package-{}".format(str(conanfile.settings.build_type).lower())
```
czoido, line 290:
```suggestion
                self.folders.package = "package-{}".format(str(self.settings.build_type).lower())
```

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+
+        from conans import ConanFile
+        from conan.tools.layout import LayoutPackager
+
+        class Pkg(ConanFile):
+
+            def layout(self):
+                # ###### FOLDERS
+                # The sources can be found in the root dir
+                self.folders.source = ["."]
```
czoido, line 280:
```suggestion
                self.folders.source = "."
```

developing_packages/package_layout.rst
```diff
@@ -0,0 +1,326 @@
...
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.37.0 <https://github.com/conan-io/conan/releases>`_
+
+You can declare a ``layout()`` method in the recipe to describe the package contents,
+not only the final package in the cache but also the package while developing.
+As the package will have the same structure in the cache and in our local directory, the recipe development become easier.
```
czoido, line 14:
```suggestion
As the package will have the same structure in the cache and in our local directory, the recipe development becomes easier.
```

czoido (approved):

reference/conanfile/methods.rst
```diff
@@ -1445,3 +1445,93 @@ multiple recipes.
...
+
+layout()
+--------
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.37.0 <https://github.com/conan-io/conan/releases/tag/1.37.0>`_
+
```
czoido, line 1458:
```suggestion
Available since: `1.37.0 <https://github.com/conan-io/conan/releases>`_

```

reference/conanfile/tools/layout.rst
```diff
@@ -0,0 +1,38 @@
+.. _conan_tools_layout:
+
+conan.tools.layout
+==================
+
+Available since: `1.37.0 <https://github.com/conan-io/conan/releases/tag/1.37.0>`_
```
czoido, line 6:
```suggestion
Available since: `1.37.0 <https://github.com/conan-io/conan/releases>`_
```

reference/conanfile/methods.rst
```diff
@@ -1445,3 +1445,93 @@ multiple recipes.
...
+Read about the feature :ref:`here<package_layout>`.
+
+In the layout() method you can adjust ``self.folders``, ``self.cpp`` and ``self.patterns``.
+
+self.folders
+++++++++++++
+
+- **self.folders.source** (Defaulted to ""): Specifies a subfolder where the sources are. The ``self.source_folder`` attribute and
```
memsharded, line 1466: If this is a copy of the other section, maybe leave this here, and simplify/reduce a little bit the other section, and use cross-reference?

reference/conanfile/tools/layout.rst
```diff
@@ -0,0 +1,38 @@
...
+
+LayoutPackager
```
memsharded, line 8: I don't love this name, but I cannot think of anything better either.

memsharded (approved):
Very good job, I think it is clear docs.

PR #2099, opened 2021-05-20
Docs for tools.rename()

danimtb:
Docs for conan-io/conan#8915
Issue conan-io/conan#8897

### DIFF
--- a/reference/conanfile/tools/files.rst
+++ b/reference/conanfile/tools/files.rst
@@ -84,6 +84,6 @@ windows. This function renames file or folder using :command:`robocopy` to avoid
       tools.rename(self, "lib-sources-abe2h9fe", "sources")  # renaming a folder
 
 Parameters:
-    - **conanfile**: Conanfile object.
-    - **src** (Required): Path to be renamed.
-    - **dst** (Required): Path to be renamed to.
+  - **conanfile**: Conanfile object.
+  - **src** (Required): Path to be renamed.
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/files.rst
```diff
@@ -67,3 +67,23 @@ Example of ``conandata.yml`` with different patches for different versions:
...
+
+conan.tools.rename()
+--------------------
+
+.. code-block:: python
+
+    def rename(conanfile, src, dst)
+
+Utility functions to rename a file or folder *src* to *dst* with retrying. ``os.rename()`` frequently raises "Access is denied" exception on
+windows. This function renames file or folder using :command:`robocopy` to avoid the exception on Windows.
```
czoido, line 79: Here we can clarify that the retry will be done in Windows if robocopy is found in the system .
Also, we may limit in Conan the number of retries, looks like it will retry 1.000.000 times with 30 seconds intervals by default? https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/robocopy

danimtb, line 79:
```suggestion
Utility functions to rename a file or folder *src* to *dst*. On Windows, it is very common that ``os.rename()`` raises an "Access is denied" exception, so this tool uses:command:`robocopy` if available. If that is not the case, or the rename is done in a non-Windows machine, it falls back to the ``os.rename()`` implementation.
```

czoido (approved):

PR #2114, opened 2021-05-31
cmake docs split into files

memsharded:
This section is getting too long, better move each thing to its own file.


lasote (approved):
Totally agree. If only moved contents feel free to merge.

PR #2118, opened 2021-06-03
add more layout packager details

solvingj:
I was reading the documentation and trying to understand the `layouts()` feature in order to write the blog post.  One thing I noticed is that it feels difficult to comprehend all the implications of all the values one can set in layouts (at least for me).  

I think this makes sense and is a natural consequence of the tradeoffs we're chosing to make.  Previously, `self.copy()` had the nice property that the `src=` and `dest=` and `pattern=`  were all basically declared exactly where they were used. However, now that we've realized we need to use that information in other places, and compose that information in different ways, we needed to raise the level of abstraction. The unfortunate loss is that it will now be difficult to look at a recipe and determine what patterns will be copied from what directories.  You have to do the composition in your head, which looks like it's going to be exceptionally difficult because it's a matrix of root paths, sub-paths and patterns, and then doubled due to `cpp.source` and `cpp.build`.  

This PR is actually just an initial pass at helping people connect the dots. I think we need to spend quite a bit more effort making it "make sense".

### DIFF
--- a/reference/conanfile/tools/layout.rst
+++ b/reference/conanfile/tools/layout.rst
@@ -83,7 +83,7 @@ Here is the equivalent `package()` method:
 
 
 In the example above, we defined explicit values for the directories and
-patterns for illustration purposes. However, the many of the directories and
+patterns for illustration purposes. However, many of the directories and
 patterns have default values. So, here we show the equivalent `self.copy()`
 statements for the case where ``LayoutPackager.package()`` is used with the
 default values of `self.cpp` and `self.patterns`. 
--- a/reference/conanfile/tools/layout.rst
+++ b/reference/conanfile/tools/layout.rst
@@ -8,19 +8,9 @@ Available since: `1.37.0 <https://github.com/conan-io/conan/releases>`_
 LayoutPackager
 --------------
 
-The ``LayoutPackager`` together with the :ref:`package layouts<package_layout>` feature, allow to automatically
-package the files following the declared information in the ``layout()`` method:
-
-It will copy (being xxx => ``build`` and ``source`` (if destination is only one dir):
-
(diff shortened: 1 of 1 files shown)


solvingj:
For those reviewing this PR, I want to re-iterate that you consider the following:  

I think one of the goals of the `layout()` method is to unify/simplify the relationship between a new "local flow", and the `source()`/`build()`/`package()` methods.  However, with this first release, we've focused on adding the necessary power and flexibility which makes it more complex to understand and use.  I think that's a totally reasonable challenge to have in a first release of an ambitious and powerful feature. 

The suggestion here is simply to look for simple/creative ways to reduce that complexity for the next release if possible. 

I think the first and most obvious thing people are likely to do with `layouts()` is try to replace an existing `package()` method with `layout()` + `package()` + `LayoutPackager()`, and THEN try the local flow with the new recipe. So, perhaps think about users who are trying to achieve "functional equivalence" between `self.copy` and `LayoutPackager`.  

Also, think about what to do for ultra-common case of package methods which just call `cmake.install()`.  To be honest, unfortunately, these recipes seem like a challenge to the new feature.

lasote (commented):
I like the example and the easy explanation. Some observations:

- That page is the "reference" section where we usually don't write examples, is more intended to be a reference of attributes/methods. So it might be better to move elsewhere.
- For me, the `self.cpp.build` and `self.cpp.source` objects are primarily intended for editable packages rather than auto-packaging, which is a nice consequence of describing the build and source folders. So I agree the example might give the impression of having bedeviled the way we can package artifacts, but just because the main idea is not "now the way to package is LayoutPackage" but: 
    - Same experience with local commands=> you can use local methods and conan create the same way, especially in "conan 2.0" where the syntax is the same.
    - Same experience with editable packages=> you can put the package as editable and everything keeps working.
    - Nice consequences of describing your package come for free, like auto packaging.

memsharded:
Closing as outdated

PR #2124, opened 2021-06-15
Feature/conf cli support

franramirez688:
Added documentation related to new conan/feature: https://github.com/conan-io/conan/pull/9103

Added new parameter -c, --conf to provide inline configuration

```
  -c CONF_HOST, --conf CONF_HOST
                        Configuration to build the package, overwriting the defaults (host machine). e.g.: -c
                        tools.cmake.cmaketoolchain:generator=Xcode
  -c:b CONF_BUILD, --conf:build CONF_BUILD
                        Configuration to build the package, overwriting the defaults (build machine). e.g.: -c:b
                        tools.cmake.cmaketoolchain:generator=Xcode
  -c:h CONF_HOST, --conf:host CONF_HOST
                        Configuration to build the package, overwriting the defaults (host machine). e.g.: -c:h
                        tools.cmake.cmaketoolchain:generator=Xcode
```

You can use these options for the following commands: export_pkg, test, create, install, info, workspace install and lock create

### DIFF
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -322,6 +322,24 @@ With the :command:`-o` parameters you can only define specific package options.
     You can use :ref:`profiles <profiles>` files to create predefined sets of **settings**,
     **options** and **environment variables**.
 
+conf
+----
+
+With the :command:`-c` parameters you can define specific package configurations.
+
+.. code-block:: bash
+
+    $ conan install . -c tools.microsoft.msbuild:verbosity=Diagnostic
+    $ conan install . -c core:required_conan_version>=1.34 -c tools.build:processes=10
+
+
+.. note::
+    To list all possible configurations available, run :command:`conan config list`.
+
+.. seealso::
+
+    You can see more information about configurations in :ref:`global.conf section <conan_cfg>`.
+
 
 reference
 ---------
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -325,6 +325,10 @@ With the :command:`-o` parameters you can only define specific package options.
 conf
 ----
 
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
 With the :command:`-c` parameters you can define specific package configurations.
 
 .. code-block:: bash
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -338,7 +338,7 @@ With the :command:`-c` parameters you can define specific package configurations
 
 .. seealso::
 
-    You can see more information about configurations in :ref:`global.conf section <conan_cfg>`.
+    You can see more information about configurations in :ref:`global.conf section <global_conf>`.
 
 
 reference
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -15,6 +15,7 @@ conan install
                     [-pr PROFILE_HOST] [-pr:b PROFILE_BUILD]
                     [-pr:h PROFILE_HOST] [-s SETTINGS_HOST]
                     [-s:b SETTINGS_BUILD] [-s:h SETTINGS_HOST]
+                    [-c CONF_HOST] [-c:b CONF_BUILD] [-c:h CONF_HOST]
                     [--lockfile-node-id LOCKFILE_NODE_ID]
                     path_or_reference [reference]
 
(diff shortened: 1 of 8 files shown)


danimtb (approved):

reference/commands/consumer/install.rst
```diff
@@ -312,6 +322,24 @@ With the :command:`-o` parameters you can only define specific package options.
...
 
+conf
+----
+
+With the :command:`-c` parameters you can define specific package configurations.
+
+.. code-block:: bash
+
+    $ conan install . -c tools.microsoft.msbuild:verbosity=Diagnostic
+    $ conan install . -c core:required_conan_version>=1.34 -c tools.build:processes=10
```
memsharded, line 333: The ``core`` configuration does not work that way, because it is checked elsewhere in the flow, it is not obtained from the profile.

I'd recommend to document things from the tests of the feature, wherever possible, this is wrongly assuming this will work.

memsharded (changes requested):

reference/commands/consumer/install.rst
franramirez688: Absolutely right. Fixed!

reference/commands/consumer/install.rst
```diff
@@ -312,6 +322,24 @@ With the :command:`-o` parameters you can only define specific package options.
     You can use :ref:`profiles <profiles>` files to create predefined sets of **settings**,
     **options** and **environment variables**.
 
+conf
+----
+
```
memsharded, line 327: Add here a note that the [conf] is an experimental feature

reference/profiles.rst
```diff
@@ -131,6 +131,29 @@ They accept patterns too, like ``-s *@myuser/*``, which means that packages that
         [env]
         PATH=[/some/path/to/my/tool]
 
+Tools configurations
+--------------------
+
```
memsharded, line 136: Add here a note that the [conf] is an experimental feature

memsharded (approved):

reference/profiles.rst
franramirez688: Done! 😉

reference/commands/consumer/install.rst
franramirez688: Done 😉

reference/commands/creator/create.rst
```diff
@@ -143,6 +144,16 @@ to know more about 'test_folder' project.
...
                             defaults (host machine). e.g.: -s:h compiler=gcc
+      -c CONF_HOST, --conf CONF_HOST
+                            Configuration to build the package, overwriting the defaults (host machine). e.g.: -c
+                            tools.cmake.cmaketoolchain:generator=Xcode
+      -c:b CONF_BUILD, --conf:build CONF_BUILD
+                            Configuration to build the package, overwriting the defaults (build machine). e.g.: -c:b
+                            tools.cmake.cmaketoolchain:generator=Xcode
+      -c:h CONF_HOST, --conf:host CONF_HOST
+                            Configuration to build the package, overwriting the defaults (host machine). e.g.: -c:h
+                            tools.cmake.cmaketoolchain:generator=Xcode
```
franramirez688, line 155: @memsharded should I add a quote like "EXPERIMENTAL" at the beginning of each explanation?

franramirez688:
@memsharded it's ready for merging 😁

memsharded (approved):

reference/commands/consumer/install.rst
```diff
@@ -312,6 +322,28 @@ With the :command:`-o` parameters you can only define specific package options.
...
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+With the :command:`-c` parameters you can define specific package configurations.
```
ytimenkov, line 332: Being pedantic: "config" defines not "package" configuration, but rather "tool" one. It's not per-package style (like settings or options `zlib:tools.cmake.cmaketoolchain:generator=Xcode`), is it?

memsharded: Actually, it is, it would totally make sense to force one generator to a specific package but not the others, wouldn't it? For example, you could use Ninja generator as default, but Visual Studio one for a certain package that you need to open the IDE for working on it.

ytimenkov: > Actually, it is, it would totally make sense to force one generator to a specific package but not the others, wouldn't it?

Yes. Another example is if some package can't be built with certain generator or in parallel.

But I'm not requesting this feature :). My suggestion was to replace word "package" with word "tool" because current text looks lika an incorrect copy-paste...

memsharded: Ah, yes, you are right! I misunderstood it :)

Please @franramirez688, a quick change to this would be nice.

PR #2133, opened 2021-06-25
Added PkgConfigDeps generator

franramirez688:
Added documentation related to new conan/feature: conan-io/conan#9152

Added new `PkgConfigDeps` generator in `conan.tools.gnu.pkgconfigdeps.py`

### DIFF
--- a/reference/conanfile/tools/gnu/pkgconfigdeps.rst
+++ b/reference/conanfile/tools/gnu/pkgconfigdeps.rst
@@ -56,30 +56,31 @@ And it can also be fully instantiated in the conanfile ``generate()`` method:
 
     class App(ConanFile):
         settings = "os", "arch", "compiler", "build_type"
+        requires = "zlib/1.2.11"
 
         def generate(self):
             pc = PkgConfigDeps(self)
             pc.generate()
 
-The ``PkgConfigDeps`` will generate after a ``conan install`` command the `*.pc`:
+The ``PkgConfigDeps`` will generate the `*.pc` file after a ``conan install`` command:
 
 .. code-block:: bash
 
     $ conan install .
     # Check the [PC_FILE_NAME].pc created in your current folder
 
-For instance, the `.pc` created could look like this one:
+Now, running this command using the previous ``conanfile.py``, you can check the `zlib.pc` file created into your current folder:
 
 .. code-block:: text
 
-    libdir=/my_absoulte_path/mylib/lib
-    libdir2=${prefix}/lib2
-    includedir=/my_absoulte_path/mylib/include
+    prefix=/Users/YOUR_USER/.conan/data/zlib/1.2.11/_/_/package/647afeb69d3b0a2d3d316e80b24d38c714cc6900
+    libdir=${prefix}/lib
+    includedir=${prefix}/include
 
-    Name: MyLib
-    Description: Conan package: MyLib
-    Version: 0.1
-    Libs: -L"${libdir}" -L"${libdir2}"%s
+    Name: zlib
+    Description: Conan package: zlib
+    Version: 1.2.11
+    Libs: -L"${libdir}" -lz  -Wl,-rpath,"${libdir}" -F Frameworks
     Cflags: -I"${includedir}"
 
 
--- a/reference/conanfile/tools/gnu/pkgconfigdeps.rst
+++ b/reference/conanfile/tools/gnu/pkgconfigdeps.rst
@@ -1,0 +1,83 @@
+.. _conan_tools_pkgconfig:
+
+
+conan.tools.gnu.pkgconfigdeps
+=============================
+
+.. warning::
+
+    These tools are **experimental** and subject to breaking changes.
+
+
+PkgConfigDeps
+-------------
+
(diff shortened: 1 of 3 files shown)


reference/conanfile/tools/gnu/pkgconfigdeps.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+PkgConfigDeps
+-------------
+
+Available since: `1.38.0 <https://github.com/conan-io/conan/releases/tag/1.38.0>`_
+
+
+The ``PkgConfigDeps`` is the dependencies generator for pkg-config. Generates pkg-config files named *<PKG-NAME>.pc*
+(where ``<PKG-NAME`` is the name declared by dependencies in ``cpp_info.name`` or in ``cpp_info.names["pkg_config"]``
```
franramirez688, line 19: It's copied from the original [PkgConfig generator](https://docs.conan.io/en/latest/reference/generators/pkg_config.html) but I have doubts if I have to add/change something else...

reference/conanfile/tools/gnu/pkgconfigdeps.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+Components
+++++++++++
+
+If a recipe uses :ref:`components<package_information_components>`, the files generated will be *<COMP-NAME>.pc* with their corresponding
+flags and require relations.
+
+Additionally, a *<PKG-NAME>.pc* is generated to maintain compatibility for consumers with recipes that start supporting components. This
+*<PKG-NAME>.pc* file will declare the all the components of the package as requires while the rest of the fields will be empty, relying on
```
czoido, line 79:
```suggestion
*<PKG-NAME>.pc* file will declare all the components of the package as requires while the rest of the fields will be empty, relying on
```

reference/conanfile/tools/gnu/pkgconfigdeps.rst
```diff
@@ -0,0 +1,83 @@
...
+the propagation of flags coming from the components *<COMP-NAME>.pc* files.
+
+Go to :ref:`Integrations/pkg-config and pc files/Use the pkg_config generator<pkg_config_generator_example>`
+if you want to learn how to use this generator.
```
czoido, line 83:
```suggestion
```
That points to the old `pkg_config` generator

reference/conanfile/tools/gnu/pkgconfigdeps.rst
czoido: You can also add that it can be defined with `pkg_config_name` using `set_property`, (that will be the default way of setting this for 2.0). Please check this: https://docs.conan.io/en/latest/reference/conanfile/attributes.html?highlight=cpp_info%20set_property

czoido (approved):

reference/conanfile/tools/gnu/pkgconfigdeps.rst
memsharded: Actually, for new stuff, we can push as much as we want for new features. If you want to describe ``set_property()`` as the mechanism for this it is fine.  Not necessary to change the PR, it is good so far, but just the idea for moving forward.

reference/conanfile/tools/gnu/pkgconfigdeps.rst
```diff
@@ -0,0 +1,94 @@
...
+.. code-block:: bash
+
+    $ conan install .
+    # Check the [PC_FILE_NAME].pc created in your current folder
+
+For instance, the `.pc` created could look like this one:
+
+.. code-block:: text
+
+    libdir=/my_absoulte_path/mylib/lib
```
memsharded, line 75: I'd probably use a real example, make the above ``App`` depend on ``zlib/1.2.11`` and copy here the generated .pc, do not invent one.

reference/conanfile/tools/gnu/pkgconfigdeps.rst
```diff
@@ -0,0 +1,94 @@
...
+    Description: Conan package: MyLib
+    Version: 0.1
+    Libs: -L"${libdir}" -L"${libdir2}"%s
+    Cflags: -I"${includedir}"
+
+
+Components
+++++++++++
+
+If a recipe uses :ref:`components<package_information_components>`, the files generated will be *<COMP-NAME>.pc* with their corresponding
```
memsharded, line 89: Isn't the name ``PKG-NAME_COMP-NAME.pc``? I think it could be better this way, to avoid potential conflicts with components of different packages. Or is not possible?

memsharded (approved):
Good enough to be merged for the release, but a couple of notes to take into account (can be addressed later)

reference/conanfile/tools/gnu/pkgconfigdeps.rst
franramirez688: The thing is that I got this doc-lines about the Components from this one https://docs.conan.io/en/latest/reference/generators/pkg_config.html Anyway, I'm gonna check it 😉

lasote (approved):
Probably we want to change the pc file names when components using the @memsharded suggestion for the next release.

PR #2141, opened 2021-06-29
cmake_layout

memsharded:
Docs for https://github.com/conan-io/conan/pull/9057


lasote (changes requested):
I think it might be useful and didactic to document that the build folder is xxx for single config and yyy for multi and the default for the cpp.build and cpp.source. Otherwise is too much of a magic black box with unknown behavior

memsharded:
> I think it might be useful and didactic to document that the build folder is xxx for single config and yyy for multi and the default for the cpp.build and cpp.source. Otherwise is too much of a magic black box with unknown behavior

Agree, I will add its implementation and comment it.

lasote (approved):
👏

PR #2157, opened 2021-07-13
remove conanfile reference from tools.rename

Renari:
it looks like passing in a conanfile is no longer required for rename


memsharded (changes requested):
It seems you are mistaken, there are 2 renames:

- ``conan.tools.rename()`` this one, that receives the conanfile. New one for Conan 2.0
- ``conans.tools.rename()`` legacy one, that doesn't receive conanfile, will be removed in Conan 2.0

Renari:
Ah alright then.

PR #2160, opened 2021-07-19
& to specify consumer packages

lasote:
Docs for https://github.com/conan-io/conan/pull/9267

### DIFF
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -302,7 +302,7 @@ With the :command:`-s` parameters you can define:
 - Global settings (:command:`-s compiler="Visual Studio"`). Will apply to all the requires.
 - Specific package settings (:command:`-s zlib:compiler="MinGW"`). Those settings will be applied only to
   the specified packages. They accept patterns too, like ``-s *@myuser/*:compiler=MinGW``, which means that packages that have the username "myuser" will use MinGW as compiler.
-- Settings only for the consumer package. (:command:`-s &:compiler="MinGW"`). If `&` is specified as the package name it will apply
+- **Experimental:** Settings only for the consumer package. (:command:`-s &:compiler="MinGW"`). If `&` is specified as the package name it will apply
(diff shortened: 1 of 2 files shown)


reference/profiles.rst
```diff
@@ -119,6 +119,15 @@ They accept patterns too, like ``-s *@myuser/*``, which means that packages that
     compiler.version=4.9
     compiler.libcxx=libstdc++11
 
+Also `&` can be specified as the package name. It will apply only to the consumer conanfile (.py or .txt).
+This is a special case because the consumer conanfile might not declare a `name` so it would be impossible to reference it.
+
```
memsharded, line 124:
```suggestion
Also, as a **experimental** feature, `&` can be specified as the package name. It will apply only to the consumer conanfile (.py or .txt).
This is a special case because the consumer conanfile might not declare a `name` so it would be impossible to reference it.

```

reference/commands/consumer/install.rst
```diff
@@ -302,7 +302,9 @@ With the :command:`-s` parameters you can define:
 - Global settings (:command:`-s compiler="Visual Studio"`). Will apply to all the requires.
 - Specific package settings (:command:`-s zlib:compiler="MinGW"`). Those settings will be applied only to
   the specified packages. They accept patterns too, like ``-s *@myuser/*:compiler=MinGW``, which means that packages that have the username "myuser" will use MinGW as compiler.
-
+- Settings only for the consumer package. (:command:`-s &:compiler="MinGW"`). If `&` is specified as the package name it will apply
+  only to the consumer conanfile (.py or .txt). This is a special case because the consumer conanfile might not declare a `name` so
+  it would be impossible to reference it.
```
memsharded, line 307:
```suggestion
- **Experimental:** Settings only for the consumer package. (:command:`-s &:compiler="MinGW"`). If `&` is specified as the package name it will apply
  only to the consumer conanfile (.py or .txt). This is a special case because the consumer conanfile might not declare a `name` so
  it would be impossible to reference it.
```

memsharded (changes requested):

czoido (approved):

PR #2178, opened 2021-07-28
Use validate instead of configure

uilianries:
The current documentation shows `ConanInvalidConfiguration` associated mainly to `configure` method. However, since Conan 1.32, the new method `validate` has been used for validation, which is well reflected on CCI.

As this is not clear in Docs, users still can [get confused](https://github.com/conan-io/conan-center-index/pull/6468#discussion_r678351314) about the correct usage.

This PR replaces `configure` by `validate` when assuming `ConanInvalidConfiguration` and `check_min_cppstd`. The difference between both is well explained [here](https://docs.conan.io/en/latest/reference/conanfile/methods.html#validate), so I believe this PR is clear enough in terms of need.

### DIFF
--- a/devtools/create_installer_packages.rst
+++ b/devtools/create_installer_packages.rst
@@ -42,7 +42,7 @@ the ``nasm`` tool for building assembler:
        settings = "os", "arch"
        description="Nasm for windows. Useful as a build_require."
 
-       def configure(self):
+       def validate(self):
            if self.settings.os != "Windows":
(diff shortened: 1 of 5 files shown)


memsharded (approved):
Please target "master" branch, so this is published now, no need to wait for next release.

uilianries:
> Please target "master" branch, so this is published now, no need to wait for next release.

Done.

memsharded (approved):

PR #2179, opened 2021-07-29
Make sanitizer_hook.py actually work

mathbunnyru:


howtos/sanitizers.rst
```diff
@@ -155,15 +155,17 @@ flags to the build. It could be something like:
...
     :caption: *sanitizer_hook.py*
 
-    def set_sanitize_address_flag(self):
-        self._old_cxx_flags = os.environ.get("CXXFLAGS")
-        os.environ["SOURCE_DATE_EPOCH"] = _old_flags + " -fsanitize=address"
+    _OLD_CXX_FLAGS = None
 
-    def reset_sanitize_address_flag(self):
-        if self._old_cxx_flags is None:
+    def set_sanitize_address_flag():
```
memsharded, line 160: It is not totally evident why the ``self.old_cxx_flags`` doesn't work, is it because it is called from different instances of the hook?

memsharded (commented):
While this could be fixing this hook, I am not convinced that this is good practice for defining the environment and injecting sanitizers. I think this should be better explicitly defined in the profiles, and we have the mechanism for this (new ``[buildenv]`` better than ``[env]``).

howtos/sanitizers.rst
mathbunnyru: What is `self._old_cxx_flags`?
It doesn't seem, that we're in some class method here, so `self` doesn't make sense for me.
That's why we have to somehow save the state between function calls, that's why I used global variable as the simplest way to do so.

czoido:
Hi @mathbunnyru,
Yes, there are some typos and lack of information in the example of the docs that need to be fixed, in its origin I think that the full source code of the hook would be something like this:

```
import os


class SanitizerHook(object):
    def __init__(self):
        self._old_cxx_flags = ""

    def set_sanitize_address_flag(self):
        self._old_cxx_flags = os.environ.get("CXXFLAGS", "")
        os.environ["CXXFLAGS"] = self._old_cxx_flags + " -fsanitize=address"

    def reset_sanitize_address_flag(self):
        if self._old_cxx_flags is None:
            del os.environ["CXXFLAGS"]
        else:
            os.environ["CXXFLAGS"] = self._old_cxx_flags


sanitizer = SanitizerHook()


def pre_build(output, conanfile, **kwargs):
    sanitizer.set_sanitize_address_flag()


def post_build(output, conanfile, **kwargs):
    sanitizer.reset_sanitize_address_flag()
```
I think we should have added more context to the docs because it's unclear. Maybe adding the full hook in the docs would fix this.

mathbunnyru:
@czoido I applied your suggestion, but with some fixes.
In your example self._old_cxx_flags is never `None`, I fixed this.

howtos/sanitizers.rst
```diff
@@ -155,15 +155,34 @@ flags to the build. It could be something like:
...
+
+
+    def pre_build(output, conanfile, **kwargs):
+        sanitizer.set_sanitize_address_flag()
+
+
+    def post_build(output, conanfile, **kwargs):
+        sanitizer.reset_sanitize_address_flag()
 
 And then calling those functions from a *pre_build* and a *post_build* hook:
```
czoido, line 187: If we are adding all the source code maybe we could remove this last part.

mathbunnyru: Done, thanks.

czoido:
I overlooked the original @memsharded [comment](https://github.com/conan-io/docs/pull/2179#pullrequestreview-718210634) and I think he is right, this should be better explicitly defined in the profiles. Maybe we should remove the hook example and stop recommending this way of adding the sanitizers?

mathbunnyru:
Let's maybe for now add the warning that this method is not recommended for production usage?

PR #2183, opened 2021-08-05
An example of profile template composition.

ytimenkov:


reference/profiles.rst
```diff
@@ -337,6 +337,20 @@ Some of the capabilities of the profile templates are:
        [conf]
        tools.cmake.cmaketoolchain:toolchain_file = {{ os.path.join(profile_dir, "toolchain.cmake") }}
 
+- Including or importing other files from ``profiles`` folder:
```
memsharded, line 340: I am not 100% sure this always works. The syntax for using resources from other files is ``{% import "profile_vars.jinja" as vars %}``.

If this works, better add a unittest to Conan itself first, to guarantee it is working as expected.

memsharded (changes requested):

reference/profiles.rst
ytimenkov: It should. I was looking at the PR to check whether you pass Loader or not. And you do (FileLoader with base_path).

So it opens a lot of possibilities to compose profiles, for example allow to adjust some parameters by creating local files (e.g. via `{% include "user_vars.jinja" ignore missing %}`).

I can change example to import if it is what you ask (and of course add a test). Actually I use more imports with macros and variables rather than inheritance.

memsharded: It is fine to use jinja blocks and if the above works, it is ok to document it. But it would not be the first time I get something wrong in jinja loading, so it would be necessary to test it in Conan too, also for guaranteeing that we will no be breaking a documented feature in the future.

ytimenkov: Sorry for the delay. Was busy with customer issue :(

I was about to write a test but found out that the existing test `test_profile_template_inclusion` does exactly that: loads a file from profile directory.

Both `import` and `extends` use same loader so testing for both is unnecessary.

I suggest rewriting this part to use `import` instead of `extends`.

memsharded: yes, I agree, lets go with the tested ``imports`` in the docs, which also promotes the usage of "composition over inheritance", which I think it is a good thing.

ytimenkov: Done.

I decided to repeat the test, however I think more logical is to provide macros.

For example a "library of functions" is distributed via `conan config install` mechanism (an example could be OS detection) and final profiles may be composed / customized from them. Something like:

```jinja2
{% macro get_os() -%}
{{ {"Darwin": "Macos"}.get(platform.system(), platform.system()) }}
{%- endmacro %}
```

```jinja2
{% import profile_funcs.jinja2 as funcs %}
[settings]
os = {{ funcs.get_os() }}
```

memsharded (approved):

PR #2198, opened 2021-08-18
Remove packages based on revisions

uilianries:
Some people asked how to remove packages from remotes, but based only on revisions.

Related to PR: https://github.com/conan-io/conan/pull/9459


danimtb:
IMO we should address https://github.com/conan-io/conan/issues/6545 first as this is the current syntax in commands using recipe revisions

danimtb (approved):

PR #2203, opened 2021-08-28
lockfiles/configurations: Add build/host profile note (conan-io/conan #9446)

hannahwhy:
This PR adds a note to the base lockfiles page re: what was learned in https://github.com/conan-io/conan/issues/9446#issuecomment-904846681.

Specifically, this PR attempts to clarify the following:

1. In Conan 1.X, base lockfiles, if used in a separate build/host profile configuration, must also use separate build and host profiles.
2. The profiles used to create the base lockfile should match the widest possible set of requirements to maximize its applicability to all desired build/host profile combinations that will be in the generated lockfiles.

This is a first draft; comments are certainly welcome.


memsharded:
Many thanks for contributing this!

PR #2205, opened 2021-08-31
Add local dev flow instructions for conandata.yml users.

p-groarke:
I'm not sure about the formatting, let me know what's wrong.
Adds fix instructions for conandata.yml users.

### DIFF
--- a/developing_packages/package_dev_flow.rst
+++ b/developing_packages/package_dev_flow.rst
@@ -24,6 +24,15 @@ phase.
 
 We will use this `conan flow example <https://github.com/memsharded/example_conan_flow>`_ to follow the steps in the order below.
 
+If you use `conandata.yml` to target multiple package versions, running the following steps as-is will fail since no version is specified. To fix this, we can generate a lock file using our desired version. For example, with `0.1.0`.
+
+.. code-block:: bash
+
+    $ cd example_conan_flow
+    $ conan install . 0.1.0@user/channel
+
+This will allow you to run the following steps as expected.
+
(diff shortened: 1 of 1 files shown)


developing_packages/package_dev_flow.rst
```diff
@@ -24,6 +24,15 @@ phase.
 
 We will use this `conan flow example <https://github.com/memsharded/example_conan_flow>`_ to follow the steps in the order below.
 
+If you use `conandata.yml` to target multiple package versions, running the following steps as-is will fail since no version is specified. To fix this, we can generate a lock file using our desired version. For example, with `0.1.0`.
```
memsharded, line 27: I think this is a bit out of place, as the referenced and used example is not using ``conandata.yml`` at all. Also lockfiles are not involved here either.

I think ``conandata.yml`` explanations deserves its own section in https://docs.conan.io/en/latest/creating_packages.html, it is a resource to be used by recipe writers

Then, this example command can be added to the https://docs.conan.io/en/latest/developing_packages/package_dev_flow.html#conan-install section, saying something like: 

> If the recipe doesn't directly encode the version (for example, if using ``conandata.yml`` to target multiple package versions), then it is necessary to pass the version in the command line:

memsharded: (No need to add the new ``conandata.yml`` section in this PR, btw)

p-groarke:
Fair enough, we just have new users scratching their heads how to pass versions to the commands.

Have a great day!

PR #2211, opened 2021-09-02
new Creating packages getting started

memsharded:
Docs for https://github.com/conan-io/conan/pull/9516

### DIFF
--- a/creating_packages/getting_started.rst
+++ b/creating_packages/getting_started.rst
@@ -31,11 +31,12 @@ Using the :command:`conan new` command will create a "Hello World" C++ library e
     File saved: test_package/src/example.cpp
 
 
-On the root folder, there is a *conanfile.py* which is the main recipe file, responsible for
-defining our package and the *src* folder that contains the simple C++ "hello" library with a simple generic
-*CMakeLists.txt* to build it, with nothing specific about Conan in it. Also, there is an optional *test_package*
-folder, which contains an *example* application that links the that will require and link with the created package.
-It is not mandatory, but it is useful to check that our package is correctly created.
+The generated files are:
+
+- **conanfile.py**: On the root folder, there is a *conanfile.py* which is the main recipe file, responsible for defining how the package is built and consumed.
+- **src** folder: the *src* folder that contains the simple C++ "hello" library with a simple generic *CMakeLists.txt* to build it, with nothing specific about Conan in it.
+- (optional) **test_package** folder: contains an *example* application that links the that will require and link with the created package.
+  It is not mandatory, but it is useful to check that our package is correctly created.
 
 Let's have a look at the package recipe *conanfile.py*:
 
@@ -81,7 +82,7 @@ Let's have a look at the package recipe *conanfile.py*:
             self.cpp_info.libs = ["hello"]
 
 
-Without going into detail, these are the basics:
+Let's explain a little bit this recipe:
 
 - The binary configuration is composed by ``settings`` and ``options``. See more in :ref:`this section<settings_vs_options>`.
   When something changes in the configuration, the resulting binary built and packaged will be different:
--- a/creating_packages/getting_started.rst
+++ b/creating_packages/getting_started.rst
@@ -1,377 +1,221 @@
 .. _packaging_getting_started:
 
-Getting Started
+Getting started
 ===============
 
-To start learning about creating packages, we will create a package from the existing source code
-repository: https://github.com/conan-io/hello. You can check that project, it is a very simple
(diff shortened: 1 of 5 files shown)


creating_packages/getting_started.rst
```diff
@@ -1,377 +1,221 @@
...
-                                  '''PROJECT(MyHello)
-    include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-    conan_basic_setup()''')
+        options = {"shared": [True, False], "fPIC": [True, False]}
+        default_options = {"shared": False, "fPIC": True}
+
+        # Sources are located in the same place as this recipe, copy them to the recipe
+        exports_sources = "src/*"
+
+        def config_options(self):
```
lasote, line 60: Too much for starting? Maybe remove the `fPIC` option? and explain the options later?

creating_packages/getting_started.rst
```diff
@@ -1,377 +1,221 @@
...
-- The ``settings`` field defines the configuration of the different binary packages. In
-  this example, we defined that any change to the OS, compiler, architecture or build type will
-  generate a different binary package. Please note that Conan generates different binary packages for
-  different introduced configuration (in this case settings) for the same recipe.
 
-  Note that the platform on which the recipe is running and the package being built differ from
-  the final platform where the code will be running (``self.settings.os`` and ``self.settings.arch``) if
-  the package is being cross-built. So if you want to apply a different build depending on the current
-  build machine, you need to check it:
+Without going into detail, these are the basics:
```
lasote, line 84: If I read this I would say... wow, luckily it is only the basics because what a s** load of text!

lasote (approved):

creating_packages/getting_started.rst
```diff
@@ -1,377 +1,221 @@
...
+    File saved: src/hello.cpp
+    File saved: src/hello.h
+    File saved: test_package/conanfile.py
+    File saved: test_package/src/CMakeLists.txt
+    File saved: test_package/src/example.cpp
 
-This will generate the following files:
 
-.. code-block:: text
+On the root folder, there is a *conanfile.py* which is the main recipe file, responsible for
```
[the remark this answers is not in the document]
memsharded: Done

creating_packages/getting_started.rst
memsharded: This is the recipe exactly as it is created with ``conan new --template=cmake_lib``, not sure if we should hide it from the docs. I think the real feature is to automate that in Conan, so it doesn't need to be in every recipe, but at the moment, not sure what is the best approach.

creating_packages/getting_started.rst
memsharded: Think it is better to try to reduce it?

lasote: Now is better.

lasote (approved):

franramirez688 (approved):
Looks much better and simpler 👏

creating_packages/getting_started.rst
```diff
@@ -1,377 +1,222 @@
...
+.. important::
 
-    For this concrete example you will need, besides a C++ compiler, both *CMake* and *git*
-    installed and in your path. They are not required by Conan, so you could use your own build system
-    and version control instead.
+    This is a **tutorial** section. You are encouraged to execute these commands.
+    For this concrete example you will need **CMake** installed  in your path.
+    It is not strictly required by Conan to create packages, you can use
+    other build systems (as VS, Meson, Autotools and even your own) to create packages, without any dependency
+    to CMake.
```
czoido, line 16:
```suggestion
    For this concrete example, you will need **CMake** installed  in your path.
    It is not strictly required by Conan to create packages, you can use
    other build systems (as VS, Meson, Autotools and even your own) to do that, without any dependency
    to CMake.
```

creating_packages/getting_started.rst
```diff
@@ -1,377 +1,222 @@
...
+The generated files are:
 
-    conanfile.py
-    test_package
-      CMakeLists.txt
-      conanfile.py
-      example.cpp
+- **conanfile.py**: On the root folder, there is a *conanfile.py* which is the main recipe file, responsible for defining how the package is built and consumed.
+- **src** folder: the *src* folder that contains the simple C++ "hello" library with a simple generic *CMakeLists.txt* to build it, with nothing specific about Conan in it.
+- (optional) **test_package** folder: contains an *example* application that links the that will require and link with the created package.
```
czoido, line 38:
```suggestion
- (optional) **test_package** folder: contains an *example* application that will require and link with the created package.
```

PR #2223, opened 2021-09-13
Add ios-cmake toolchain docs

a4z:


lasote:
Thanks for this. I think first we need to try supporting all this into the new `CMakeToolchain` to have native support in Conan.

a4z:
Not sure what you mean,  ios-cmake is independent, you can use it basically stand alone, with out anything else, and it will just work

lasote:
Then probably shouldn't be documented in the conan docs. Or not like this, I meant we should explore a better integration with the toolchain than using a package with a build require. That's all.

a4z:
Then explain why the so called conan-darwin-toolchain is mentioned. 
It was a point of confusion for many people over the last years. They got through the conan docs to that package, instead to the ios-cmake toolchain,
I fix that here. The way to build for iOS tvOS watchOS is ios-cmake, that is proven to work for many people since years.

lasote:
Because the project evolves. Nowadays I won't add `conan-darwin-toolchain` either.

PR #2229, opened 2021-09-21
Fix typo in Conan toolchain filename

jwillikers:
`conantoolchain.cmake` is actually `conan_toolchain.cmake`.


memsharded (approved):

memsharded:
Thanks for fixing this!
Changing it to "master" branch, so it is published as soon as possible.

PR #2241, opened 2021-10-02
Add LDFLAGS variable to Address Sanitizer example

jwillikers:
It's deceptive to think that using only `CXXFLAGS` / `CFLAGS` to configure Address Sanitizer would work without linking to Address Sanitizer using `LDFLAGS` in this example. I've just added `LDFLAGS` to make the example functional.


lasote (approved):
Thanks! :)

PR #2247, opened 2021-10-05
multi-config env

memsharded:
Docs for https://github.com/conan-io/conan/pull/9543/files


lasote (changes requested):
We need to update the general environment page too with the "group" and removing the "auto_activate":

https://docs.conan.io/en/latest/reference/conanfile/tools/env/environment.html

Also VCVars:

https://docs.conan.io/en/latest/reference/conanfile/tools/microsoft.html?highlight=auto_activate

PR #2258, opened 2021-10-12
[CMakeToolchain] -- Fix constructor argument name

daravi:


memsharded:
The ``namespace`` thing was removed, this is outdated.

PR #2262, opened 2021-10-13
Remove outdated paragraph about package_revision_mode

chausner-audeering:
The docs claim that when using `package_revision_mode`, dependency graphs cannot be built automatically. However, support for this was implemented in https://github.com/conan-io/conan/pull/4958. This updates the docs by removing that whole paragraph.


danimtb:
Thanks a lot for the fix @chausner-audeering. Would you mind signing the CLA please?

chausner-audeering:
> I would also like to start advocating for the idea of removing package_revision_mode for Conan 2.0. It is a very problematic mode and makes many things challenging.

Thank you for your comment, this is very interesting for me because we have just planned to transition to `package_revision_mode` in the near future because we'v started running into compatibility problems with the default `semver_direct_mode`.

Can you point me to any documentation or discussion where the problems of `package_revision_mode` are explained? What would you suggest instead? `recipe_revision_mode`? If you prefer, I can also open an issue for discussion on the main repo.

memsharded:
> Thank you for your comment, this is very interesting for me because we have just planned to transition to package_revision_mode in the near future because we'v started running into compatibility problems with the default semver_direct_mode.

> Can you point me to any documentation or discussion where the problems of package_revision_mode are explained? What would you suggest instead? recipe_revision_mode? If you prefer, I can also open an issue for discussion on the main repo.

Yes, we know that ``semver_direct_mode`` is not a good default, and Conan 2.0 will change to a better mode by default too.

In the meantime, unless you have a patological environment in which unnecesary re-builds are being done with uncontrolled changes to the environment (compiler, etc), that want to rely on Conan to rebuild downstream (that would need a whole rebuild anyway), ``recipe_revision_mode`` is an "exact" mode, in the sense that it will not skip any change to the source code or recipe of the dependencies. Depending on the balance and tradeoffs you want to do between accuracy and speed, you might want to use that for the safest+requires more building, or if you use some good versioning approach you could try to go to ``minor_mode`` or something like that (but require more developer discipline too).


At the moment the pain is more about issues, bug reports, and coding than an open discussion in a explicit Github ticket. We will open it as a Tribe discussion soon I think.

PR #2264, opened 2021-10-14
Minor clarifications to global.conf page

chausner-audeering:
These changes aim to reduce some potential confusion about the global.conf documentation (see https://github.com/conan-io/conan/issues/9529#issuecomment-940884107). It adds an example `global.conf` file and tries to make it more clear that the snippet shown under "Tools configurations" is profile syntax, not global.conf syntax.


prince-chrismc:
@czoido Is this a good spot for adding an example?

memsharded (approved):

PR #2267, opened 2021-10-20
Generate conan 2.0-alpha docs

czoido:
This will generate the docs for `conan 2.0-alpha` from the `release/2.0.0-alpha` branch.
We have to improve ci pipelines now that we want to generate 2.0 and 1.0 docs at the same time, they are not prepared for doing that easily.

### DIFF
--- a/.ci/publish.jenkins
+++ b/.ci/publish.jenkins
@@ -41,7 +41,8 @@ Map<String, String> versions = [
         'release/1.6.1': '1.6',
         'release/1.5.2': '1.5',
         'release/1.4.5': '1.4',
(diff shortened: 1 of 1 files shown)


czoido:
Yes, this should not show [release/2.0.0-alpha](https://github.com/conan-io/docs/tree/release/2.0.0-alpha) as latest. Just show it in the versions list at the bottom of the page. I think it's based in the order on the list, not ordered by version, so version 2.0-alpha should appear after 1.3 version (I did this on purpose to be less visible while we add documentation to it as this is currently empty, but can be changed at any time)

PR #2268, opened 2021-10-21
- move cmake_find_package_multi generator to the top

SSE4:
for newbies, it's not obvious what should be the reasonable default choice for the cmake generator amongst others:
- cmake
- cmake_multi
- cmake_paths
- cmake_find_package
- cmake_find_package_multi
- CMakeDeps

as we expect (most of) readers to read the documentation from the top to the bottom, IMO it's reasonable to place a default choice to the top rather to the very bottom.

### DIFF
--- a/integrations/build_system/cmake.rst
+++ b/integrations/build_system/cmake.rst
@@ -15,11 +15,11 @@ Conan can be integrated with CMake using different generators, build helpers and
 .. toctree::
    :maxdepth: 2
 
+   cmake/cmake_find_package_multi_generator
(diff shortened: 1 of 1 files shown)


Croydon:
Outdated and I think this can be closed.

1.x docs already have a huge warning, referring to the new CMakeDeps and CMakeToolchain generators

https://docs.conan.io/1/integrations/build_system/cmake.html

memsharded:
Thanks all, and thanks @Croydon for raising it.

PR #2269, opened 2021-10-21
- move cmake_find_package_multi generator to the top

SSE4:
for newbies, it's not obvious what should be the reasonable default choice for the cmake generator amongst others:
- cmake
- cmake_multi
- cmake_paths
- cmake_find_package
- cmake_find_package_multi
- CMakeDeps

as we expect (most of) readers to read the documentation from the top to the bottom, IMO it's reasonable to place a default choice to the top rather to the very bottom.

### DIFF
--- a/reference/generators.rst
+++ b/reference/generators.rst
@@ -18,11 +18,11 @@ Available generators:
 .. toctree::
    :maxdepth: 1
 
+   generators/cmake_find_package_multi
    generators/cmake
    generators/cmakemulti
    generators/cmake_paths
    generators/cmake_find_package
-   generators/cmake_find_package_multi
    generators/msbuild
    generators/visualstudio
    generators/visualstudiomulti


czoido:
Yes, I agree with marking the current state and the plans for the future. In fact, in my opinion, I would only recommend _CMakeDeps_ that is going to be default for the examples in the documentation...

prince-chrismc:
I think I can finally close this with https://github.com/conan-io/docs/pull/2878 as it's will have updated the language around both the old and new generators 🤞

Croydon:
> so, to clarify, it should look like:
> 
> ```
> Recommended generators:
>  * CMakeDeps (default in Conan v2)
> 
> Not recommended (already deprecated):
>  * cmake (to be deprecated in v2)
>  * cmake_find_package_multi (to be deprecated in v2)
>  * cmake_find_package (to be deprecated in v2)
>  * cmake_paths (to be deprecated in v2)
>  * cmake_multi (to be deprecated in v2)
> ```


I think this is still the best idea as it gives you a really quick overview

PR #2284, opened 2021-10-31
docs: add missing word

bhumitattarde:


memsharded:
Thanks (this change will be published together with next release)

PR #2285, opened 2021-11-02
POC of integrations with a single page using include

lasote:
Might be this what we were looking for?

### DIFF
--- a/integrations/build_systems/visual_studio.rst
+++ b/integrations/build_systems/visual_studio.rst
@@ -1,0 +1,1 @@
+.. include:: /single_pages/visual_studio.rst
--- a/integrations.rst
+++ b/integrations.rst
@@ -3,3 +3,12 @@
 
 Integrations
 ============
+
+
+.. toctree::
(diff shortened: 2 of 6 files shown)


memsharded:
Branch should be named release/2.0.0-alpha.1 better? @czoido ?

integrations/build_systems/visual_studio.rst
```diff
@@ -0,0 +1 @@
+.. include:: /single_pages/visual_studio.rst
```
memsharded, line 1: I would say it should be a redirect, not an include (different URLs)

czoido:
> Branch should be named release/2.0.0-alpha.1 better? @czoido ? Or the docs will be continuous for all alphas?

Maybe we could use the branch. `release/2.0.0-alpha` as if it was something similar to our current `develop` branch, and branch `release/2.0.0-alpha1` from there when we have merged all the docs in `release/2.0.0-alpha`. The, while developing, people don't have to take into account to which alpha version branch they have to point their PR.

PR #2286, opened 2021-11-03
Conan2/creating packages getting started

tapia:
Should we add a brief explanation here on what the package ID and revision ID are?


creating_packages.rst
```diff
@@ -0,0 +1,11 @@
+.. _packaging:
+
+Creating Packages
```
memsharded, line 3: This should be already in the "tutorial" folder and section

memsharded (changes requested):

creating_packages.rst
tapia: Done

PR #2288, opened 2021-11-04
Conan2/creating packages getting started

tapia:
Should we add a brief explanation here on what the package ID and revision ID are?


tutorial/creating_packages/getting_started.rst
```diff
@@ -0,0 +1,243 @@
...
+          compiler.version=12.0
+          os=Macos
+        options:
+          fPIC=True
+          shared=True
+      hello/0.1@demo/testing#afa4685e137e7d13f2b9845987c5af77:a5c01fc21d2db712d56189dff69fc10f12b22375
+        settings:
+          arch=x86_64
+          build_type=Debug
+          compiler=apple-clang
```
memsharded, line 223: This was on purposely hidden in the current Getting started, to be as generic as possible, leaving only the "differences", build_type and shared

memsharded (approved):

tutorial/creating_packages/getting_started.rst
tapia: I think it's ok to show the full output. Nothing too confusing there.

PR #2289, opened 2021-11-04
Add docu for get_filename

blackliner:
from here: https://github.com/conan-io/conan/blob/f230bf104842c160be6ea377497151fe1ea64c38/conans/model/build_info.py#L215


memsharded (commented):
This interface is already considered legacy, and superseded by the ``self.dependencies["dep"].cpp_info``, that doesn't have the ``get_filename`` anymore, as we are already moving to the new ``set_property()`` mechanism.

If it hasn't been documented so far, I prefer not to document it now, we don't want people to start using it now

blackliner:
Ok, no worries, I just found it and need it to distinct between
- find_package calls
- package names
- global target name

memsharded:
Thanks!

PR #2296, opened 2021-11-11
- [meson] add note about cross-building

SSE4:
for https://github.com/conan-io/conan/issues/9922

### DIFF
--- a/reference/conanfile/tools/meson.rst
+++ b/reference/conanfile/tools/meson.rst
@@ -62,6 +62,15 @@ etc. This includes but is not limited to the following:
 but with additional information to describe host, target, and build machines (such
 as the processor architecture).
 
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/meson.rst
```diff
@@ -62,6 +62,15 @@ etc. This includes but is not limited to the following:
 but with additional information to describe host, target, and build machines (such
 as the processor architecture).
 
+**NOTE**: unlike CMake, meson cannot automatically deduce the compiler executable in case of cross-building, so it has to be
+explicitly specified within conan profile, for instance, the following might be required in case of Visual Studio:
+
+.. code-block:: text
+
+	[env]
```
lasote, line 70: @memsharded I'm wondering if we should keep using the `env` but I can't find docs for `buildenv` and `runenv`

PR #2306, opened 2021-11-29
Conan2/installation

tapia:


memsharded (approved):

memsharded:
Maybe better to target it to ``release/2.0.0-alpha1``, so current released docs are fixed.

PR #2313, opened 2021-12-02
test_requires and some general 2.0 info

memsharded:
Docs for https://github.com/conan-io/conan/pull/10027

### DIFF
--- a/conan_v2.rst
+++ b/conan_v2.rst
@@ -18,14 +18,14 @@ This section summarizes some of the necessary changes during Conan 1.X to be rea
 - Use generators and helpers only from ``conan.tools.xxxx`` space. All the other ones are going to be removed.
 - Use always build and host profiles. You can enable it by passing ``-pr:b=default`` in the command line to most commands.
   Do not use ``os_build``, ``arch_build`` anywhere in your recipes or code.
-- Use ``self.test_requires()`` to define test requirements instead of the legacy ``self.build_requires(..., force_host_context).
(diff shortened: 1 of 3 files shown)


conan_v2.rst
```diff
@@ -9,10 +9,24 @@ behaviors, clean the codebase and add space for new developments. Development is
...
-there is a lot of work ahead, as you can see in `our backlog <https://github.com/conan-io/conan/milestones>`_.
+Conan 2.0-alpha is already released. You can access its documentation with the right version label.
+It can be installed from PyPI with ``pip install conan==2.0.0-alpha1``
 
-Meanwhile, in version 1.23 we have introduced an environment variable to raise errors in case of using features
-that will be deprecated in Conan 2.0. Read more about ``CONAN_V2_MODE`` in :ref:`this section <conan_v2_mode>`.
+This section summarizes some of the necessary changes during Conan 1.X to be ready to upgrade to Conan 2.0:
+
+
+- Use generators and helpers only from ``conan.tools.xxxx`` space. All the other ones are going to be removed.
```
czoido, line 18: I guess that this list is enough to keep moving, but I think that we must add practical examples of what all these things means in terms of changes in your recipes, how to activate configurations in 2.0 with examples of the Conan commands, and so on... This section is very important so people can understand what migration to 2.0 in practical terms means

czoido (approved):

lasote (approved):

conan_v2.rst
memsharded: Yes, totally agree. I wanted to at least add some notes as a guideline of things that should be explained, but indeed there is a lot to be added there.

Issue #2321, opened 2021-12-06
New 'Conan migration guide to 2.0' has outdated properties info


eigenwhat:
https://docs.conan.io/en/1.43/conan_v2.html#new-properties-model-for-the-cpp-info-in-conan-2-0-generators

Part of the documentation still refers to the new properties affecting legacy generators.

> New properties defined for CMake generators family, used by CMakeDeps generator:
>
> * `cmake_file_name` property will affect all cmake generators that accept the `filenames` property _(cmake_find_package and cmake_find_package_multi)._
> * `cmake_target_name` property will affect all cmake generators that accept the `names` property _(cmake, cmake_multi, cmake_find_package, cmake_find_package_multi and cmake_paths)._

memsharded:
You are right, lets fix this asap, thanks very much for reporting.

franramirez688:
Thanks for reporting it @eigenwhat 
The fix https://github.com/conan-io/docs/pull/2322 is already merged.

PR #2322, opened 2021-12-06
fix properties description in v2 migrate guide

memsharded:
Fix https://github.com/conan-io/docs/issues/2321

### DIFF
--- a/conan_v2.rst
+++ b/conan_v2.rst
@@ -99,8 +99,8 @@ They have a new way of setting this information using ``set_property`` and
 
 .. code-block:: python
 
-    def set_property(self, property_name, value, generator=None)
-    def get_property(self, property_name, generator=None):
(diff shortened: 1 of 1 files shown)


franramirez688:
@memsharded 

Taking advantage of this PR I just realized another property should be changed. A few lines below your changes, you can find out:

```
Properties related to *pkg_config*, used by legacy ``pkg_config`` and new :ref:`PkgConfigDeps<PkgConfigDeps>`:
.......
- **pkg_config_aliases** property sets some aliases of any package/component name for *pkg_config* generator. This property only accepts list-like Python objects.
```

But ``pkg_config_aliases`` is not available in the legacy one. Could you please add something like ``(only available in PkgConfigDeps)`` or similar?

franramirez688 (approved):

PR #2324, opened 2021-12-08
Provide short git.checkout_submodules doc line

Bedzior:
Hi!

I realised there's a missing line of documentation for a neat function in the `Git` helper tool.

Basically, I was looking for the most straightforward way to checkout submodules after a shallow clone. Luckily enough my IDE pointed me to `checkout_submodules`, which does the submodule checkout only, without duplicating checkout for root repository (as `checkout` would).

I would be happy to see this accepted, unless I am doing something horribly wrong, in which case I am open to criticism.

R.


Bedzior:
If any other part of the documentation could see an update to showcase shallow cloning to a wider audience, please let me know!

memsharded:
Sorry this was never reviewed or managed.
Unfortunately, this ``Git`` tool has been superseded by ``from conan.tools.scm import Git`` new one, so these docs are no longer necessary. Thanks!

PR #2342, opened 2021-12-28
Renamed build_requires to tool_requires

lasote:
Docs for https://github.com/conan-io/conan/pull/10168/files

### DIFF
--- a/cheatsheet.rst
+++ b/cheatsheet.rst
@@ -418,7 +418,7 @@ A package recipe is a Python class, defined in a file called conanfile.py:
             ...                                                          # tasks, and using them is often preferable to using self.run()
                                                                          # See the link below for more information
 
-        def tool_requirements(self):                                    # Responsible for specifying non-trivial build requirements logic
+        def build_requirements(self):                                    # Responsible for specifying non-trivial build requirements logic
             if self.options.myoption1:                                   # Specify a conditional tool requirement
                 self.tool_requires("zlib/1.2@user/testing")
 
--- a/cheatsheet.rst
+++ b/cheatsheet.rst
@@ -97,7 +97,7 @@ Profiles
 ++++++++
 
 Profiles allow users to set aspects of the build configuration. This includes settings_, options_, environment variables
-and build requirements. They can be installed into ~/.conan/profiles. They can also be stored in project directories,
+and tool requirements. They can be installed into ~/.conan/profiles. They can also be stored in project directories,
 which can be useful for specific compilation cases, for example cross-compiling.
 
 Profiles are stored in text files with no file extension. An example profile:
@@ -126,8 +126,8 @@ Profiles are stored in text files with no file extension. An example profile:
     CC=$CROSS_GCC-gcc             # Strings can be defined and substituted
     CXX=$CROSS_GCC-g++
 
-    [build_requires]              # Requirements for package builds only
-    cmake/3.16.3                  # Specifying build requirements here rather than in the recipe makes them less binding
+    [tool_requires]               # Requirements for package builds only
+    cmake/3.16.3                  # Specifying tool requirements here rather than in the recipe makes them less binding
 
 List profiles:
 
@@ -185,7 +185,7 @@ Using packages in an application
     boost/1.72.0                   # Versions override versions upstream in the dependency graph
     poco/1.9.4
 
-    [build_requires]               # The Conan packages which are used to build the application
+    [tool_requires]                # The Conan packages which are used to build the application
     7zip/16.00
 
     [generators]                   # Generators create build system files that capture the dependency information,
@@ -406,7 +406,7 @@ A package recipe is a Python class, defined in a file called conanfile.py:
 
         default_options = {"shared": False}
         requires = "requiredlib/0.1@user/stable"           # Defines package requirements
-        build_requires = "tool_a/0.2@user/testing"         # Defines requirements that are only used when the package is
+        tool_requires = "tool_a/0.2@user/testing"          # Defines requirements that are only used when the package is
                                                            # built. These should be build and test tools only
         generators = "cmake"                               # Generator for the package: specifies which build system
                                                            # type will be generated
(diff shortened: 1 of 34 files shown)


reference/conanfile/attributes.rst
```diff
@@ -643,19 +643,19 @@ Expressions are those defined and implemented by [python node-semver](https://py
 
     Go to :ref:`Mastering/Version Ranges<version_ranges>` if you want to learn more about version ranges.
 
-build_requires
+tool_requires
 --------------
 
-Build requirements are requirements that are only installed and used when the package is built from sources. If there is an existing pre-compiled binary, then the build requirements for this package will not be retrieved.
+Build requirements are requirements that are only installed and used when the package is built from sources. If there is an existing pre-compiled binary, then the tool requirements for this package will not be retrieved.
```
czoido, line 649:
```suggestion
Tool requirements are requirements that are only installed and used when the package is built from sources. If there is an existing pre-compiled binary, then the tool requirements for this package will not be retrieved.
```

cheatsheet.rst
```diff
@@ -418,9 +418,9 @@ A package recipe is a Python class, defined in a file called conanfile.py:
             ...                                                          # tasks, and using them is often preferable to using self.run()
                                                                          # See the link below for more information
 
-        def build_requirements(self):                                    # Responsible for specifying non-trivial build requirements logic
-            if self.options.myoption1:                                   # Specify a conditional build requirement
-                self.build_requires("zlib/1.2@user/testing")
+        def tool_requirements(self):                                    # Responsible for specifying non-trivial build requirements logic
```
czoido, line 421: Does the `tool_requirements` method exist?

devtools/build_requires.rst
```diff
@@ -16,72 +16,72 @@ In short:
...
-- ``&`` means apply these build requirements to the consumer conanfile
-- ``&!`` means apply the build requirements to all packages except the consumer one.
+- ``&`` means apply these tool requirements to the consumer conanfile
+- ``&!`` means apply the tool requirements to all packages except the consumer one.
 
 Remember that the consumer conanfile is the one inside the *test_package* folder or the one referenced in the :command:`conan install`
 command.
 
-Build requirements can also be specified in a package recipe, with the ``build_requires`` attribute and the ``build_requirements()`` method:
+Tool requirements can also be specified in a package recipe, with the ``tool_requires`` attribute and the ``tool_requirements()`` method:
```
czoido, line 64: Same comment above about the `tool_requirements` method

devtools/build_requires.rst
```diff
@@ -175,18 +175,18 @@ The behavior of ``build_requires`` is the same irrespective if they are defined
...
 for both of them.
 
-Build requirements can be checked for existence (whether they've been applied) in the recipes, which can be useful for conditional logic in
+Tool requirements can be checked for existence (whether they've been applied) in the recipes, which can be useful for conditional logic in
 the recipes. In this example, we could have one recipe with the following ``build()`` method:
 
 .. code-block:: python
 
-    def build_requirements(self):
+    def tool_requirements(self):
```
czoido, line 187: Same comment above about the `tool_requirements` method

devtools/build_requires.rst
```diff
@@ -111,10 +111,10 @@ be linked to the generated library or other executable we want to deploy to the
 .. code-block:: python
 
     class MyPkg(ConanFile):
-        build_requires = "nasm/2.14"  # 'build' context (nasm.exe will be available)
+        tool_requires = "nasm/2.14"  # 'build' context (nasm.exe will be available)
 
-        def build_requirements(self):
-            self.build_requires("protobuf/3.6.1")  # 'build' context (protoc.exe will be available)
+        def tool_requirements(self):
```
czoido, line 116: Same comment above about the `tool_requirements` method

devtools/build_requires.rst
```diff
@@ -16,72 +16,72 @@ In short:
...
-                self.build_requires("tool_win/0.1@user/stable")
+                self.tool_requires("tool_win/0.1@user/stable")
 
 The above ``tool_a`` and ``tool_b`` will always be retrieved and used for building this recipe, while the ``tool_win`` one will only be used
 only in Windows.
 
-If any build requirement defined inside ``build_requirements()`` has the same package name as the one defined in the ``build_requires``
-attribute, the one inside the ``build_requirements()`` method will prevail.
+If any tool requirement defined inside ``tool_requirements()`` has the same package name as the one defined in the ``tool_requires``
+attribute, the one inside the ``tool_requirements()`` method will prevail.
```
czoido, line 81: Same comment above about the `tool_requirements` method

devtools/build_requires.rst
```diff
@@ -1,6 +1,6 @@
 .. _build_requires:
 
-Build requirements
+Tool requirements
```
czoido, line 3: I'm a bit concerned that users suddenly are not going to find anything about build requires because they have been aliased to tool_requires and we want them to use the new name thinking in 2.0, but I think that we should add a note in certain places like here explaining about the new name, and that now they should start naming their build_requires as tool_requires, if not this may be confusing... WDYT?

cheatsheet.rst
memsharded: No it doesn't

devtools/build_requires.rst
memsharded: I wouldn't like all users just replacing "build_requires" with "tool_requires" without really understanding that the scope is just for tools, and not for things like libraries (that they want to hide transitivity). I also see the other point, that if we are not pushing for this new naming, it won't happen because users will not realize.

memsharded (changes requested):
Seems the ``tool_requirements()`` method needs to be fixed.

devtools/build_requires.rst
lasote: Well, the explanation of "this is only for tools" is everywhere, I'm not sure we can do much more. For me, if you read the documentation, you see "tool require" and you read "this is only for tools"..., this is an open-and-shut case.

devtools/build_requires.rst
```diff
@@ -1,8 +1,14 @@
...
 ==================
 
+.. important::
+
+   The test requirement was formerly named "build requirement" and has been renamed to highlight that the usage of
```
czoido, line 8:
```suggestion
   The tool requirement was formerly named "build requirement" and has been renamed to highlight that the usage of
```

systems_cross_building/cross_building.rst
```diff
@@ -128,8 +128,8 @@ You can find working examples at the :ref:`bottom of this section <cross_buildin
 
 .. _cross_building_build_requires:
 
-Using build requires
-++++++++++++++++++++
+Using tool requires
```
czoido, line 131: Maybe I would also add the note saying that tool requires used to be build requires here...

lasote: done

czoido (approved):

memsharded (approved):

PR #2349, opened 2021-12-30
better 2 profiles guide

lasote:


conan_v2.rst
```diff
@@ -71,9 +71,14 @@ available for Conan 2.0.
 Host and build profiles and new cross-building model
 ----------------------------------------------------
 
-Use always build and host profiles. You can enable it by passing ``-pr:b=default`` in the
-command line to most commands. Do not use ``os_build``, ``arch_build`` anywhere in your
-recipes or code.
+Use always build and host profiles.
```
czoido, line 74: maybe we can add a link to: https://docs.conan.io/en/latest/reference/profiles.html#build-profiles-and-host-profiles ?

lasote: sure!

memsharded (approved):

conan_v2.rst
lasote: oh, you meant in that paragraph, wait 1 min

conan_v2.rst
```diff
@@ -71,9 +71,14 @@ available for Conan 2.0.
...
+You can enable it:
```
czoido, line 75: Maybe we can add extend the sentence a bit? Like: _Conan 1.x will use one profile by default, to start using two profiles, please do the following_ or something like that?

lasote: sure

czoido (approved):

PR #2353, opened 2022-01-11
Document microsoft.is_msvc

uilianries:
Related to https://github.com/conan-io/conan/pull/10310

Screenshot 2022-01-11 at 13-25-21 conan tools microsoft — conan 1 44 0 documentation

### DIFF
--- a/reference/conanfile/tools/microsoft.rst
+++ b/reference/conanfile/tools/microsoft.rst
@@ -321,7 +321,7 @@ conan.tools.microsoft.is_msvc()
 
 Validate ``self.settings.compiler`` for which compiler is being used.
 It returns ``True`` when the host compiler is ``Visual Studio`` or ``msvc``, otherwise, returns ``False``.
-When the ``compiler`` is empty, it raises a ``ConanException``.
+When the ``compiler`` is empty, it returns ``False``.
 
 Parameters:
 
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/microsoft.rst
```diff
@@ -310,3 +310,28 @@ Parameters:
...
+conan.tools.microsoft.is_msvc()
+-------------------------------
+
+.. code-block:: python
+
+    def is_msvc(conanfile):
+
+Validate ``self.settings.compiler`` for which compiler is being used.
+It returns ``True`` when the host compiler is ``Visual Studio`` or ``msvc``, otherwise, returns ``False``.
+When the ``compiler`` is empty, it raises a ``ConanException``.
```
SSE4, line 324: why not just False?

memsharded: Yes, I wouldn't mind if it doesn't raise, and just return False if compiler not defined.

uilianries: It's more like, if you pass ConanFile to is_msvc, but you don't have compiler, it's a violation. When returning False, it's saying, "That's not MSVC", but your compiler is valid.

I understand your point, avoid more exception points, so I'll remove it.

PR #2361, opened 2022-01-17
Fix name of include directories sections

Tomeamis:
The docs specify that the name of the section for include directries is include_dirs[_XXX], but in fact it is includedirs[_XXX]


memsharded:
You are right, thanks for fixing this!

memsharded:
(this fix will be released together with next release)

PR #2370, opened 2022-01-24
Document test_type

### LINKED ISSUE #2367: test_type is not well docummented
from https://github.com/conan-io/conan-center-index/pull/8947#discussion_r789910861

the search results for `test_type` are only 4 pages, where it is briefly mentioned, but there is no reference page for the attribute

possible values (from code) seems to be: None, `explicit`, `requires`, `build_requires` (and seems like it can be None-type, string or list of strings, as I can deduce).

uilianries:
The attribute is not documented, only some example cases.

I didn't understand very well the injection, because when declares only `build_requires`, it solves requires too. For instance, you create a regular lib, and set `test_type = "build_requires"`, the test_package will consume the library and link to test package application.

Documentation based on the code:

https://github.com/conan-io/conan/blob/38c0a51b5709db2bfb0e713bbc223da63287e913/conans/client/graph/graph_manager.py#L256

closes #2367


reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+test_type
+---------
+
+.. warning::
+
+    Components is a **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
```
SSE4, line 1685: I think it should be explicitly stated that it might be a list (comma separated), e.g. the following should be a valid example:
```
test_type = 'requires', 'build_requires', `explicit`
``

File: reference/conanfile/attributes.rst [hunk 2]
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
         win_bash = True
+
+.. _test_type:
+
+test_type
+---------
+
+.. warning::
+
+    Components is a **experimental** feature subject to breaking changes in future releases.
```
SSE4, line 1681:
```suggestion
    Test type is an **experimental** feature subject to breaking changes in future releases.
```

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+It will become the standard behavior for 2.0.
+The possible values are:
+
+ - ``requires`` (by default): It will consume all requirements automatically as usual.
```
SSE4, line 1689: I think it will not consume **all** requirements, but it means it only adds tested package as a requirement
in other words, it adds implicit `requires = <tested package>`

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+ - ``build_requires``: It will consume all build requirements automatically. It can be combined with ``requires``.
```
SSE4, line 1690: same, not "all build requirements", but only package being tested

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+ - ``explicit``: The test package will not solve its dependencies, need to declare explicitly with ``self.tested_reference_str``. This will be the default behavior for Conan 2.0
```
SSE4, line 1691: seems like `tested_reference_str` also misses a reference documentation 🤔

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,24 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+---------
+
+.. warning::
+
+    Components is a **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+It will become the standard behavior for 2.0.
```
SSE4, line 1686: `it` - what exactly? I think it needs better wording here

SSE4:
> I didn't understand very well the injection, because when declares only `build_requires`, it solves requires too. For instance, you create a regular lib, and set `test_type = "build_requires"`, the test_package will consume the library and link to test package application.


might be a bug?

reference/conanfile/attributes.rst
uilianries: Copy/Paste problem :sweat_smile:

reference/conanfile/attributes.rst
uilianries: I was afraid of it :disappointed: let's document it too.

uilianries:
> might be a bug?

Let me try to explain, if I run:

    conan new -s -t foo/0.1

And, in the `test_package/conanfile.py` I add `test_type = "build_requires"` it has same effect as using `requires` but only when using single profile approach. When using build context (dual profile), `build_requires` fails .

reference/conanfile/attributes.rst
memsharded: No, only ``explicit`` should be documented from now on.

memsharded (changes requested):
Please, do not document the things that haven't been documented yet and are to be retired.
From now on, only ``test_type = "explicit"`` should be recommended, and using the ``self.tested_reference_str``

uilianries:
> Please, do not document the things that haven't been documented yet and are to be retired. From now on, only `test_type = "explicit"` should be recommended, and using the `self.tested_reference_str`

And how people which are migrating to Conan 2.0 can learn about it? The current documentation is vague and only few experts are able to dig the source code.

memsharded:
> And how people which are migrating to Conan 2.0 can learn about it? The current documentation is vague and only few experts are able to dig the source code.

Document it properly in the conan_v2.rst migration guide, for those that were using it. For new users, only the latest way to do it with ``test_type = "explicit"`` is what should be presented as "attributes" in the reference.

uilianries:
>  Document it properly in the conan_v2.rst migration guide, for those that were using it. For new users, only the latest way to do it with test_type = "explicit" is what should be presented as "attributes" in the reference.

Understood.

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,66 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+In Conan 2.0 the ``test_type`` attribute will be ignored, the behavior will be always explicit, so declaring ``test_type="explicit"`` will make the test recipe compatible with Conan 2.0.
+The possible values are:
+
+ - ``requires`` (by default): It will consume the tested requirement automatically as usual.
+ - ``build_requires``: It will consume the test build requirement automatically. It can be combined with ``requires``.
+ - ``explicit``: The test package will not solve its dependencies, need to declare explicitly with ``self.tested_reference_str``. This will be the default behavior for Conan 2.0. The ``explicit`` can not be combined with ``requires`` neither ``build_requires``, it doesn't generate an error, but only ``explicit`` will be computed.
```
lasote, line 1691:
```suggestion
 - ``explicit``: The test package will not solve its dependencies automatically, you need to declare it explicitly using the reference at ``self.tested_reference_str``. This will be the only behavior for Conan 2.0. The additional values ``requires`` and ``build_requires`` (if specified) will be ignored.
```

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,66 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+    Test type is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+In Conan 2.0 the ``test_type`` attribute will be ignored, the behavior will be always explicit, so declaring ``test_type="explicit"`` will make the test recipe compatible with Conan 2.0.
+The possible values are:
+
+ - ``requires`` (by default): It will consume the tested requirement automatically as usual.
+ - ``build_requires``: It will consume the test build requirement automatically. It can be combined with ``requires``.
```
lasote, line 1690:
```suggestion
 - ``build_requires``: The package being tested will be consumed as a build requirement automatically. It can be combined with ``requires`` to be required in both ways.
```

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,66 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+
+    Test type is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+In Conan 2.0 the ``test_type`` attribute will be ignored, the behavior will be always explicit, so declaring ``test_type="explicit"`` will make the test recipe compatible with Conan 2.0.
+The possible values are:
+
+ - ``requires`` (by default): It will consume the tested requirement automatically as usual.
```
lasote, line 1689:
```suggestion
 - ``requires`` (default): The package being tested will be consumed as a regular requirement automatically.
```

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,66 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+ .. code-block:: python
+
+    from conans import ConanFile, CMake
+    import os
+
+    class TestPackage(ConanFile):
+        test_type = "explicit"
+
+        def requirements(self):
+            self.requires(self.tested_reference_str)
```
lasote, line 1723:
```suggestion
            self.requires(self.tested_reference_str)
            # and, or
            self.build_requires(self.tested_reference_str)
```

lasote (changes requested):

reference/conanfile/attributes.rst
```diff
@@ -1670,3 +1670,68 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+
+    Test type is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.44.0 <https://github.com/conan-io/conan/releases/tag/1.44.0>`_
+
+This attribute allows testing requirements and build requiments explicitly on test package.
+In Conan 2.0 the ``test_type`` attribute will be ignored, the behavior will be always explicit, so declaring ``test_type="explicit"`` will make the test recipe compatible with Conan 2.0.
+The possible values are:
+
+ - ``requires`` (default): The package being tested will be consumed as a regular requirement automatically.
```
memsharded, line 1689: I still don't see the point of documenting this now. Why not documenting only the ``explicit`` approach? We don't want anyone to use ``test_type = "build_requires"`` anymore, it will still require yet another migration. 

Lets omit it.

memsharded (changes requested):

reference/conanfile/attributes.rst
SSE4: the possible reason to document it is that once someone encounters `test_type = "build_requires"` in the recipe, he can go to the documentation and read what does it mean. otherwise, the only choice is to read conan's source code to understand how does recipe work.
e.g. in CCI we now have:
```
grep -RnI test_type .
./recipes/libtool/all/test_package/conanfile.py:11:    test_type = "build_requires", "requires"
./recipes/autoconf-archive/all/test_package/conanfile.py:12:    test_type = "build_requires"
./recipes/openjdk/all/test_package/conanfile.py:9:    test_type = "build_requires"
./recipes/pkgconf/all/test_package/conanfile.py:10:    test_type = "build_requires", "requires"
./recipes/gettext/all/test_package/conanfile.py:10:    test_type = "build_requires", "requires"
./recipes/android-ndk/all/test_package/conanfile.py:7:    test_type = "build_requires"
./recipes/gtk-doc-stub/all/test_package/conanfile.py:9:    test_type = "requires", "build_requires"
./recipes/emsdk/all/test_package/conanfile.py:9:    test_type = "build_requires"
./recipes/autoconf/all/test_package/conanfile.py:12:    test_type = "build_requires"
./recipes/flex/all/test_package/conanfile.py:9:    test_type = "build_requires", "requires"
./recipes/automake/all/test_package/conanfile.py:13:    test_type = "build_requires", "requires"
```
from time to time, people ask what does it mean, but we cannot provide any page describing what's the meaning of this line.
if it's something that shouldn't be documented, then we will have to remove it from all the existing recipes (probably, adding a hook restricting it and allowing only `test_type = "explicit"`

memsharded: Ok then: But lets keep an eye, and remove it as soon as possible.

PR #2377, opened 2022-01-31
doc --output-folder

memsharded:
Docs for https://github.com/conan-io/conan/pull/10274

### DIFF
--- a/reference/commands/consumer/install.rst
+++ b/reference/commands/consumer/install.rst
@@ -6,20 +6,13 @@ conan install
 
 .. code-block:: bash
 
-    $ conan install [-h] [-g GENERATOR] [-if INSTALL_FOLDER] [-m [MANIFESTS]]
-                     [-mi [MANIFESTS_INTERACTIVE]] [-v [VERIFY]]
-                     [--no-imports] [--build-require] [-j JSON] [-b [BUILD]]
-                     [-r REMOTE] [-u] [-l LOCKFILE]
-                     [--lockfile-out LOCKFILE_OUT] [-e ENV_HOST]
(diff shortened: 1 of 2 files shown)


reference/commands/development/editable.rst
```diff
@@ -64,6 +68,12 @@ the file ``win_layout``.
     $ conan editable add . cool/version@user/dev --layout=win_layout
 
 
+The ``--output-folder`` and ``--source-folder`` will specify the location of the output (build files, generators, etc)
+and the source folders, with the same meaning as the ``conan install`` command.
+By default if not specified, they wil point to the folder containing the ``conanfile.py`` These arguments work
```
czoido, line 73:
```suggestion
By default, if not specified, they will point to the folder containing the ``conanfile.py``. These arguments work
```

reference/commands/consumer/install.rst
```diff
@@ -333,6 +330,22 @@ because the consumer conanfile might not declare a `name` so it would be impossi
...
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+
+The ``--output-folder`` and ``--source-folder`` define together with the ``layout()`` recipe
+method the location of the output files. For example, the files created by build system integrations
+such as ``CMakeToolchain`` or ``PkgConfigDeps`` will be created in the folder defined by the
+``layout()`` ``generators`` folder, inside the defined ``--output-folder``. By default, both the
+``--output-folder`` and the ``--source-folder`` are the folder containing the ``conanfile.py`` or
+the current folder otherwise.
```
czoido, line 346: Maybe we can add an example or something? I think this part is a bit hard to understand

lasote (approved):
I agree with @czoido, you cannot understand these arguments without an example, of a) installing something in a different base folder or b) building something locally and use an editable package.

PR #2388, opened 2022-02-03
Send slack alert when broken links

czoido:
image


danimtb (approved):
Much better! Great job there 🍻

PR #2390, opened 2022-02-03
Update settings.yml.rst

janroergaardhansen:
Indentation was wrong for msvc compiler


memsharded:
Thanks for fixing it!

PR #2391, opened 2022-02-04
Remove warning about dependencies bug

### LINKED ISSUE #2384: dependencies - are these bugs already addressed?
the following [page](https://docs.conan.io/en/latest/reference/conanfile/dependencies.html) about dependencies says:
```
These tools are very experimental and subject to breaking changes.
It also contains some known bugs regarding tool_requires, to be addressed in next Conan 1.39
```

it's already conan 1.44, are these bugs already addressed or still there?
maybe it's time to update the notice?

uilianries:
closes #2384

### DIFF
--- a/reference/conanfile/dependencies.rst
+++ b/reference/conanfile/dependencies.rst
@@ -5,6 +5,9 @@ Dependencies
 
 Introduced in Conan 1.38.
 
+.. warning::
+
+    These tools are **experimental** and subject to breaking changes.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


SSE4 (approved):

memsharded (changes requested):
I think we want to keep the (normal) ``experimental`` note. It is not that we are declaring ``self.dependencies`` stable. We can remove the "very" and the "some bugs"

uilianries:
Done, new preview:

Screenshot 2022-02-22 at 10-05-02 Dependencies — conan 1 45 0 documentation

memsharded (approved):

PR #2410, opened 2022-02-22
Explore adding tab support for code examples...

czoido:
For windows/linux, macos examples, with the current css looks like this...

image


memsharded (changes requested):
Good for me, but I would bring this ONLY for ``develop2``, not to master.

czoido:
> Good for me, but I would bring this ONLY for `develop2`, not to master.

But can we update the requirements to support this in master even if we just use that for develop2?
Using different requirements for different branches can complicate the ci...

memsharded:
> But can we update the requirements to support this in master even if we just use that for develop2?
Using different requirements for different branches can complicate the ci...

Yes, yes, of course, I didn't think of that, I was concerned that someone would want to add tabs to 1.X docs.

memsharded (approved):

PR #2423, opened 2022-03-03
Update markdown generator

czoido:
Docs for: https://github.com/conan-io/conan/pull/10673


reference/generators/markdown.rst
```diff
@@ -10,7 +10,7 @@ consume them using different build systems.
 
 .. code-block:: bash
 
-    $ conan install poco/1.11.1@ --generator markdown
+    $ conan install poco/1.10.1@ --generator markdown
```
memsharded, line 13: why the downgrade?

memsharded (approved):

reference/generators/markdown.rst
czoido: Because the screenshot is for poco/1.10.1

PR #2427, opened 2022-03-03
[conf] Tools configurations improvements

franramirez688:
Docs for https://github.com/conan-io/conan/pull/10537  https://github.com/conan-io/conan/pull/10687


reference/config_files/global_conf.rst
```diff
@@ -36,6 +36,37 @@ have priority over globally defined ones in *global.conf*, and can be defined as
     tools.build:jobs=10
 
 
+Tools template
```
memsharded, line 39:
```suggestion
Configuration file template
```

Or something like that. But not tools, right?

reference/profiles.rst
```diff
@@ -165,6 +165,33 @@ Tools configurations can also be used in profile files and *global.conf* one. Pr
     You can see more information about configurations in :ref:`global.conf section <global_conf>`.
 
 
+.. _profiles_tools_values_evaluation:
+
+Tools values evaluation
```
memsharded, line 170: Not clear title.

reference/config_files/global_conf.rst
```diff
@@ -36,6 +36,37 @@ have priority over globally defined ones in *global.conf*, and can be defined as
...
+
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
+
+It is possible to use **jinja2** template engine for *global.conf*. When Conan loads this file, immediately parses
+and renders the template, which must result in a standard tools-configuration text.
+
+  .. code:: jinja
+
+     [conf]
+     # Using all the cores automatically
```
memsharded, line 50:
```suggestion
     # Using all the cores automatically
```

global.conf doesn't have the [conf] section, everything is conf

memsharded (approved):

reference/config_files/global_conf.rst
franramirez688: Then, it's wrong in the `Tools configurations` section above as well.

reference/profiles.rst
```diff
@@ -217,6 +244,68 @@ The ``include()`` statement has to be at the top of the profile file:
...
+
+Configuration data operators
++++++++++++++++++++++++++++++++++
+
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
+
+It's also possible to use some extra operators when you're composing different tools configurations:
+
+.. code-block:: text
+    :caption: global.conf
```
memsharded, line 258: This is not global.conf, but a profile, isn't it?
```suggestion
    :caption: profile_apple
```

memsharded (changes requested):

reference/profiles.rst
franramirez688: I wrote global.conf to show users that `global.conf` is the first configuration loaded (if it exists) even before the profiles. I can change it by another profile for sure but I found it interesting. Perhaps, it's not needed here...

memsharded: The thing is that this file contains [settings], [conf] sections, so it cannot be the ``global.conf`` file, it is confusing.

franramirez688: you're right, then, if I delete those sections it could run this example, right? Or would it be confusing yet?

memsharded: Yes, if they are deleted, it would be a bit clearer.

reference/profiles.rst
```diff
@@ -162,7 +162,34 @@ Tools configurations can also be used in profile files and *global.conf* one. Pr
...
+
+
+.. _profiles_configuration_data_types:
+
+Configuration data types
+++++++++++++++++++++++++++
+
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
+
+All the values will be understood by Conan as the result of the python built-in `eval()` function:
```
czoido, line 175:
```suggestion
All the values will be interpreted by Conan as the result of the python built-in `eval()` function:
```

reference/profiles.rst
```diff
@@ -217,6 +244,68 @@ The ``include()`` statement has to be at the top of the profile file:
...
+
+.. code-block:: text
+    :caption: global.conf
+
+    [settings]
+    ...
+
+    [conf]
+    # Defining several lists
+    user.myconf.build:ldflags=["-mmacosx-version-min"]
```
czoido, line 261: I think that using `-mmacosx-version-min` that is passed by the XcodeToolchain depending of the value of `os.version` or `-arch`, `-isysroot` flags make this example a bit confusing... could we use another flags or another type of conf for the example?

franramirez688: If they could be confusing then I think that I cannot use any kind of real flags because it'll be confusing as well. I'll use some fake ones only to demonstrate the lists part

memsharded: Yes, please, use fake values, not real ones, unless it is a real example of something that really works.

reference/profiles.rst
```diff
@@ -217,6 +244,64 @@ The ``include()`` statement has to be at the top of the profile file:
     zlib:CC=/usr/bin/clang
     zlib:CXX=/usr/bin/clang++
 
+
+.. _profiles_configuration_data_operators:
```
czoido, line 248: Why this section is not right after Configuration **data types** section and has **profile compositon** and **profile includes** in the middle of both?

reference/conanfile/tools/cmake/cmaketoolchain.rst
```diff
@@ -200,7 +200,7 @@ There are two ways of providing a custom CMake toolchain file:
             # tool_requires
             user_toolchains = []
             for dep in self.dependencies.direct_build.values():
-                ut = dep.conf_info["tools.cmake.cmaketoolchain:user_toolchain"]
+                ut = dep.conf_info.get("tools.cmake.cmaketoolchain:user_toolchain")
```
memsharded, line 203: Actually, this use case in the docs is probably irrelevant now? No need that recipes do weird things with this to aggregate a list, this is exactly the reason this feature was introduced?

franramirez688: I don't know what you mean. I only changed the way to get the value (it does not matter if it's a list or not), I was only updating it.

reference/config_files/global_conf.rst
```diff
@@ -36,6 +36,91 @@ have priority over globally defined ones in *global.conf*, and can be defined as
...
+    [settings]
+    [options]
+    [build_requires]
+    [env]
+    [conf]
+    user.myconf.build:cflags=!
+    user.myconf.build:ldflags=['--prefix prefix-value', '--flag1 value1', '--flag2 value2']
+    ...
+
+
```
czoido, line 120: Maybe this could be under the title: "using conf in profiles" or something like that?

reference/config_files/global_conf.rst
```diff
@@ -106,7 +191,47 @@ configuration as:
         name = "android_ndk"
 
         def package_info(self):
-            self.conf_info["tools.android:ndk_path"] = os.path.join(self.package_folder, "ndk")
+            self.conf_info.define("tools.android:ndk_path", os.path.join(self.package_folder, "ndk"))
 
 
 Note that this only propagates from the immediate, direct ``tool_requires`` of a recipe.
+
+
```
czoido, line 199: maybe we could move this before the tool_requires  part and create a two sections like:
- Using config in profiles: with the content about the profiles above
- Using config in recipes, with this content

reference/conanfile/tools/cmake/cmaketoolchain.rst
memsharded: Yes, sure, not part of your code change. But it would make sense to change or to remove that specific example from the docs, because that doesn't make sense anymore. It was a workaround for the lack of lists in conf.

reference/config_files/global_conf.rst
```diff
@@ -36,6 +36,83 @@ have priority over globally defined ones in *global.conf*, and can be defined as
...
+    user.myconf.build:flags+=["-f2"]
+
+    # Prepend the value ["-f0"] => ["-f0", "-f1", "-f2"]
+    user.myconf.build:flags=+["-f0"]
+
+    # Unset the value
+    user.myconf.build:flags=!
+
+
+Configuration list command
```
czoido, line 113: I think this section should be moved right after the "Tools configurations" section because this command is connected to that section

reference/config_files/global_conf.rst
```diff
@@ -89,6 +88,169 @@ To list all possible configurations available, run :command:`conan config list`.
     tools.system.package_manager:sudo_askpass: Use the '-A' argument if using sudo in Linux to invoke the system package manager (False by default)
 
 
+Configuration file template
+++++++++++++++++++++++++++++
```
czoido, line 92:
```suggestion
Configuration file template
---------------------------
```

reference/config_files/global_conf.rst
```diff
@@ -89,6 +88,169 @@ To list all possible configurations available, run :command:`conan config list`.
...
+     user.myconf.system:name = {{platform.system()}}
+
+
+.. note::
+
+    The Python packages passed to render the template are only ``os`` and ``platform``.
+
+
+Configuration data types
+++++++++++++++++++++++++++
```
czoido, line 113:
```suggestion
Configuration data types
------------------------
```

reference/conanfile/tools/cmake/cmaketoolchain.rst
```diff
@@ -159,15 +159,16 @@ CMake legacy generators (like ``cmake``, or ``cmake_paths``) with it.
 Using a custom toolchain file
 +++++++++++++++++++++++++++++
 
-There are two ways of providing a custom CMake toolchain file:
+There are two ways of providing a custom CMake toolchain files:
 
-- The ``conan_toolchain.cmake`` file can be completely skipped and replaced by a user one, defining the
-  ``tools.cmake.cmaketoolchain:toolchain_file=<filepath>`` configuration value
+- The ``conan_toolchain.cmake`` file can be completely skipped and replaced by an user one, defining the
+  ``tools.cmake.cmaketoolchain:toolchain_file=["<filepath>"]`` configuration list or even appending to any
```
czoido, line 165: But `tools.cmake.cmaketoolchain:toolchain_file` is not a list right? The one that is a list is: `tools.cmake.cmaketoolchain:user_toolchain`

memsharded: yes, ``user_toolchain`` is a list, this one must be a single value, as it is the cli arg -DCMAKE_TOOLCHAIN_FILE

franramirez688: oh, sure. Didn't see the whole field...

reference/config_files/global_conf.rst
```diff
@@ -89,6 +88,169 @@ To list all possible configurations available, run :command:`conan config list`.
...
+
+.. code-block:: python
+
+    import os
+    from conans import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+
+        def package_info(self):
```
czoido, line 224: The example is ok but I think that the available methods like `define`, `get`, `append`, `prepend`, `update`... and their arguments deserve a more detailed documentation

czoido (approved):

PR #2429, opened 2022-03-03
Deprecate self.cpp_info.name

### LINKED ISSUE #2404: using components still recommends cpp_info.name
from https://github.com/conan-io/conan-center-index/pull/9409#discussion_r810590913

the page https://docs.conan.io/en/latest/creating_packages/package_information.html#using-components recommends using `cpp_info.name` (even not `names`!), which was outdated a long time ago.
I think it should be updated to use `set_properties` instead.

uilianries:
The `self.cpp_info.name` is no longer used and instead, `self.cpp_info.names` should be used.

- Removed all references related to `self.cpp_info.name`
- Replaced `cpp_info.name` by `cpp_info.names`
- Added deprecation message on `cpp_info.name` description

closes #2404


memsharded (approved):

reference/conanfile/attributes.rst
```diff
@@ -967,18 +967,16 @@ This object should be filled in ``package_info()`` method.
 +--------------------------------------+---------------------------------------------------------------------------------------------------------+
 | self.cpp_info.rootpath               | Filled with the root directory of the package, see ``deps_cpp_info``                                    |
 +--------------------------------------+---------------------------------------------------------------------------------------------------------+
-| self.cpp_info.name                   | | Alternative name for the package used by generators to create files or variables.                     |
-|                                      | | Defaulted to the package name. Supported by `cmake`, `cmake_multi`, `cmake_find_package`,             |
-|                                      | | `cmake_find_package_multi`, `cmake_paths` and `pkg_config` generators.                                |
+| self.cpp_info.name                   | **[DEPRECATED]** Use ``self.cpp_info.names`` instead.                                                   |
```
SSE4, line 970: maybe set_property?

reference/generators/cmake_find_package.rst
```diff
@@ -18,8 +18,7 @@ a ``Findasio.cmake`` file will be generated.
 Variables in Find<PKG-NAME>.cmake
 ---------------------------------
 
-Being ``<PKG-NAME>`` the package name used in the reference (by default) or the one declared in ``cpp_info.name`` or in
-``cpp_info.names["cmake_find_package"]`` if specified:
+Being ``<PKG-NAME>`` the package name used in the reference (by default) or the one declared in ``cpp_info.names["cmake_find_package"]`` if specified:
```
SSE4, line 21: I think going forward, it should be `set_property("cmake_target_name")`?

reference/generators/pkg_config.rst
```diff
@@ -5,7 +5,7 @@ pkg_config
 ==========
 
 Generates pkg-config files named *<PKG-NAME>.pc* (where ``<PKG-NAME`` is the name declared by dependencies in
-``cpp_info.name`` or in ``cpp_info.names["pkg_config"]`` if specified), containing a
+``cpp_info.names["pkg_config"]`` if specified), containing a
```
SSE4, line 8: also `set_property("pkg_config_name")`, otherwise, `PkgConfigDeps` is not supported

uilianries:
Good question, should enforce `set_property` instead?

memsharded:
> Good question, should enforce set_property instead?

No, ``set_property`` only works for new generators, not for ``cmake_find_package`` ones, right? so current docs should feature both, explaining the usage for which generators, I guess.

uilianries:
Okay, so we can do it in a next PR, as it's a new feature

SSE4 (approved):

reference/generators/pkg_config.rst
czoido: I agree, as properties are supported by this generator it would be good to add a note about set_property here

PR #2445, opened 2022-03-15
Update install.rst, removing `[-sf SOURCE_FOLDER]`

Bassem-Ramzy:
Removing `[-sf SOURCE_FOLDER]` from arguments as it's not supported by the command.


memsharded:
Thanks for fixing this!

PR #2500, opened 2022-04-08
Fix CLI example for creating a new package from a template

hernando:
The command line as is produces this error:
```
conan new: error: unrecognized arguments: --template=cmake_lib
```

I had to read the code to figure out what the problem was. In particular, the command `conan new cmake_lib` produces a cryptic error form Jinja. It would be good to have a better error output from the command as well.


hernando:
[This link](https://cla-assistant.io/conan-io/docs?pullRequest=2500) doens't work for me in two different browsers.

memsharded (approved):
Thanks for your contribution, yes, it seems this is legacy from 1.X.

The CLA service we are using seems broken at the moment, lets try later (we will keep an eye on it and tell)

hernando:
The CLA is now signed.

PR #2502, opened 2022-04-18
Update confs in docs

czoido:


memsharded (approved):
Good catch, seems a wrong copy & paste from 2.0 -> 1.X docs

PR #2510, opened 2022-04-28
[CMakeDeps] Added cmake_set_interface_link_directories property info

franramirez688:
Docs for https://github.com/conan-io/conan/pull/10984

### DIFF
--- a/reference/conanfile/tools/cmake/cmakedeps.rst
+++ b/reference/conanfile/tools/cmake/cmakedeps.rst
@@ -179,8 +179,19 @@ The following properties affect the CMakeDeps generator:
 - **cmake_module_target_name**: Same as **cmake_target_name**  but when generating modules with ``cmake_find_mode=module/both``.  If not specified it will default to **cmake_target_name**.
 - **cmake_build_modules**: List of ``.cmake`` files (route relative to root package folder) that are automatically
   included when the consumer run the ``find_package()``.
+- **cmake_set_interface_link_directories**: boolean value that should be only used by dependencies that have a ``#pragma comment(lib, "foo")`` (automatic link) declared at code. Those dependencies should
+  add to their *conanfile.py* files this property at root ``cpp_info`` level (components not supported for now). For instance:
 
-Example:
+.. code-block:: python
+
+    def package_info(self):
+        self.cpp_info.includedirs.append('include')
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/cmake/cmakedeps.rst
```diff
@@ -179,8 +179,19 @@ The following properties affect the CMakeDeps generator:
...
+  add this property to their *conanfile.py* files at root ``cpp_info`` level (components not supported for now). For instance:
 
-Example:
+.. code-block:: python
+
+    def package_info(self):
+        self.cpp_info.includedirs.append('include')
+        self.cpp_info.libdirs.append('lib')
+        self.cpp_info.libs = []
+        self.cpp_info.set_property("cmake_set_interface_link_directories", True)
```
czoido, line 191: I think we can skip the example or just add the `self.cpp_info.set_property("cmake_set_interface_link_directories", True)` with a comment to the example below

reference/conanfile/tools/cmake/cmakedeps.rst
```diff
@@ -179,8 +179,19 @@ The following properties affect the CMakeDeps generator:
 - **cmake_module_target_name**: Same as **cmake_target_name**  but when generating modules with ``cmake_find_mode=module/both``.  If not specified it will default to **cmake_target_name**.
 - **cmake_build_modules**: List of ``.cmake`` files (route relative to root package folder) that are automatically
   included when the consumer run the ``find_package()``.
+- **cmake_set_interface_link_directories**: boolean value that should be only used by dependencies that have a ``#pragma comment(lib, "foo")`` (automatic link) declared at code. Those dependencies should
```
lasote, line 182:
```suggestion
- **cmake_set_interface_link_directories**: boolean value that should be only used by dependencies that don't declare `self.cpp_info.libs` but have ``#pragma comment(lib, "foo")`` (automatic link) declared at the public headers. Those dependencies should
```

lasote: I agree with @czoido, the example is not so relevant. I think the sentence above is enough.

lasote (changes requested):

lasote (approved):

PR #2514, opened 2022-04-28
Renamed all the base_source_folder -> export_sources_folder

franramirez688:
Docs for https://github.com/conan-io/conan/pull/11055

_Coming from the original Conan PR https://github.com/conan-io/conan/pull/11055:_

> Changelog: Fix: Renamed self.base_source_folder to self.export_source_folder. That variable was introduced to reference the folder where the export_sources are. Currently, they are copied to the source folder but might be changed in the future to avoid copying them, so self.export_source_folder will always point to the exports_sources.

### DIFF
--- a/developing_packages/package_layout.rst
+++ b/developing_packages/package_layout.rst
@@ -443,8 +443,8 @@ also able to locate the ``my_tool`` correctly, because it is using the same ``fo
 
 
 
-Example: base_source_folder
----------------------------
+Example: export_sources_folder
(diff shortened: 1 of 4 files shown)


SSE4:
it lacks some motivation behind the rename, could you add some notes on why was it needed, or what kind of problem were you trying to solve by this rename?

lasote (approved):

czoido (approved):

PR #2519, opened 2022-05-03
Example for conanrun

franramirez688:
Docs for https://github.com/conan-io/docs/issues/2465

### DIFF
--- a/reference/conanfile/tools/env/virtualrunenv.rst
+++ b/reference/conanfile/tools/env/virtualrunenv.rst
@@ -63,7 +63,6 @@ current active configuration, like ``deactivate_conanrunenv-release-x86_64.bat``
 Given that, let's see, for instance, how to set the ``PYTHONPATH`` through the ``runenv_info`` to be applied when running a command
 from ``test()`` side:
 
-conanfile.py
 
 .. code-block:: python
     :caption: conanfile.py
--- a/reference/conanfile/tools/env/virtualrunenv.rst
+++ b/reference/conanfile/tools/env/virtualrunenv.rst
@@ -68,32 +68,81 @@ from ``test()`` side:
     :caption: conanfile.py
 
     import os
+
     from conan import ConanFile
+    from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
 
-    class HelloConan(ConanFile)
 
-        # ...
+    class HelloConan(ConanFile):
+        name = "hello"
+        version = "1.0"
+        # Binary configuration
+        settings = "os", "compiler", "build_type", "arch"
+        options = {"shared": [True, False], "fPIC": [True, False]}
+        default_options = {"shared": False, "fPIC": True}
+        # Sources are located in the same place as this recipe, copy them to the recipe
+        exports_sources = "CMakeLists.txt", "src/*", "include/*"
+
+        def config_options(self):
+            if self.settings.os == "Windows":
+                del self.options.fPIC
+
+        def layout(self):
+            cmake_layout(self)
+
+        def generate(self):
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/env/virtualrunenv.rst
```diff
@@ -60,6 +60,43 @@ After the execution of one of those files, a new deactivation script will be gen
...
+
+    import os
+    from conan import ConanFile
+
+    class HelloConan(ConanFile)
+
+        # ...
+
+        def package_info(self):
+            self.runenv_info.append("PYTHONPATH", os.path.join(self.package_folder, "my-site-packages"))
```
memsharded, line 78:
```suggestion
            self.runenv_info.append_path("PYTHONPATH", os.path.join(self.package_folder, "my-site-packages"))
```
Maybe even prepend_path?

reference/conanfile/tools/env/virtualrunenv.rst
```diff
@@ -60,6 +60,43 @@ After the execution of one of those files, a new deactivation script will be gen
...
+
+    class HelloTestConan(ConanFile):
+        # VirtualBuildEnv and VirtualRunEnv can be avoided if "tools.env.virtualenv:auto_use" is defined
+        # (it will be defined in Conan 2.0)
+        generators = "VirtualRunEnv"
+
+        # ...
+
+        def test(self):
+            self.run("set PYTHONPATH", env="conanrun")
```
memsharded, line 94: This is Windows only, better clarify or remove it

memsharded (commented):
Is this extracted from some running test? Maybe it is worth copy & pasting from a test to avoid possible minor issues?

reference/conanfile/tools/env/virtualrunenv.rst
franramirez688: Oh, sure!

reference/conanfile/tools/env/virtualrunenv.rst
franramirez688: I can tell it in a comment

memsharded: In any case, the ``_path`` is necessary, isn't it?

franramirez688: that's it

franramirez688:
> Do we need the entire source code? Maybe just use the relevant parts?

Perhaps, it's clearer to see a full-working example. I guess it's fair enough to put the relevant parts but a few lines more are worth it... 

> Also, shouldn't we document this in the [conanfile.py](https://docs.conan.io/en/1.3/reference/conanfile/other.html) section? About the env argument in the run method and that it's defaulted to conanbuild ?

I already added some info there. Have a look at this PR: https://github.com/conan-io/docs/pull/2516/files If you think it's not enough I can add something more.

reference/conanfile/tools/env/virtualrunenv.rst
```diff
@@ -60,6 +60,45 @@ After the execution of one of those files, a new deactivation script will be gen
 environment, so the environment can be restored when desired. The file will be named also following the
 current active configuration, like ``deactivate_conanrunenv-release-x86_64.bat``.
 
+Given that, let's see, for instance, how to set a ``MYPATH`` variable through the ``runenv_info`` to be applied when running a command
+from ``test()`` side:
```
czoido, line 64: Maybe we can rephrase, something similar to this?

```suggestion
Let's see and example on how to add an environment variable to the runenv_info and getting its value later in the consumer explicitly invoking the `conanrun` environment.
```

memsharded (approved):

PR #2525, opened 2022-05-06
CMakeToolchain examples

lasote:


_themes/conan/static/css/theme.css
```diff
@@ -18,4 +18,15 @@
...
+    margin-left: 10px;
+    list-style:none;
+ }
+
+ .container.examples li:before{
+    content: '\1F4D8';
+    margin-left: -20px;
+    margin-right: 10px;
+ }
+
```
lasote, line 31: I missed to do some "art"
image

memsharded (approved):

examples/tools/cmake/cmake_toolchain/build_project.rst
```diff
@@ -0,0 +1,79 @@
+
+.. _examples-tools-cmake-toolchain-build-project:
+
+CMakeToolchain: Building your project
```
czoido, line 4: I think that maybe this should not be an example, this overlaps a bit with this part of the tutorial: https://docs.conan.io/en/2.0-alpha/tutorial/consuming_packages/the_flexibility_of_conanfile_py.html#use-the-layout-method where we explain how the layout() works and that you can use the pre-defined cmake_layout(). Maybe it's just a matter of adding a couple more sentences there, instead of adding a whole example.

czoido (approved):

PR #2528, opened 2022-05-09
Tutorial cross building

czoido:
TODO: add some diagram depicting host and build context.

Goes with https://github.com/conan-io/examples2/pull/17

### DIFF
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -1,0 +1,35 @@
+.. _consuming_packages_cross_building_with_conan:
+
+How to cross-compile your applications using Conan: host and build contexts
+===========================================================================
+
+.. important::
+
+    In this example, we will retrieve Conan packages from a Conan repository with
+    packages compatible with Conan 2.0. To run this example successfully you should add this
+    remote to your Conan configuration (if did not already do it) doing:
+    ``conan remote add conanv2 https://conanv2beta.jfrog.io/artifactory/api/conan/conan --index 0``
+
+
+In the previous examples, we learned how to use a *conanfile.py* or *conanfile.txt* to
+build an application that compresses strings using the *Zlib* and *CMake* Conan
+packages. Also, we explained that you can set information like the operating system,
+compiler or build configuration in a file called the Conan profile.
+
+For all these examples we used the same platform for building and running the application.
+But, what if you want to build the application in your machine running Macos but you would
+like that application to run it other platform like, for example, a Raspberry Pi? Conan
+can model that case by using two different profiles, one for the machine that **builds**
+the application (Macos) and another one for the machine that **runs** the application
+(Raspberry Pi). We will explain this "two profiles" approach in the section that follows.
+
+Conan two profiles: build and host profiles
+-------------------------------------------
+
+...
+
+Read more
+---------
+
+- Using Conan with Android Studio
+
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -82,7 +82,7 @@ could use the default profile, that in our case looks like this:
     compiler.libcxx=libstdc++11
     compiler.version=9
 
-And the profile for the Raspberry Pi that is the **host** machine could look like this:
+And the profile for the Raspberry Pi that is the **host** machine would be:
 
 .. code-block:: bash
     :caption: <local folder>/profiles/raspberry
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -51,7 +51,7 @@ Is equivalent to:
 
 .. code-block:: bash
 
-    $ conan install . --build=missing --profile:host=someprofile --profile:build=someprofile
+    $ conan install . --build=missing --profile:host=someprofile --profile:build=default
 
 
 As you can see we used two new arguments:
@@ -62,6 +62,10 @@ As you can see we used two new arguments:
 * ``profile:build``: This is the profile that that defines the platform where the we will
   build the binaries. For our string compressor application this profile would be the one
   applied to the *CMake* tool that will run on the **Ubuntu Linux** machine.
+
+Note that when you just use one argument for the profile ``--profile`` is equivalent to
+``--profile:host``. If you don't specify the ``--profile:build`` argument, Conan will use
+the *default* profile internally.
 
 So, if we want to build the compressor application in the Ubuntu Linux machine but run it in a
 Raspberry Pi, we could define two different profiles. The profile for our **build** machine
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -28,8 +28,8 @@ argument (``--profile``) to invoke the :command:`conan install`. We also explain
 not specifying that profile is equivalent to using the ``--profile=default`` argument.
 
 For all those examples, we used the same platform for building and running the
-application. But what if you want to build the application in your machine running Ubuntu
-Linux, but you would like that application to run on other platform like, for example, a
+application. But, what if you want to build the application on your machine running Ubuntu
+Linux and then run it on another platform like a
 Raspberry Pi? Conan can model that case using two different profiles, one for the
 machine that **builds** the application (Ubuntu Linux) and another for the machine that
 **runs** the application (Raspberry Pi). We will explain this "two profiles" approach in
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -151,9 +151,8 @@ context" and the "host context":
 
 * The **host context** is populated with the root package (the one specified in the
   :command:`conan install` or :command:`conan create` command) and all its requirements.
-  In some cases, it may also include tools that should run in the host machine (for example
-  a test framework that will run in the host machine). In this case, this includes the
-  compressor application and the **zlib/1.2.11** dependency.
+  In this case, this includes the compressor application and the **zlib/1.2.11**
+  dependency.
 
 * The **build context** contains the tool requirements used in the build machine. This
   category typically includes all the developer tools like CMake, compilers and linkers.
@@ -216,5 +215,6 @@ Read more
 ---------
 
 - Cross-build using a tool_requires
+- How to require test frameworks: using ``test_requires``
 - Using Conan to build for Android
 - Using Conan to build for iOS
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -86,7 +86,7 @@ could use the default profile, that in our case looks like this:
 And the profile for the Raspberry Pi that is the **host** machine could look like this:
 
 .. code-block:: bash
-    :caption: <conan home>/profiles/raspberry
+    :caption: <local folder>/profiles/raspberry
     :emphasize-lines: 9-12
 
     [settings]
@@ -174,7 +174,7 @@ dependency built for *armv7hf* architecture and a **cmake/3.19.8** version that 
 
 .. code-block:: bash
     
-    $ conan install . --build missing -pr:b=./profiles/default -pr:h=./profiles/raspberry
+    $ conan install . --build missing -pr:b=default -pr:h=./profiles/raspberry
 
 Then, let's call CMake to build the application. As we did in the previous example we have
 to activate the **build environment** running ``source generators/conanbuild.sh``. That will
--- a/tutorial/consuming_packages/cross_building_with_conan.rst
+++ b/tutorial/consuming_packages/cross_building_with_conan.rst
@@ -11,6 +11,15 @@ How to cross-compile your applications using Conan: host and build contexts
     ``conan remote add conanv2 https://conanv2beta.jfrog.io/artifactory/api/conan/conan --index 0``
 
 
+Please, first clone the sources to recreate this project. You can find them in the
+`examples2.0 repository <https://github.com/conan-io/examples2>`_ in GitHub:
+
+.. code-block:: bash
+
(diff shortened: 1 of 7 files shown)


tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+.. code-block:: bash
+
+    $ conan install . --build=missing --profile=someprofile
+
+Is equivalent to:
+
+.. code-block:: bash
+
+    $ conan install . --build=missing --profile:host=someprofile --profile:build=someprofile
+
```
lasote, line 55: I don't think that is correct:

 - `--profile` is equivalent to `--profile:host` 
 - If you don't specify `--profile:build` the applied build profile `is not` `someprofile`, is `default`.

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+We will need the application to build for the Raspberry Pi with the cross-build
+toolchain and also link the **zlib/1.2.11** library built for the same platform. On the
+other side, we need the **cmake/3.19.8** binary to run in Ubuntu Linux. Conan manages this
+internally in the dependency graph differentiating between what we call the "build
+context" and the "host context":
+
+* The **host context** is populated with the root package (the one specified in the
+  :command:`conan install` or :command:`conan create` command) and all its requirements.
+  In some cases, it may also include tools that should run in the host machine (for example
+  a test framework that will run in the host machine). In this case, this includes the
```
lasote, line 152: mention `self.test_requires()`?

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+build an application that compresses strings using the *Zlib* and *CMake* Conan packages.
+Also, we explained that you can set information like the operating system, compiler or
+build configuration in a file called the Conan profile. You can use that profile as an
+argument (``--profile``) to invoke the :command:`conan install`. We also explained that
+not specifying that profile is equivalent to using the ``--profile=default`` argument.
+
+For all those examples, we used the same platform for building and running the
+application. But what if you want to build the application in your machine running Ubuntu
+Linux, but you would like that application to run on other platform like, for example, a
+Raspberry Pi? Conan can model that case using two different profiles, one for the
```
danimtb, line 33:
```suggestion
application. But, what if you want to build the application on your machine running Ubuntu
Linux and then run it on another platform like a
Raspberry Pi? Conan can model that case using two different profiles, one for the
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+    $ git clone https://github.com/conan-io/examples2.git
+    $ cd examples2/tutorial/consuming_packages/cross_building
+
+
+In the previous examples, we learned how to use a *conanfile.py* or *conanfile.txt* to
+build an application that compresses strings using the *Zlib* and *CMake* Conan packages.
+Also, we explained that you can set information like the operating system, compiler or
+build configuration in a file called the Conan profile. You can use that profile as an
+argument (``--profile``) to invoke the :command:`conan install`. We also explained that
+not specifying that profile is equivalent to using the ``--profile=default`` argument.
```
danimtb, line 28:
```suggestion
argument (:command:`--profile`) to invoke the :command:`conan install`. We also explained that
not specifying that profile is equivalent to using the :command:`--profile=default` argument.
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+``cmake_layout``.
+
+We will need the application to build for the Raspberry Pi with the cross-build
+toolchain and also link the **zlib/1.2.11** library built for the same platform. On the
+other side, we need the **cmake/3.19.8** binary to run in Ubuntu Linux. Conan manages this
+internally in the dependency graph differentiating between what we call the "build
+context" and the "host context":
+
+* The **host context** is populated with the root package (the one specified in the
+  :command:`conan install` or :command:`conan create` command) and all its requirements.
```
lasote, line 150: mention `self.requires()` ?

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+Linux, but you would like that application to run on other platform like, for example, a
+Raspberry Pi? Conan can model that case using two different profiles, one for the
+machine that **builds** the application (Ubuntu Linux) and another for the machine that
+**runs** the application (Raspberry Pi). We will explain this "two profiles" approach in
+the next section.
+
+Conan two profiles model: build and host profiles
+-------------------------------------------------
+
+Although for every conan command you can specify only one ``--profile`` argument, Conan
```
danimtb, line 41:
```suggestion
Although for every command you can specify only one :command:`--profile` argument, Conan
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+
+* The **host context** is populated with the root package (the one specified in the
+  :command:`conan install` or :command:`conan create` command) and all its requirements.
+  In some cases, it may also include tools that should run in the host machine (for example
+  a test framework that will run in the host machine). In this case, this includes the
+  compressor application and the **zlib/1.2.11** dependency.
+
+* The **build context** contains the tool requirements used in the build machine. This
+  category typically includes all the developer tools like CMake, compilers and linkers.
+  In this case, this includes the **cmake/3.19.8** tool.
```
lasote, line 157: mention `self.tool_requires()` ?

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+
+
+As you can see we used two new arguments:
+
+* ``profile:host``: This is the profile that defines the platform where the built binaries
+  will run. For our string compressor application this profile would be the one applied
+  for the *Zlib* library that will run in a **Raspberry Pi**.
+* ``profile:build``: This is the profile that that defines the platform where the we will
+  build the binaries. For our string compressor application this profile would be the one
+  applied to the *CMake* tool that will run on the **Ubuntu Linux** machine.
```
danimtb, line 64:
```suggestion
* ``profile:build``: This is the profile that defines the platform where the binaries will be built. For our string compressor application, this profile would be the one
  used by the *CMake* tool that will compile it on the **Ubuntu Linux** machine.
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+    [settings]
+    os=Linux
+    arch=x86_64
+    build_type=Release
+    compiler=gcc
+    compiler.cppstd=gnu14
+    compiler.libcxx=libstdc++11
+    compiler.version=9
+
+And the profile for the Raspberry Pi that is the **host** machine could look like this:
```
danimtb, line 86:
```suggestion
And the profile for the Raspberry Pi that is the **host** machine:
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+the **cmake/3.19.8** dependency.
+
+Now, let's build the application. First, call :command:`conan install` with the
+profiles for the build and host platforms. This will install the  **zlib/1.2.11**
+dependency built for *armv7hf* architecture and a **cmake/3.19.8** version that runs for
+64-bit architecture.
+
+.. code-block:: bash
+    
+    $ conan install . --build missing -pr:b=./profiles/ubuntu -pr:h=./profiles/raspberry
```
lasote, line 173: Given the comments I did about the default build profile when not specified, Is it really useful to show the "ubuntu" profile given that it would be probably `just the default`?

lasote (approved):
Good job!

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+            self.requires("zlib/1.2.11")
+            self.tool_requires("cmake/3.19.8")
+
+        def layout(self):
+            cmake_layout(self)
+
+As you can see, this is practically the same *conanfile.py* we used in the previous
+example. We will require **zlib/1.2.11** as a regular dependency and **cmake/3.19.8** as a
+tool needed for building the application. Also, we are using the pre-defined
+``cmake_layout``.
```
danimtb, line 145: - If "the previous example" of the recipe is importat to undestard this section, it'd be helpful to have a hyperlink to it. If it isn't, then I think the plural "example**s**" is better here to denote that "it is almost the same but not important to check it now"
- ``cmake_layout`` does not look very important to me in this section, as I want to focus on crossbuilding and avoid any distraction, so I would omit it here (unless it is important later in the explanation -in that case, I would add a hyperlink to the documentation of the layouts as well-).
```suggestion
As you can see, this is practically the same *conanfile.py* we used in the previous
examples. We will require **zlib/1.2.11** as a regular dependency for our app and **cmake/3.19.8** as a
tool to build it.
```

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,217 @@
...
+dependency built for *armv7hf* architecture and a **cmake/3.19.8** version that runs for
+64-bit architecture.
+
+.. code-block:: bash
+    
+    $ conan install . --build missing -pr:b=./profiles/ubuntu -pr:h=./profiles/raspberry
+
+Then, let's call CMake to build the application. As we did in the previous example we have
+to activate the **build environment** running ``source generators/conanbuild.sh``. That will
+set the environment variables needed to locate the cross-build toolchain and build the
```
danimtb, line 181: I am missing here information about this build environment file. I understand the purpose but would like to understand how conan is able to generate that file for me and where the information comes from.
- Does it come from the profile?
- Is Conan detecting something for me? 
- Does it work "out of the box" or do I have to configure something?

Maybe those questions arise because I have not read (or can't remember) other/previous sections of the documentation. If it is too much, then a link to that reference could be useful for the reader. But I think this "build environment" setup is key for users doing crossbuilding.

tutorial/consuming_packages/cross_building_with_conan.rst
memsharded: I kind of like making it explicit here for educational purposes, at least once, it can be commented that it could be the machine default and unspecified, and then drop in next commands.

danimtb (commented):
Great explanations, simple read and simple to understand. Just some suggestions from my side 👍

tutorial/consuming_packages/cross_building_with_conan.rst
danimtb: I think having it in the recipe is enough and adding it here will probably make the explanation more complex to understand

tutorial/consuming_packages/cross_building_with_conan.rst
danimtb: I think having it in the recipe is enough and adding it here will probably make the explanation more complex to understand

tutorial/consuming_packages/cross_building_with_conan.rst
danimtb: IMO it is too much to explain it here, better to keep focus on understanding the cross-building scenarios and the different contexts

lasote: But then let`s remove the `for example
  a test framework that will run in the host machine)`, If I read that I think... so how do I do that?

tutorial/consuming_packages/cross_building_with_conan.rst
lasote: My point is, `and all its requirements.`  might be misleading, the `CMake` package is also a requirement.

tutorial/consuming_packages/cross_building_with_conan.rst
lasote: ok in this case.

tutorial/consuming_packages/cross_building_with_conan.rst
```diff
@@ -0,0 +1,221 @@
...
+Linux, but you would like that application to run on other platform like, for example, a
+Raspberry Pi? Conan can model that case using two different profiles, one for the
+machine that **builds** the application (Ubuntu Linux) and another for the machine that
+**runs** the application (Raspberry Pi). We will explain this "two profiles" approach in
+the next section.
+
+Conan two profiles model: build and host profiles
+-------------------------------------------------
+
+Although for every conan command you can specify only one ``--profile`` argument, Conan
```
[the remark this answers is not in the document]
czoido: Yes, it's a bit confusing, I'll rephrase that

tutorial/consuming_packages/cross_building_with_conan.rst
czoido: Yes, I think I'll remove the sentence and leave the test_requires for a read more

tutorial/consuming_packages/cross_building_with_conan.rst
czoido: Ok, I'll clarify that those are added via self.requires()

danimtb: In that case, to refer to CMake and so, it could be `its tool-requirements` then 😆 but I understand the point that it could be misleading. Up to @czoido choice in my case 🤓

tutorial/consuming_packages/cross_building_with_conan.rst
czoido: There's a brief introduction [here](https://docs.conan.io/en/2.0-alpha/tutorial/consuming_packages/use_tools_as_conan_packages.html) about the VirtualBuildEnv generator and also here to the [VirtualRunEnv](https://docs.conan.io/en/2.0-alpha/tutorial/consuming_packages/different_configurations.html). I'll add a comment there so that the user is aware that adding the [buildenv] section the VirtualBuildEnv generator is automatically invoked.

danimtb (approved):

PR #2532, opened 2022-05-11
reference tools.gnu

lasote:
First merge: https://github.com/conan-io/conan/pull/11229 Otherwise CI fail because of an incorrect docstring


reference/tools/gnu/autotools.rst
```diff
@@ -0,0 +1,111 @@
...
+Reference
+---------
+
+.. currentmodule:: conan.tools.gnu.autotools
+
+.. autoclass:: Autotools
+    :members:
+
+
+A note about relocatable shared libraries in macOS built the  Autotools build helper
```
memsharded, line 41: I think I would make this its own page, and linked it from here. Not sure where, but it feels a bit too much here (can stay here, if we don't find a better place)

reference/tools/gnu/autotoolstoolchain.rst
```diff
@@ -0,0 +1,175 @@
...
+
+    .. note::
+        It is **not valid** to change the self.cpp_info  at the ``package_info()`` method.
+
+
+Customizing the environment
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If your ``Makefile`` or ``configure`` scripts need some other environment variable rather than ``CPPFLAGS``, ``LDFLAGS``,
+``CXXFLAGS`` or ``CFLAGS``, you can customize it before calling the ``generate()`` method.
```
memsharded, line 137: Probably some pointer (at least a TODO comment) that this might be better done with the new ``conf....CXXFLAGS`` from profiles ``conf`` section.

memsharded (approved):

reference/tools/gnu/autotools.rst
lasote: I agree, but let's agree first where.

reference/tools/gnu/autotoolstoolchain.rst
lasote: The list of [conf] was defined later in the page at the reference, I added a note.

memsharded (approved):

czoido (approved):

PR #2546, opened 2022-05-25
conanfile reference

lasote:
- Conanfile.py attribute reference.
- A lot of <MISSING PAGE> left that we can remove as far as we progress with the docs.
- It is recommended to render the docs doing a checkout of XXX to review the contents.


memsharded (approved):
good job

reference/conanfile.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. currentmodule:: conan
+.. autoclass:: ConanFile
+
+   .. code-block:: python
+
+       from conan import ConanFile
+
+       class HelloConan(ConanFile):
+           ...
```
czoido, line 15: This looks a bit weird to me as it is, I would give a bit of introduction or maybe remove it? Is the autoclass really helpful there?
image

reference/conanfile/attributes.rst
```diff
@@ -0,0 +1,1022 @@
...
+
+   A valid name has:
+
+    - A minimum of 2 and a maximum of 101 characters (though shorter names are recommended).
+    - Matches the following regex ``^[a-z0-9_][a-z0-9_+.-]{1,100}$``: so starts with alphanumeric or ``_``,
+      then from 1 to 100 characters between alphanumeric, ``_``, ``+``, ``.`` or ``-``.
+
+   The name is only necessary for ``export``-ing the recipe into the local cache (``export`` and ``create`` commands),
+   if they are not defined in the command line.
+
```
czoido, line 28: Are all of the indents on purpouse? Shouldn't it be aligned to the left?
image

lasote: It was on purpose, so the text outside the ``docstrings is in the same tabulation, so you don't care if that is from docstring or from the docs. But could be changed.

czoido: I prefer to have them differentiated, but let's discuss and see what others think.

reference/conanfile/attributes.rst
```diff
@@ -0,0 +1,1022 @@
...
+  .. code-block:: python
+
+     class MyPkg(ConanFile):
+         tool_requires = "tool_a/0.2", "tool_b/0.2@user/testing"
+
+
+  .. seealso::
+
+      - Check <MISSING PAGE> tool_requires() conanfile.py method.
```
czoido, line 403: I think I would replace this with a paragraph explaining that this is a declarative way of adding the tool_requires but if you want to do that "programmatically" you have the tool_requires() method and add the link. The same for build_requires, etc...

reference/conanfile/attributes.rst
lasote: Everything is now automatically without margin left.

czoido (approved):

PR #2560, opened 2022-06-01
Add note for components support

czoido:
Docs for: https://github.com/conan-io/conan/pull/11233

### DIFF
--- a/reference/conanfile/tools/apple.rst
+++ b/reference/conanfile/tools/apple.rst
@@ -103,6 +103,14 @@ If you want to add this dependencies to you Xcode project, you just have to add 
 *conan_config.xcconfig* configuration file for all of the configurations you want to use
 (usually *Debug* and *Release*).
 
+Components support
+++++++++++++++++++
+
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/apple.rst
```diff
@@ -103,6 +103,14 @@ If you want to add this dependencies to you Xcode project, you just have to add
 *conan_config.xcconfig* configuration file for all of the configurations you want to use
 (usually *Debug* and *Release*).
 
+Components support
+++++++++++++++++++
+
+Since Conan version `1.49.0 <https://github.com/conan-io/conan/releases>`_, this generator
+supports packages with components. That means that if a library only requires certain
```
memsharded, line 110: Not completely clear. I think it is worth clarifying:

- If a **dependency** ``package_info()`` declares ``cpp_info.requires`` on some components, the generated xcconfig will contain only those components
- The current package ``requires`` will be fully dependent on and generated in the ``xcconfig`` at this moment. Recall that the ``package_info()`` only applies for consumers, but not to the current package.

PR #2583, opened 2022-06-09
Added documentation for traits.

ZbigniewRA:
This is based on https://github.com/conan-io/tribe/blob/main/design/026-requirements_traits.md
It is a step towards #2581


memsharded (commented):
This is looking good. The only thing is that it should probably be a "second" level in hierarchy, as it is a thing of ``self.requires`` (an attribute/method), which already has a section.

I'll let @czoido propose a location

czoido:
Hi @ZbigniewRA,
Thanks a lot for the contribution, I'll make some changes to move this to the conanfile reference under the self.requires section so we can merge this.

reference/conanfile/methods.rst
```diff
@@ -1,2 +1,130 @@
...
+
+.. _conan_conanfile_methods:
+
 Methods
-=======
\ No newline at end of file
+=======
+
+
+requirements()
```
czoido, line 12: I added no autodocs, just leaving this here so we don't lose it, it may go to the dosctring in the future.

PR #2587, opened 2022-06-14
Feature/creating other packages header only

lasote:
Examples here: https://github.com/conan-io/examples2/pull/24


tutorial/creating_packages/other_types_of_packages/header_only_packages.rst
```diff
@@ -1,2 +1,251 @@
...
+
+.. code-block:: bash
+
+    $ conan create . -s build_type=Debug
+    $ conan list packages sum/0.1#latest
+    Local Cache:
+        sum/0.1#154c10cda76635e3ca85bcd7e21c2de9:da39a3ee5e6b4b0d3255bfef95601890afd80709
+
+Even in the ``test_package`` executable is built for Debug, we get the same binary package for the header-only library.
+
```
czoido, line 95: Maybe here we can comment again that this is because we did not declare the settings in the recipe, because people may not be aware that this is the result from that

czoido (approved):

PR #2605, opened 2022-06-23
Basic create tool requires

lasote:
With: https://github.com/conan-io/examples2/pull/29


tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,139 @@
...
+- The generators ``VirtualBuildEnv`` and ``VirtualRunEnv`` are automatically used.
+- The ``VirtualRunEnv`` is reading the ``tool-requires`` and is creating a launcher like ``conanbuildenv-release-x86_64.sh`` appending
+  all ``cpp_info.bindirs`` to the ``PATH``, all the ``cpp_info.libdirs`` to the ``LD_LIBRARY_PATH`` environment variable and
+  declaring each variable of ``self.buildenv_info``.
+- Every time conan executes the ``self.run``, by default, activates the ``conanbuild.sh`` file before calling any command.
+  The ``conanbuild.sh`` is including the ``conanbuildenv-release-x86_64.sh``, so the application is in the PATH
+  and the enviornment variable "MYVAR" has the value declared in the tool-require.
+
+
+More complex recipes
```
lasote, line 134: What else do we want to explain?

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,139 @@
...
+
+Let's create a binary package for the tool_require:
+
+.. code-block:: bash
+
+    $ conan create .
+    ...
+    my_tool/1.0: Calling package()
+    my_tool/1.0 package(): Packaged 1 '.sh' file: say_hello.sh
+    my_tool/1.0 package(): Packaged 1 '.bat' file: say_hello.bat
```
memsharded, line 69: This is one or the other line, right?

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,139 @@
...
+        settings = "os", "arch", "compiler", "build_type"
+        tool_requires = "my_tool/1.0"
+
+        def build(self):
+            if self.settings_build.os != "Windows"
+                self.run("say_hello.sh")
+                self.run("echo MYVAR=$MYVAR")
+            else:
+                self.run("say_hello.bat")
+                self.run("echo MYVAR=%MYVAR%")
```
memsharded, line 94:
```suggestion
                self.run("set MYVAR")
```
To match the example

memsharded (approved):

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
memsharded: Good question, maybe the case of using a real .exe? So it requires settings, and the build-host context becomes more evident regarding ``tool_requires``

memsharded: Those could be ReadMore examples, not necessarily here in main line of the tutorial

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,139 @@
...
+    class my_tool(ConanFile):
+        name = "my_tool"
+        version = "1.0"
+        package_type = "application"
+
+        settings = "os", "arch"
+
+        def package(self):
+            extension = "sh" if self.settings_build.os != "Windows" else "bat"
+            app_path = os.path.join(self.package_folder, "bin", "say_hello.{}".format(extension))
```
czoido, line 40: I think I would prefer to see a more close to reality example here, with an executable... maybe some tool that verifies something in the consumer build? Because this could be closer to the typicall question that sometines users ask on "how to package all their build scripts and use them in another build?" we may put in a separate example

memsharded: I also want to see the case of the executable, I suggested that above, as the possible following example.
But I don't dislike this one either, I think it is a super common use case. We know tons of users that are packaging scripts, cmake utilities and other stuff in ``build_requires``. Maybe even more common that packaging executable tools?

czoido: As I see it, I would explain the case of an executable in the main tutorial history (something inspired in tools like cmake, meson, ninja, nasm... ) and then the case of packaging build scripts in an example. If we want to show the case of the build scripts here in the main tutorial then I would make it a bit closer to the usual case exporting the scripts, not generating them... and make the title more close to: "How to package my build scripts to use in other recipes" that is the question that people will search for.

memsharded: Yes, that makes sense too, the case of using exported scripts, I agree, I already commented that.

I'd probably keep the "package my build script" here in the main tutorial line, not in an example, but I am fine either way.

lasote: This is not trying to show how to package "build shell scripts" but the concept of packaging any application to be run in build. But as it is confusing you I'll try to change it. Before you suggest, I'll not use the ``cmake_exe`` template because the readers will learn nothing.

czoido:
There's a common pattern for tool requires recipes that is adding this:

```
    def package_id(self):
        del self.info.settings.compiler
```
So that you can use the same binary regardless of the compiler you are using.

SSE4:
> There's a common pattern for tool requires recipes that is adding this:
> 
> ```
>     def package_id(self):
>         del self.info.settings.compiler
> ```
> 
> So that you can use the same binary regardless of the compiler you are using. Maybe we should mention this

I wouldn't say it's a common pattern. for some tools (`meson`, `autoconf`) compiler is for sure irrelevant (there is no binary, actually, there are just some scripts). for others (`cmake`) it's relevant, as binary is different per different different compiler (at very least, they link to different runtime DLLs).

czoido:
> I wouldn't say it's a common pattern. for some tools (`meson`, `autoconf`) compiler is for sure irrelevant (there is no binary, actually, there are just some scripts). for others (`cmake`) it's relevant, as binary is different per different different compiler (at very least, they link to different runtime DLLs).

That is done in 43 recipes in Conan Center, included cmake (I think most of them are tool requires). I thought the point of that was that if you want to use cmake, for example, Conan does not tell you to build it if you are using a compiler that is different from the one that was used to build cmake.

czoido:
> Visual Studio CRT

I really don't know if we should recommend this or not for tool requires, but if it's useful in Conan Center, I think it's worth mentioning this in the docs.

SSE4 (approved):

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,226 @@
...
+
+    $ conan create .
+    ...
+    Package '82339cc4d6db7990c1830d274cd12e7c91ab18a1' created
+
+    $ conan create . -s build_type=Debug
+    ...
+    Package '82339cc4d6db7990c1830d274cd12e7c91ab18a1' created
+
+We got the same binary package.
```
memsharded, line 218:
```suggestion
We got the same binary ``package_id``. The second ``conan create . -s build_type=Debug`` created and overwrote (created a newer package revision) of the previous Release binary, because they have the same ``package_id`` identifier. It is typical to create only the ``Release`` one, and if for any reason managing both Debug and Release binaries is intended, then the approach would be not removing the ``del self.info.settings.build_type``
```

tutorial/creating_packages/other_types_of_packages/tool_requires_packages.rst
```diff
@@ -1,2 +1,226 @@
...
+        version = "1.0"
+        settings = "os", "compiler", "build_type", "arch"
+        ...
+
+        def package_id(self):
+            del self.info.settings.compiler
+            del self.info.settings.build_type
+
+
+So, if we call :command:`conan create` with different ``build_type`` we will get exactly the same binary.
```
memsharded, line 205:
```suggestion
So, if we call :command:`conan create` with different ``build_type`` we will get exactly the same ``package_id``.
```

memsharded (approved):

PR #2606, opened 2022-06-23
Preparing build section about the generate() method

czoido:
Goes with: https://github.com/conan-io/examples2/pull/30


memsharded (approved):
Looks great. 

Agree with @lasote, a bit of explanation about the purpose and flow of ``generate()`` will help, but overall is looking great.

memsharded (approved):

PR #2617, opened 2022-06-28
'rm' tool from conan.tools.files

lasote:
https://github.com/conan-io/conan/pull/11443/files

### DIFF
--- a/reference/conanfile/tools/files/basic.rst
+++ b/reference/conanfile/tools/files/basic.rst
@@ -138,6 +138,31 @@ Parameters:
     - **encoding** (Optional, Defaulted to ``utf-8``): Specifies the input and output files text encoding.
 
 
+conan.tools.files.rm()
+----------------------
+
+.. code-block:: python
+
+    def rm(conanfile, pattern, folder, recursive=False)
+
+
+Remove the files following the ``pattern`` (fnmatch) from the specified ``folder``.
+
+.. code-block:: python
+
+    from conan.tools.files import rm
+
+    rm(self, "*.tmp", self.build_folder, recursive=True)
+
+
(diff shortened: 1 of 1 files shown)


czoido (approved):

reference/conanfile/tools/files/basic.rst
```diff
@@ -138,6 +138,31 @@ Parameters:
...
+    from conan.tools.files import rm
+
+    rm(self, "*.tmp", self.build_folder, recursive=True)
+
+
+Parameters:
+    - **conanfile**: Conanfile object.
+    - **pattern** (Required): Pattern that the files to be removed have to match (fnmatch).
+    - **folder**: Folder to search/remove the files.
+    - **recursive**: If ``recursive`` is specified it will search in the subfolders.
```
franramirez688, line 162: Is it worth putting `Required` or `Optional` for every parameter? I mean I think it's weird to put only one of them.

```
    - **conanfile** (Required): Conanfile object.
    - **pattern** (Required): Pattern that the files to be removed have to match (fnmatch).
    - **folder** (Required): Folder to search/remove the files.
    - **recursive** (Optional): If ``recursive`` is specified it will search in the subfolders.
```

memsharded: Yeah, I would probably define as ``(optional)`` only the optional ones, and maybe not even that, as the signature is there and it is visible what arguments have defaults

memsharded (approved):

reference/conanfile/tools/files/basic.rst
lasote: Removed optional and required in the whole module.

PR #2628, opened 2022-06-30
Docs about configure_options, introduction to package id...

czoido:
With: https://github.com/conan-io/examples2/pull/34


tutorial/creating_packages/configure_options_settings.rst
```diff
@@ -0,0 +1,269 @@
+Configure settings and options in recipes
+=========================================
+
+.. important::
+
+    In this example, we retrieve the CMake Conan package from a Conan repository with
```
memsharded, line 6: It seems this is no longer true? This is not requiring the ``cmake`` conan package, or maybe it is just not seen in excerpts, but it is there, as continuation of previous exercises?

tutorial/creating_packages/configure_options_settings.rst
```diff
@@ -0,0 +1,269 @@
...
+* Collect the settings and options applied, along with some information about the
+  requirements and calculate the hash for the corresponding package ID.
+
+* If that package ID matches one of the packages stored in the local Conan cache Conan
+  will use that. If not, and we have any Conan remote configured, it will search for a
+  package with that package ID in the remotes.
+
+* If that calculated package ID does not exist in the local cache and remotes, Conan will
+  try to build that package from sources (this depends on the value of the ``--build``
+  argument). This build will generate a new package ID in the local cache.
```
memsharded, line 175:
```suggestion
* If that calculated package ID does not exist in the local cache and remotes, Conan will
  fail with a "missing binary" error message, or will try to build that package from sources (this depends on the value of the ``--build`` argument). This build will generate a new package ID in the local cache.
```

First, comment the "default" Conan behavior, which is erroring.

tutorial/creating_packages/configure_options_settings.rst
```diff
@@ -0,0 +1,269 @@
...
+package, and the second one, regardless of the different value of the fPIC option, said we
+already had the ``2a899fd0da3125064bf9328b8db681cd82899d56`` package installed.
+
+C libraries
+^^^^^^^^^^^
+
+There are other typical cases where you want to delete certain settings. Imagine that you
+are packaging a C library. When you build this library, there are settings like the
+compiler C++ standard (``settings.compiler.cppstd``) or the standard library used
+(``self.settings.compiler.libcxx``) that won't affect the resulting binary at all. Then it
```
memsharded, line 230: This goes beyond the ``package_id``. When used in ``configure()`` you are also removing them from all the toolchain, and the build system integrations will do a different thing, because the C++ settings do not exist. 

Finally, it will not be part of the ``package_id``, but the main different with respect to being removed in ``package_id()`` is that they are discarded from the build process too

memsharded (commented):
The image contains 2 identical package_ids, which I guess are expected to be different.

tutorial/creating_packages/configure_options_settings.rst
czoido: Yes, It does not require CMake but fmt, I change CMake for fmt 👍

tutorial/creating_packages/configure_options_settings.rst
czoido: I can add a note. I did not want to give too many details so that users start to digest the concepts little by little...

lasote (approved):

PR #2634, opened 2022-07-05
Note that install's `-if` is now `-of`

prince-chrismc:


migrating_to_2.0/commands.rst
```diff
@@ -20,6 +20,7 @@ Remember that in Conan 1.X you have to specify the build profile or activate the
 
     $ conan install . [--name=mylib] [--version=1.0] [-pr:b=build_profile] [-pr:h=host_profile]
 
+In addition the `--install-folder` has been replaced with `--output-folder`.
```
memsharded, line 23:
```suggestion
In addition the `--install-folder` has been replaced with `--output-folder`. You might need to provide both arguments in Conan 1.X as some legacy generated files (``conaninfo.txt``, ``conanbuildinfo.txt``, etc) are not affected by ``--output-folder``.
```

memsharded (approved):

migrating_to_2.0/commands.rst
memsharded: ``--output-folder=something`` doesn't work?

prince-chrismc: It does not work the same way as `-if` in 1.50 which was my expectation from your comment

memsharded: Yes, it is different. It will provide a "base" output folder, and then the ``layout()`` will still define there the relative layout of the folders, but it is not the "final" destination folder of the artifacts

PR #2643, opened 2022-07-11
Add docs for check_max_cppstd and valid_max_cppstd

MathPlayer:
This pull request would add documentation for the tooling to check the max cppstd, as indicated by conan-io/conan#11080

### DIFF
--- a/tutorial/creating_packages/preparing_the_build.rst
+++ b/tutorial/creating_packages/preparing_the_build.rst
@@ -94,8 +94,7 @@ As you can see:
 * Based on the value of the ``with_fmt`` option:
 
     - We install or not the ``fmt/8.1.1`` Conan package.
-    - We require or not a minimum and a maximum C++ standard as the *fmt* library requires at least C++11 and our
-      settings support at most C++14.
+    - We require or not a minimum and a maximum C++ standard as the *fmt* library requires at least C++11 and it will not compile if we try to use a standard above C++14 (just an example, *fmt* can build with more modern standards)
     - We inject the ``WITH_FMT`` variable with the value ``True`` to the :ref:`CMakeToolchain<conan-cmake-toolchain>` so that we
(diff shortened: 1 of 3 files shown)


MathPlayer:
I am assuming the docs pipeline will start working after conan-io/conan#11610 gets merged.

czoido (approved):

tutorial/creating_packages/preparing_the_build.rst
```diff
@@ -93,7 +94,8 @@ As you can see:
 * Based on the value of the ``with_fmt`` option:
 
     - We install or not the ``fmt/8.1.1`` Conan package.
-    - We require or not a minimum C++ standard as the *fmt* library requires at least C++11.
+    - We require or not a minimum and a maximum C++ standard as the *fmt* library requires at least C++11 and our
```
memsharded, line 97: Some clarification might be needed here: it is not that "our settings support at most C++14", as that could mean "our Conan settings like ``compiler.version=xxx`` compiler does not support C++17. This kind of check shouldn't be done in recipes.
There is a new ``profile_plugin.py`` plugin that can be used to test and check the validity of our settings, and it could be used to raise if some compiler doesn't support some C++ standard (I think the default implementation of the plugin already does that?)

So maybe this sentence should be clarified: "and our build system or source code would break if tried to compile with a C++ standard higher than 14"

tutorial/creating_packages/preparing_the_build.rst
```diff
@@ -93,7 +94,8 @@ As you can see:
...
+      settings support at most C++14.
```
memsharded, line 98:
```suggestion
    - We require or not a minimum and a maximum C++ standard as the *fmt* library requires at least C++11 and it will not compile if we try to use a standard above C++14 (just an example, *fmt* can build with more modern standards)
```

PR #2658, opened 2022-07-25
[bug] get_cpus is not supporting cgroup2 #11635

MariaMozgunova:
Docs for: https://github.com/conan-io/conan/pull/11667

Documents the changes made for the cgroupv2 support in the `cpu_count()` function.


SSE4 (approved):

memsharded (changes requested):
This documentation is not completely in the right place, because the changes were done to the new tool in ``from conan.tools``, no in this one which is the legacy ``from conans`` one.

It needs to be documented in https://docs.conan.io/en/latest/reference/conanfile/tools/build.html, as the ``build_jobs`` helper and the new ``conf``.

This is not this PR fault, but this was missing before, sorry about that.

memsharded (approved):

PR #2670, opened 2022-07-28
ported docs from 1.51 for 2.0

lasote:


reference/conanfile/methods.rst
```diff
@@ -128,3 +128,35 @@ direct
 
 If the dependency is a direct one, that is, it has explicitly been declared by the current
 recipe, or if it is a transitive one.
+
+
+validate_build()
```
lasote, line 133: The reference of methods is a mess. But I prefered to copy that to not lose it when we improve it.

memsharded (approved):

PR #2662, opened 2022-07-27
layout folders.subproject new feature

memsharded:
Docs for https://github.com/conan-io/conan/pull/11556/

PR #2682, opened 2022-07-29
docs: missing folders available in `layout()`

prince-chrismc:
I wanted to use the new feature in 1.51 for `self.folders.subproject` but the docs were missing. I did my best but I am not sure what more I could write.

I also added `root` since that was missing.

***

I would love a way to link `cmake_layout` to https://github.com/conan-io/docs/pull/2662/files

https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmake_layout.html

https://docs.conan.io/en/latest/developing_packages/package_layout.html?highlight=multiple%20subprojects#example-multiple-subprojects


reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
...
   build folder and when running the :command:`conan imports` command it will be relative to the current working directory.
 
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
+
+- **self.folders.root** (Defaulted to None): Specifies a parent directory where the sources, generators, etc are located specifically when the ``conanfile.py`` is located in a spereated subdirectory.
+
+Available since: `1.51.0 <https://github.com/conan-io/conan/releases>`_
+
+- **self.folders.subproject** (Defaulted to None): Specifies a subfolder where to the ``conanfile.py`` relative  to the project root. This is particulary useful for :ref:`package_layout<example_multiple_subprojects>`.
+
```
prince-chrismc, line 1519: Would this be a good spot to high light

https://docs.conan.io/en/latest/developing_packages/package_layout.html

***

 I missed https://docs.conan.io/en/latest/reference/conanfile/tools/layout.html has link to package layout too 😕

reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
   method in a ``conanfile.py``. In the cache, when running the :command:`conan create`, this subfolder will be relative to the root
   build folder and when running the :command:`conan imports` command it will be relative to the current working directory.
 
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
```
franramirez688, line 1511: I think we could start adding the exact tag to each one:

```suggestion
Available since: `1.46.0 <https://github.com/conan-io/conan/releases/tag/1.46.0>`_
```

reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
...
+
+- **self.folders.root** (Defaulted to None): Specifies a parent directory where the sources, generators, etc are located specifically when the ``conanfile.py`` is located in a spereated subdirectory.
+
+Available since: `1.51.0 <https://github.com/conan-io/conan/releases>`_
```
franramirez688, line 1515:
```suggestion
Available since: `1.51.0 <https://github.com/conan-io/conan/releases/tag/1.51.0>`_
```

reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
   method in a ``conanfile.py``. In the cache, when running the :command:`conan create`, this subfolder will be relative to the root
   build folder and when running the :command:`conan imports` command it will be relative to the current working directory.
 
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases>`_
+
+- **self.folders.root** (Defaulted to None): Specifies a parent directory where the sources, generators, etc are located specifically when the ``conanfile.py`` is located in a spereated subdirectory.
```
franramirez688, line 1513:
```suggestion
- **self.folders.root** (Defaulted to None): Specifies a parent directory where the sources, generators, etc., are located specifically when the ``conanfile.py`` is located in a separated subdirectory.
```

reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
...
+
+Available since: `1.51.0 <https://github.com/conan-io/conan/releases>`_
+
+- **self.folders.subproject** (Defaulted to None): Specifies a subfolder where to the ``conanfile.py`` relative  to the project root. This is particulary useful for :ref:`package_layout<example_multiple_subprojects>`.
```
franramirez688, line 1517:
```suggestion
- **self.folders.subproject** (Defaulted to None): Specifies a subfolder where the ``conanfile.py`` is relative to the project root. This is particularly useful for :ref:`package_layout<example_multiple_subprojects>`.
```

reference/conanfile/methods.rst
```diff
@@ -1508,6 +1508,14 @@ self.folders
...
+
```
franramirez688, line 1518: Perhaps, we could change the above:

```
Read about the feature :ref:`here<package_layout>`.
```
And put it in this line:
```
Read more information about layouts:
* :ref:`Package Layout <package_layout>`.
```

reference/conanfile/methods.rst
franramirez688: Apart from that, I would be more verbose with that reference, for instance, `package_layout_example_multiple_subprojects`.

Then, remember that you'll have to add that new one in `developing_packages/package_layout.rst` file before the `Example: Multiple subprojects` section:

```
              folder = os.path.join(self.recipe_folder, "..")
              copy(self, "*.txt", folder, self.export_sources_folder)
          
          def source(self):
              # we can see that the CMakeLists.txt is inside the source folder
              cmake = load(self, "CMakeLists.txt")
          def build(self):
              # The build() method can also access the CMakeLists.txt in the source folder
              path = os.path.join(self.source_folder, "CMakeLists.txt")
              cmake = load(self, path)



.. _package_layout_example_multiple_subprojects:

Example: Multiple subprojects
-----------------------------

Lets say that we have a project that contains multiple subprojects, and some of these subprojects need
to access some information that is at their same level (sibling folders). Each subproject would be 
a Conan package.
```

Finally, that part should include this new reference. For instance:

```
useful for :ref:`layouts with multiple subprojects<package_layout_example_multiple_subprojects>`
```

franramirez688 (commented):
Thank you, @prince-chrismc, for the PR! I just added a few minor comments that will hopefully help to improve the internal links and references.

reference/conanfile/methods.rst
prince-chrismc: Let me take a stab at this, the directions are really good starting point for a first time

reference/conanfile/methods.rst
prince-chrismc:
```suggestion
* :ref:`Package Layout <package_layout>`.

```

prince-chrismc: I personally like the specific links close to the text, someone might not think to scroll up and find the example?

franramirez688 (approved):

PR #2687, opened 2022-08-05
add "available since" to all functions & classes in conan.tools & fix few links

SpaceIm:
Very useful when you migrate recipe to conan v2 and you want to properly document `required_conan_version` or check whether it will be compatible with conan client installed in your CI.


franramirez688 (approved):

czoido (approved):
This will be very useful. Thanks a lot for the contribution @SpaceIm

PR #2700, opened 2022-08-16
Document `conan.tools.files.collect_libs()`

SpaceIm:
Maybe not appropriate in `basic` section of tools.files, but which one?

### DIFF
--- a/reference/conanfile/tools/files/basic.rst
+++ b/reference/conanfile/tools/files/basic.rst
@@ -356,3 +356,46 @@ Example:
     
         def source(self):
             data = self.conan_data["sources"]["mydata"]
+
+
+conan.tools.files.collect_libs()
+--------------------------------
+
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases/tag/1.46.0>`_
+
+.. code-block:: python
+
+    def collect_libs(conanfile, folder=None)
+
+Returns a sorted list of library names from the libraries (files with extensions *.so*, *.lib*, *.a* and *.dylib*) located inside the
+``conanfile.cpp_info.libdirs`` (by default) or the **folder** directory relative to the package folder. Useful to collect not
+inter-dependent libraries or with complex names like ``libmylib-x86-debug-en.lib``.
+
+.. code-block:: python
+
+    from conan.tools.files import collect_libs
+
+    def package_info(self):
+        self.cpp_info.libdirs = ["lib", "other_libdir"]  # Default value is 'lib'
+        self.cpp_info.libs = collect_libs(self)
+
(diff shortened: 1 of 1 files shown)


reference/conanfile/tools/files/basic.rst
```diff
@@ -356,3 +356,46 @@ Example:
...
+--------------------------------
+
+Available since: `1.46.0 <https://github.com/conan-io/conan/releases/tag/1.46.0>`_
+
+.. code-block:: python
+
+    def collect_libs(conanfile, folder=None)
+
+Returns a sorted list of library names from the libraries (files with extensions *.so*, *.lib*, *.a* and *.dylib*) located inside the
+``conanfile.cpp_info.libdirs`` (by default) or the **folder** directory relative to the package folder. Useful to collect not
```
prince-chrismc, line 371: I am not sure I understand the "folder directory relative to the package folder"

Does this mean `lib/` `bin/` dirs in the `conanfile.packacage_folder`?

reference/conanfile/tools/files/basic.rst
```diff
@@ -356,3 +356,46 @@ Example:
...
+.. code-block:: shell
+
+    -rwxr-xr-x libmath.1.0.0.dylib
+    lrwxr-xr-x libmath.1.dylib -> libmath.1.0.0.dylib
+    lrwxr-xr-x libmath.dylib -> libmath.1.dylib
+
+**Parameters:**
+    - **conanfile** (Required): A ``ConanFile`` object to get the ``package_folder`` and ``cpp_info``.
+    - **folder** (Optional, Defaulted to ``None``): String indicating the subfolder name inside ``conanfile.package_folder`` where
+      the library files are.
```
franramirez688, line 396: I think these lines should go just below the function definition above:

```
.. code-block:: python
    def collect_libs(conanfile, folder=None)

**Parameters:**
    - **conanfile** (Required): A ``ConanFile`` object to get the ``package_folder`` and ``cpp_info``.
    - **folder** (Optional, Defaulted to ``None``): String indicating the subfolder name inside ``conanfile.package_folder`` where
      the library files are.
```
Then, whenever you name the arguments later, it's clear enough to know that it's referring to the arguments already mentioned.

franramirez688: Fair enough. Indeed, we could change all of them in a future PR to keep the consistency or even in Conan 2.x. In my opinion, I prefer to see all the argument definitions just after mentioning the function.

franramirez688 (approved):
It looks good!
I think it'd be great to point to `master` instead of `develop` as it should have been here since 1.46.0. Anyway, you can keep it as it is right now, we'll manage that 👍

PR #2707, opened 2022-08-24
Update conf.py

AA1999:
Fix errors related to lack of language while building the docs

Without conf.py having a defualt language sphinx shows this:

```py
sphinx-build -W -b html -d _build/doctrees   . _build/html
Running Sphinx v5.1.1

Warning, treated as error:
Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).
make: *** [Makefile:57: html] Error 2
```

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -94,7 +94,7 @@ author = u'The Conan team'
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:


uilianries:
@AA1999 please, try again. The CLA assistant, sometimes, is down. Also, you can re-check you CLA by clicking on https://cla-assistant.io/check/conan-io/docs?pullRequest=2707

uilianries:
@AA1999 YOu running under a VPN or proxy? I can access without problems here. Maybe you should try another web browser due networking configuration. I'm using Firefox, btw.

AA1999:
I tried both Firefox and chrome without any VPN
The result was the same.

Proof:

image
image

uilianries:
@AA1999 Thank you for trying. We gonna check it and try to find some workaround. Thanks!

uilianries:
@AA1999 We checked and it seems be related to your location. The CLA Assistant server should located in some country which is blocking your access. Unfortunately, we can not accept a PR without a CLA. However, you can use a VPN to change your location, then sign the CLA.

If you don't have a VPN service, you would recommend https://www.vpnbook.com/ which is free and I used few times.

Sorry for the inconvenient, but it's out of our scope.

AA1999:
I did use a VPN initially but it was that first issue.

I'm using OpenVPN for that matter which is pretty good.

However for some reason it doesn't work sadly.
Even tho I checked website cache with browser and it doesn't seem to load regardless of that.

uilianries:
@AA1999 I see ... I tried to find more related issues on cla-assistant project (https://github.com/cla-assistant/cla-assistant), but so far, only timeout errors were related to internal errors, but all temporary. I found some errors where people could not reach the website and they changed the ISP, like this one: https://github.com/cla-assistant/cla-assistant/issues/440

As last try, I would suggest using Tor browser (https://www.torproject.org/download/), so you could proxy another country. I guess worth a try. Meanwhile, I'll open an issue to the upstream, asking for support.

uilianries:
@AA1999 please, add more information if you find some valuable detail, like web browser version and other web browser that you used and your operate system. 

https://github.com/cla-assistant/cla-assistant/issues/933

uilianries:
@AA1999 Thank you for keeping trying!

Issue #2481, opened 2022-03-31
Document new compatibility() recipe method


czoido:
We have the docs for https://github.com/conan-io/conan/pull/10868 pending.
As said in the PR, we should wait to 1.48 to document in case the interface changes.

SSE4:
it's 1.51 released already, can we document it now?

czoido:
> it's 1.51 released already, can we document it now?

Yes, we are documenting this for 1.52 docs

PR #2727, opened 2022-08-30
Start documenting compatibility

czoido:
Closes: https://github.com/conan-io/docs/issues/2481

### DIFF
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -1245,6 +1245,32 @@ in any other OS will be the standard one, as if the ``build_id()`` method was no
 the build folder will be wiped at each :command:`conan create` command and a clean build will
 be done.
 
+.. _method_compatibility:
+
+compatibility()
+---------------
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+Available since Conan `1.47.0 <https://github.com/conan-io/conan/releases/tag/1.47.0>`_
+
+This method can be used in a *conanfile.py* to define packages that are compatible between
+each other. If there are not binaries available for the requested settings and options
+this mechanism will retrieve the compatible packages' binaries if they exist.  The method
+should return a list of compatible configurations. For example, if we want that binaries
+built with gcc versions 4.8, 4.7 and 4.6 are considered compatible with the ones compiled
+with 4.9 we could declare the ``compatibility()`` like this:
+
+..  code-block:: python
+
+    def compatibility(self):
+        if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+            return [{"settings": [("compiler.version", v)]}
+                    for v in ("4.8", "4.7", "4.6")]
+
+
 .. _method_deploy:
 
 deploy()
--- a/reference/conanfile/methods.rst
+++ b/reference/conanfile/methods.rst
@@ -1270,6 +1270,22 @@ with 4.9 we could declare the ``compatibility()`` like this:
             return [{"settings": [("compiler.version", v)]}
                     for v in ("4.8", "4.7", "4.6")]
 
+The format of the list returned is as shown below:
+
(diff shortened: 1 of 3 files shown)


reference/conanfile/methods.rst
```diff
@@ -1245,6 +1245,32 @@ in any other OS will be the standard one, as if the ``build_id()`` method was no
...
+should return a list of compatible configurations. For example, if we want that binaries
+built with gcc versions 4.8, 4.7 and 4.6 are considered compatible with the ones compiled
+with 4.9 we could declare the ``compatibility()`` like this:
+
+..  code-block:: python
+
+    def compatibility(self):
+        if self.settings.compiler == "gcc" and self.settings.compiler.version == "4.9":
+            return [{"settings": [("compiler.version", v)]}
+                    for v in ("4.8", "4.7", "4.6")]
```
franramirez688, line 1271: Perhaps, it's worth mentioning:

* The format of the return. List of different dict of settings? Does each dict have only one field?
* Are `options` allowed?

franramirez688 (approved):
I think it looks good! Anyway, have a look at my comment below to discuss if it could be necessary to add more details about the function.

SSE4:
I think it deserves some further explanation.
as I can see, there are now, at least 3 different ways to express package compatibility:
1. [package_id/defining ABI compatibility](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html)
```

    def package_id(self):
        if self.settings.compiler.cppstd in ["14", "17"]
            self.info.settings.compiler.cppstd = "14/17"
```
2. [compatible packages](https://docs.conan.io/en/latest/reference/conanfile/methods.html#compatible-packages)
```
    def package_id(self):
        if self.settings.compiler == "gcc" and self.settings.compiler.cppstd == "14":
            compatible_pkg = self.info.clone()
            compatible_pkg.settings.compiler.cppstd = "17"
            self.compatible_packages.append(compatible_pkg)
```
3. [compatibility method](https://github.com/conan-io/conan/pull/10868)
```
    def compatibility(self):
         if self.settings.compiler == "gcc" and self.settings.compiler.cppstd == "14":
            return [{"settings": [("compiler.cppstd", "17")]]
```

for me personally, and I believe for many readers, the following points are still unclear:
- is any of them recommended over others? if I have a choice, which approach should I prefer and why?
- are they mutually exclusive, or could be used simultaneously in the same recipe?
- if all 3 of them are defined in the same recipe, but in different ways, which one takes a priority during the actual graph resolution?

franramirez688 (approved):

czoido:
@SSE4,

I have extended the explanations and I think now it should be clear enough. Regarding your comments

> I think it deserves some further explanation. as I can see, there are now, at least 3 different ways to express package compatibility:
> 
> 1. [package_id/defining ABI compatibility](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html)
> 2. [compatible packages](https://docs.conan.io/en/latest/reference/conanfile/methods.html#compatible-packages)
> 3. [compatibility method](https://github.com/conan-io/conan/pull/10868)
> * is any of them recommended over others? if I have a choice, which approach should I prefer and why?
> * are they mutually exclusive, or could be used simultaneously in the same recipe?
> * if all 3 of them are defined in the same recipe, but in different ways, which one takes a priority during the actual graph resolution?

The difference between 1 and 2 is already explained in the docs in the introduction for [compatible packages](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#compatible-packages). They are different mechanisms, so now, we migrate from 2 to 3 that are equivalents.

Then it would not make sense to have both mechanisms (2 and 3) in a recipe, the docs now clearly state that you should move from one to the other.

Please, let me know if you have more questions, I'm merging this as we have to release 1.52 but we could improve this in another PR in the future. Thanks a lot for the feedback!

PR #2735, opened 2022-09-02
Add a mention to the V2 beta remote

prince-chrismc:
I was looking for this information I was not able to find it.

Let me know if there's a better place (this is just where I first looked)

### DIFF
--- a/conan_v2.rst
+++ b/conan_v2.rst
@@ -21,7 +21,7 @@ If you want to migrate to 2.0, there are several things you will need to change:
     the more relevant changes.
   - **General changes** not related to the recipes nor the Conan commands, "build profiles", lowercase references... etc.
 
-If you are looking for precompiled binaries, there is a very short list in a separate remote which can be added
+If you are looking for precompiled binaries, there is a very short list in a separate remote which can be added (please, check the `Conan 2.0 documentation <https://docs.conan.io/en/2.0/index.html>`_ for more information)
 
     .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


memsharded (approved):
The idea is that at this point, if users want to really try the beta, they should be reading the 2.0 beta docs, and not the 1.X docs?

I am fine with this, just wondering if this could be confusing, using trying to actively use 2.0, but following 1.X docs.

conan_v2.rst
```diff
@@ -21,6 +21,11 @@ If you want to migrate to 2.0, there are several things you will need to change:
     the more relevant changes.
   - **General changes** not related to the recipes nor the Conan commands, "build profiles", lowercase references... etc.
 
+If you are looking for precompiled binaires, there is a very short list in a seperate remote which can be added
```
czoido, line 24:
```suggestion
If you are looking for precompiled binaries, there is a very short list in a separate remote which can be added
```

czoido:
I think we may add a link to 2.0 docs here? https://docs.conan.io/en/2.0/index.html

conan_v2.rst
```diff
@@ -21,6 +21,11 @@ If you want to migrate to 2.0, there are several things you will need to change:
     the more relevant changes.
   - **General changes** not related to the recipes nor the Conan commands, "build profiles", lowercase references... etc.
 
+If you are looking for precompiled binaries, there is a very short list in a separate remote which can be added
+
```
czoido, line 25:
```suggestion
If you are looking for precompiled binaries, there is a very short list in a separate remote which can be added (please, check the `Conan 2.0 documentation <https://docs.conan.io/en/2.0/index.html>`_ for more information)

```

PR #2740, opened 2022-09-07
Start documenting build method

czoido:
Example: https://github.com/conan-io/examples2/pull/43


tutorial/creating_packages/build_packages.rst
```diff
@@ -1,2 +1,220 @@
...
+                self.requires("fmt/8.1.1")
+            self.test_requires("gtest/1.11.0")
+
+        ...
+
+        def generate(self):
+            tc = CMakeToolchain(self)
+            if self.options.with_fmt:
+                tc.variables["WITH_FMT"] = True
+            if not self.conf.get("tools.build:skip_test", default=False):
```
memsharded, line 63: We made this built-in in CMake integration, for the cmake standard BUILD_TESTING variable. It is good to use a custom one to explain, but maybe add a comment to let readers know.

tutorial/creating_packages/build_packages.rst
```diff
@@ -1,2 +1,220 @@
...
+                tc.variables["BUILD_TESTS"] = True
+            tc.generate()
+
+        def build(self):
+            cmake = CMake(self)
+            cmake.configure()
+            cmake.build()
+            if not self.conf.get("tools.build:skip_test", default=False):
```
memsharded, line 71: Maybe add a note that ``cmake.test()`` already incorporate this check?

tutorial/creating_packages/build_packages.rst
```diff
@@ -1,2 +1,220 @@
...
+    class helloRecipe(ConanFile):
+        name = "hello"
+        version = "1.0"
+
+        # Binary configuration
+        settings = "os", "compiler", "build_type", "arch"
+        options = {"shared": [True, False], "fPIC": [True, False]}
+        default_options = {"shared": False, "fPIC": True}
+
+        def build(self):
```
memsharded, line 205: This probably deserves adding the ``generate()`` method (at least pseudo-code), to illustrate that it will also generate different files

tutorial/creating_packages/build_packages.rst
```diff
@@ -1,2 +1,220 @@
...
```
memsharded, line 177: Add a note/warning, saying that patching in build() should only be done for very special cases that the code will be different, but clarify that it will make more challenging the local flow with ``conan install`` + native build, and probably for those cases it is better to try a ``conan build`` command that patches for you locally too.

memsharded (commented):
I think it is missing the tutorial style. There is no single ``conan`` command with the expected output. Why are we cloning the code if not executing commands? I would try to make things runnable by users.

czoido:
> 

Yes, that part is missing, this is just a draft of the topics covered, I'll add that part and the examples2

memsharded (approved):
These docs are fantastic

PR #2755, opened 2022-09-19
a bit of structure and content cmd_wrapper for extensions

memsharded:


reference/extensions/command_wrapper.rst
```diff
@@ -0,0 +1,33 @@
+.. _reference_extensions_command_wrapper:
+
+Command wrapper
+---------------
+
+The ``cmd_wrapper.py`` extension plugin is a Python scripts that receive the command line
```
czoido, line 6:
```suggestion
The ``cmd_wrapper.py`` extension plugin is a Python script that receives the command line
```

reference/extensions/command_wrapper.rst
```diff
@@ -0,0 +1,33 @@
...
+argument provided by ``self.run()`` recipe calls, and allows intercepting them and returning
+a new one. 
+
+This plugin must be located in the ``extensions/plugins`` cache folder, and can be installed
+with the ``conan config install`` command.
```
czoido, line 11: If the others can also be installed via conan config I would also mention it

czoido:
Pointing the PR to the release branch so it can be published right after merging.

reference/extensions.rst
```diff
@@ -0,0 +1,35 @@
...
+- Binary compatibility ``compatibility.py`` extension allows to write custom rules for
+  defining custom binary compatibility accross different settings and options
+- The ``cmd_wrapper.py`` extension allows to inject arbitrary command wrappers to any
+  ``self.run()`` recipe command invocation, which can be useful to inject wrappers as
+  parallelization tools
+- The package signing extension allows to sign and verify packages at upload and install time
+  respectively
+
+
+TODO: Put commands here?
```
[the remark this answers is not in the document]
memsharded: But aren't the commands an extension point? Any reason they shouldn't be here and not as a first level in the reference? I think classifying them as ``extensions`` in the first level makes sense, and helps quickly understanding the intent.

reference/extensions/command_wrapper.rst
memsharded: What do you mean?

reference/extensions/binary_compatibility.rst
```diff
@@ -0,0 +1,24 @@
...
+It has some built-in logic implemented, but can be customized.
+
+
+.. code-block:: python
+
+    def compatibility(conanfile):
+        result = []
+        if conanfile.settings.build_type == "Debug":
+            result.append({"settings": [("build_type", "Release")]})
+        return result
```
SSE4, line 17: will it work in conjunction to `built-in` logic, or completely override it? if second, how to invoke built-in logic from my own `compatibility.py` plug-in?

reference/extensions/command_wrapper.rst
```diff
@@ -0,0 +1,33 @@
...
+
+For example:
+
+.. code-block:: python
+
+    def cmd_wrapper(cmd):
+        return 'echo "{}"'.format(cmd)
+
+Would just intercept the commands and display them to terminal, which means that all commmands
+in all recipes ``self.run()`` will not execute, but just be echoed.
```
SSE4, line 21: that's `--dry-run`, more or less, but what if I want both execute and display them into the terminal?

reference/extensions/command_wrapper.rst
```diff
@@ -0,0 +1,33 @@
...
+
+A more common use case would be the injection of a parallelization tools over some commands,
+which could look like:
+
+.. code-block:: python
+
+    def cmd_wrapper(cmd):
+        # lets paralellize only CMake invocations
+        if cmd.startswith("cmake"):
+            return 'parallel-build "{}"  --parallel-argument'.format(cmd)
```
SSE4, line 31: it will work only if `cmake` is executed by its name, but not if it's executed by full path (like `/usr/bin/cmake`)
also, how about providing some more real and useful example (e.g. passing `-j {cpu_count}` on the cmake command line?)

reference/extensions/package_signing.rst
```diff
@@ -0,0 +1,45 @@
...
+Package signing
+---------------
+
+This plugin, which must be located in the cache ``extensions/plugins/sign/sign.py`` file contains 2 methods:
+
+- The ``sign(ref, artifacts_folder, signature_folder)`` executes for every recipe and package that is to
+  be uploaded to a server. The ``ref`` is the full reference to the artifact, it can be either a recipe 
+  reference or a package reference. The ``artifacts_folder`` is the folder containing the files to be 
+  uploaded, typically the ``conanfile.py``, ``conan_package.tgz``, ``conanmanifest.txt``, etc. The
+  ``signature_folder`` contains the folder in which the generated files should be written.
```
SSE4, line 12: is `signature_folder` uploaded to the server as well?
asking, cause in the opposite case, I want be able to upload and sign from one machine, and then download and verify signature on another machine

reference/extensions/binary_compatibility.rst
```diff
@@ -0,0 +1,24 @@
...
+  
+
+Some important rules:
+
+- The built-in ``compatibility.py`` is subject to changes in future releases. To avoid being updated
+  in the future, please remove the first comment ``# This file was generated by Conan.``
```
SSE4, line 23: this looks suspicious. comments should not influence on the behavior, it's counter-intuitive and breaks user-expectations (well, there are special comments in python to control encoding, for instance, but they have specific syntax, this example looks just as a regular comment).
if we want to prevent auto-update, why not use something more explicit (an attribute, or an entry in conf?)

reference/extensions.rst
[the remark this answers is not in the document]
SSE4: can we avoid putting TODOs in docs altogether? for reader, it doesn't bring any useful information, what should newcomer do encountering it?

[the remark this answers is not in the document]
czoido: > But aren't the commands an extension point? Any reason they shouldn't be here and not as a first level in the reference? I think classifying them as `extensions` in the first level makes sense, and helps quickly understanding the intent.

For me it's ok to have the custom commands here or where they currently are, I think both are intuitive to find as long as here there's a link to the other or the opposite.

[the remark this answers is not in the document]
memsharded: It is much better to have docs with a TODO than not having them. They are beta docs.

reference/extensions/command_wrapper.rst
memsharded: That is up to the user, we don't enter here all the possible use cases, the interface is super simple and clear enough.

reference/extensions/package_signing.rst
memsharded: Yes, it is uploaded to server.

reference/extensions/binary_compatibility.rst
memsharded: There are a bunch of files that are generated by Conan. Having a comment in them is very explicit, easy to read and understand when you open the files. Having a conf for every different file will be hidden and more complex.

reference/extensions/command_wrapper.rst
czoido: Sorry, nevermind, I was thinking about the commands and deploy folders, but they are not explained here.

reference/commands.rst
```diff
@@ -17,12 +17,3 @@ Creator commands
 ----------------
 
 
-
-Custom commands
```
czoido, line 21: I think that we should at least mention the custom commands from here and provide a link to the new location

memsharded: Definitely

memsharded: Done

czoido (approved):

PR #2766, opened 2022-09-23
Update how to build to Android

uilianries:
Recently, some users asked how to build for Android, using Docker images, and doing a quick look, docs.conan.io has an entire section about, but it does not enforce build context (dual profile) which is mandatory when cross-building.

This PR tries to make it clear, if you are building to Android and cross-building, use build context always.


uilianries:
Current view: 

Screenshot 2022-09-23 at 20-03-26 Android — conan 1 52 0 documentation

memsharded (approved):

PR #2768, opened 2022-09-29
CMakeToolchain generator allows harcoded CMake cli generator

uilianries:
Not documented, but a public parameter: `generator`

https://github.com/conan-io/conan/blob/1.52.0/conan/tools/cmake/toolchain/toolchain.py#L118

On CCI is not possible to use conf file, and some project only work with specific generators.

Master branch is target because is in production already.


prince-chrismc (approved):

prince-chrismc:
https://github.com/conan-io/conan/blob/b560c75b4d6bdf784d6835a553588eb67965a304/conan/tools/cmake/toolchain/toolchain.py#L122 👍

PR #2786, opened 2022-10-17
Enable announcement banner for "cooking with Conan" event

prince-chrismc:
Been a while since I've seen one, let me know if this is too much or just right (or too short).

image

### DIFF
--- a/_themes/conan/layout.html
+++ b/_themes/conan/layout.html
@@ -158,11 +158,11 @@
           {% include "breadcrumbs.html" %}
           <!--<select id="search" name="state" multiple="true" onchange="window.location = reldir + this.value;" style="width:90%"></select>
           <hr/>-->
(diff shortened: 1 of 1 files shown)


memsharded (approved):
Can be merged, you need to do the PR to master branch instead to be published.

prince-chrismc:
develop vs master is a messy slope 🙈

uilianries:
@prince-chrismc You can set to master branch. Need to click on edit button for the title, then the target branch will show up as a list box.

prince-chrismc:
There's other commits that need to be removed

memsharded:
Need to branch from master, cherry-pick or similar.

The problem is that this might release docs for unreleased Conan 1.X future version.

prince-chrismc:
Should look much better

czoido (approved):

PR #2789, opened 2022-10-23
Document min conan version of few classes, methods, functions & arguments

SpaceIm:
- `rm_safe()` method in `conanfile.settings` & `conanfile.options` since 1.53.0
- `conan.tools.apple.is_apple_os()` function since 1.51.3
- `conan.tools.apple.to_apple_arch()` function since 1.52.0
- `conan.tools.apple.XCRun` class since 1.53.0
- `targets` optional argument of `conan.tools.microsoft.MSBuild.build()` method since 1.52.0
- `run()` method of `conan.tools.scm.Git` class since 1.53.0


uilianries (approved):
LGTM

PR #2801, opened 2022-11-04
Fix autotools install on Windows

franramirez688:
Docs for https://github.com/conan-io/conan/pull/12193

### DIFF
--- a/reference/conanfile/tools/gnu/autotools.rst
+++ b/reference/conanfile/tools/gnu/autotools.rst
@@ -106,7 +106,7 @@ This is just an "alias" of ``self.make(target="install")``
 
 Parameters:
     - **args** (Optional, Defaulted to ``None``): List of arguments to use for the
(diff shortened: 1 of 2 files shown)


reference/conanfile/tools/gnu/autotools.rst
```diff
@@ -106,7 +106,7 @@ This is just an "alias" of ``self.make(target="install")``
 
 Parameters:
     - **args** (Optional, Defaulted to ``None``): List of arguments to use for the
-      ``make install`` call. By default an argument ``DESTDIR=self.package_folder`` is added to the
+      ``make install`` call. By default an argument ``DESTDIR=unix_path(self, self.package_folder)`` is added to the
```
czoido, line 109: I think we should add a link to the tool as it's not that evident what this does

franramirez688: Oh, absolutely!

PR #2803, opened 2022-11-04
win_bash_run

memsharded:


reference/conanfile/attributes.rst
```diff
@@ -1809,6 +1809,32 @@ When ``True`` it enables the new run in a subsystem bash in Windows mechanism. :
...
+win_bash_run
+------------
+
+.. warning::
+
+    This is an **experimental** feature subject to breaking changes in future releases.
+
+Available since: `1.54.0 <https://github.com/conan-io/conan/releases/tag/1.54.0>`_
+
+When ``True`` it enables running commands in the ``"run"`` scope, to run them inside a bash shell.
```
SSE4, line 1825: how do I enable it for an another scope?

migrating_to_2.0/recipes.rst
```diff
@@ -949,8 +949,12 @@ You need to configure how to run the commands with two config variables:
 
     - **tools.microsoft.bash:subsystem**:  Possible values: 'msys2', 'msys', 'cygwin', 'wsl' and 'sfu'
     - **tools.microsoft.bash:path** (Default "bash"): Path to the shell executable.
+    - **tools.microsoft.bash:active** (Default "None"): Used to define if Conan is already running inside a subsystem (Msys2) terminal.
 
 Any command run with ``self.run``, if ``self.win_bash == True`` will run the command inside the specified shell.
+Any command run with ``self.run(..., scope="run")`` if ``self.win_bash_run == True`` will run that command inside the shell.
+In both cases running explicitly in the bash shell only happens if ``tools.microsoft.bash:active`` is not True, because
+when it is True, it means that COnan is already running inside the shell.
```
SSE4, line 957:
```suggestion
when it is True, it means that Conan is already running inside the shell.
```

reference/conanfile/attributes.rst
memsharded: ``win_bash`` is that is used for the other scope ("build")

migrating_to_2.0/recipes.rst
```diff
@@ -949,8 +949,12 @@ You need to configure how to run the commands with two config variables:
```
czoido, line 949: should be _with three_ now

migrating_to_2.0/recipes.rst
```diff
@@ -949,8 +949,12 @@ You need to configure how to run the commands with two config variables:
 
     - **tools.microsoft.bash:subsystem**:  Possible values: 'msys2', 'msys', 'cygwin', 'wsl' and 'sfu'
     - **tools.microsoft.bash:path** (Default "bash"): Path to the shell executable.
+    - **tools.microsoft.bash:active** (Default "None"): Used to define if Conan is already running inside a subsystem (Msys2) terminal.
 
 Any command run with ``self.run``, if ``self.win_bash == True`` will run the command inside the specified shell.
+Any command run with ``self.run(..., scope="run")`` if ``self.win_bash_run == True`` will run that command inside the shell.
+In both cases running explicitly in the bash shell only happens if ``tools.microsoft.bash:active`` is not True, because
+when it is True, it means that Conan is already running inside the shell.
```
czoido, line 957: For me this part reads a bit confusing, maybe illustrating with what Conan calls for each case or some example would make this part a bit more clear?

reference/conanfile/attributes.rst
czoido: I think that's not completely clear, maybe we can add some clarification on that for `win_bash` and maybe adding here something like this is the equivalent of `win_bash` for the `run` scope ?

PR #2808, opened 2022-11-08
Fix a few typos in tools.gnu

agagniere:
including a sentence found in many pages :
```shell
grep "We encourage the usage of it to be prepared for Conan 2.0." --recursive . --files-with-matches | xargs -L1 sed 's|the usage of it|their usage|' --in-place
```


memsharded (approved):
Many thanks!

PR #2817, opened 2022-11-22
Fix cmake `build_script_folder` parameter

rpoisel:


creating_packages/package_repo.rst
```diff
@@ -51,7 +51,7 @@ Now let's have a look at *conanfile.py*:
 
         def build(self):
             cmake = CMake(self)
-            cmake.configure(source_folder="src")
+            cmake.configure(build_script_folder="src")
```
memsharded, line 54: this belongs to the ``from conan.tools.cmake import CMake`` helper, not the one documented in this section ``from conans import CMake``

memsharded (changes requested):

memsharded:
Hi @rpoisel 

Thanks for your contribution.
We are aware that the current state of docs at the moment is a bit messy, mixing things from the legacy integrations and the new, 2.0-ready ones. We are working hard on the 2.0 new docs, but the 1.X ones, we haven't manage to put enough time. Fortunately, Conan 2.0 is now very close, so hopefully all docs things will be much better soon.

rpoisel:
Thanks for the info and all what's there already! I am fairly new to conan.io and I was not sure whether my PR would make things better. :-) Looking forward to the 2.0 release as well :+1:

memsharded:
Yeah probably we can close it, but thanks anyway!

Don't hesitate to ask if you have any question.

PR #2820, opened 2022-11-28
[develop2] Commands reference docs

memsharded:
- Not all commands documented yet, but the structure created
- Several commands documented, the others will follow in other PRs (I'd like feedback on these first, before doing all)

Some practices to discuss:
- I am doing ``:hidden:`` TOCs, so I can do a small explanation of the command inline with the TOC
- Every subcommand in its own page, not one after the other (some subcommands are not that much related, and it reads worse IMO)
- I have started to add notes called "Best Practices", highlighted. Let's discuss if this is something we want


reference/commands.rst
```diff
@@ -13,13 +13,36 @@ and these :ref:`custom command examples <examples_extensions_custom_commands>`
...
+- :doc:`conan search <commands/search>`: Search packages matching a name
 
-Creator commands
-----------------
 
+.. toctree::
+   :caption: Creator commands
+   :maxdepth: 1
 
+   :conan upload: Upload packages from the local cache to a specified remote <commands/upload>
```
czoido, line 48:
```suggestion
.. toctree::
   :caption: Creator commands
   :maxdepth: 1
   :hidden:

   commands/upload

- :doc:`conan upload <commands/upload>`: Upload packages from the local cache to a specified remote
```
Maybe we can also add titles before the hidden toctrees?

reference/commands.rst
```diff
@@ -13,13 +13,36 @@ and these :ref:`custom command examples <examples_extensions_custom_commands>`
...
+   commands/profile
+   commands/remove
+   commands/remote
+   commands/search
 
 
+- :doc:`conan cache <commands/cache>`: Return the path of recipes and packages in the cache
+- :doc:`conan config <commands/cache>`: Manage Conan configuration (remotes, settings, plugins, etc)
+- :doc:`conan graph <commands/cache>`: Obtain information about the dependency graph without fetching binaries
+- :doc:`conan inspect <commands/inspect>`:
```
czoido, line 34:
```suggestion
- :doc:`conan inspect <commands/inspect>`: Inspect a conanfile.py to return the public fields
```

PR #2829, opened 2022-11-29
Docs for settings public in XCrun

czoido:


reference/conanfile/tools/apple.rst
```diff
@@ -457,8 +459,14 @@ Parameters:
...
   ``settings_target`` in the :ref:`cross-building section<cross_building>`.
 
-XCRun is a wrapper for the Apple **xcrun** tool used to get information for building. It
-has these properties:
+Public attributes and properties
+++++++++++++++++++++++++++++++++
+
+Public attributes:
+
+  - **settings**: Set of settings used to get the sdk.
```
franramirez688, line 467: Maybe, this description deserves a little example to avoid doubts about the possible values.

czoido: I have extended the explanation a bit, I can't think of an example more than getting the public attribute... I don't know if it's worth adding it... WDYT?

franramirez688: good enough! 👍

PR #2838, opened 2022-11-30
Fix docs after api changes

czoido:


SSE4:
@czoido shouldn't it cause an increment in minor version according to semver, as API has changed?
```
Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.
```
so, probably it should be marked as `2.1.0-beta6`?

memsharded (approved):

czoido:
> @czoido shouldn't it cause an increment in minor version according to semver, as API has changed?
> 
> ```
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.
> ```
> 
> so, probably it should be marked as `2.1.0-beta6`?

@SSE4 thanks for pointing that out. For the moment we are not versioning the API, we will not declare it stable until we release a couple of Conan 2.0 versions. From that point we will consider what to do with the API versioning.

PR #2844, opened 2022-12-02
Add missing methods to conan create/install order

SpaceIm:
`export()` was missing in `conan create`.
The list in `conan install` was outdated, many new methods were missing.

By the way, I'm surprised to not see `system_requirements()` in `conan install/create` when build of package is not triggered.


SSE4:
might be more methods missing, see https://github.com/conan-io/conan-center-index/pull/3614#discussion_r528179544
at least, I don't see `set_name` and `set_version`

SpaceIm:
These 2 methods can be added to `conan create` methods order, but in `conan install` there should be a clear separation between `conan install` against a local user conanfile.py and `conan install` against a reference.

prince-chrismc (approved):

reference/commands/consumer/install.rst
```diff
@@ -155,27 +155,33 @@ generators.
...
-3. ``requirements()``
-4. ``package_id()``
-5. ``package_info()``
-6. ``deploy()``
+3. ``layout()``
+4. ``requirements()``
+5. ``package_id()``
+6. ``validate()``
+7. ``package_info()``
+8. ``deploy()``
```
memsharded, line 163:
```suggestion
6. ``validate()``
7. ``validate_build()``
8. ``package_info()``
9. ``deploy()``
```

memsharded (approved):

reference/commands/consumer/install.rst
```diff
@@ -155,27 +155,34 @@ generators.
...
 2. ``configure()``
-3. ``requirements()``
-4. ``package_id()``
-5. ``package_info()``
-6. ``deploy()``
+3. ``layout()``
+4. ``requirements()``
+5. ``package_id()``
+6. ``validate()``
+7. ``validate_build()``
```
SpaceIm, line 162: @memsharded are you sure? Why validate_build() would be called when we don't build?

prince-chrismc: `--build missing` ?

SpaceIm: this list doesn't refer to --build missing, it's the list below.

memsharded: Completely sure, both methods are called sequentially: 
```python
      if hasattr(conanfile, "validate") and callable(conanfile.validate):
            with conanfile_exception_formatter(str(conanfile), "validate"):
                try:
                    conanfile.validate()
                    # FIXME: this shouldn't be necessary in Conan 2.0
                    conanfile._conan_dependencies = None
                except ConanInvalidConfiguration as e:
                    conanfile.info.invalid = str(e)

        if hasattr(conanfile, "validate_build") and callable(conanfile.validate_build):
            with conanfile_exception_formatter(str(conanfile), "validate_build"):
                try:
                    conanfile.validate_build()
                except ConanInvalidConfiguration as e:
                    # This 'cant_build' will be ignored if we don't have to build the node.
                    node.cant_build = str(e)
```

SpaceIm: ok, it's just weird to have a `validate_build()` method name which is called even when we don't build.

memsharded: It is perfectly reasonable to know if a package can be built or not with the current configuration, even if you are not building it, for example in ``conan info`` (that never builts from source). It is very valuable information, very cheap to compute, it makes sense to have the information available.

PR #2846, opened 2022-12-05
Fix some of the minor typos found while reading the docs

RubenRBS:
These are some of the low hanging typos I've found today. I've committed them mostly to check that my setup is ok, and to learn workflow for these cases.

Should I now check if they are present in develop2 and merge there too? Or is the process different?

### DIFF
--- a/devtools/build_requires.rst
+++ b/devtools/build_requires.rst
@@ -262,7 +262,7 @@ Making tool_requires affect the consumers package-ID
     This subsection should be considered a workaround, not a feature, and it might have other side effects, that will not be fixed
     as this is not recommended production code.
 
-As discussed above, the ``tool_requires`` does not affect at all the package ID. As they will not be present at all when the ``package_id``
+As discussed above, the ``tool_requires`` do not affect at all the package ID. As they will not be present at all when the ``package_id``
 is computed, it cannot be part of it. It is possible that this might change in the future in Conan 2.0, but at the moment it is not.
 In the meantime, there is a possible workaround that might be used if this is very needed: using ``python_requires`` to point to the
 same ``tool_requires`` package. Something like:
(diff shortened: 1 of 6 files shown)


devtools/build_requires.rst
```diff
@@ -262,7 +262,7 @@ Making tool_requires affect the consumers package-ID
     This subsection should be considered a workaround, not a feature, and it might have other side effects, that will not be fixed
     as this is not recommended production code.
 
-As discussed above, the ``tool_requires`` do not affect at all the package ID. As they will not be present at all when the ``package_id``
+As discussed above, the ``tool_requires`` does not affect at all the package ID. As they will not be present at all when the ``package_id``
```
memsharded, line 265:
```suggestion
As discussed above, the ``tool_requires`` do not affect at all the package ID. As they will not be present at all when the ``package_id``
```

I think this is good, it means plural, there can be several ``tool_requires``, no one of them affects package_id. The plural is also implicit in later ``As they will not be present...``

devtools/build_requires.rst
```diff
@@ -306,7 +306,7 @@ as a ``tool_requires`` or ``test_requires``:
             self.test_requires(self.tested_reference_str)
 
 
-If for some reason, it is necessary to test the same package both as a regular require and a build_require,
+If for some reason, it is necessary to test the same package both as a regular require and a tool_require,
```
memsharded, line 309:
```suggestion
If for some reason, it is necessary to test the same package both as a regular require and a ``tool_require``,
```

memsharded (approved):

devtools/build_requires.rst
RubenRBS: You're right, sorry! Nice catch

memsharded (approved):

memsharded:
Some other hints:
- If you target the ``master`` branch, merged changes will go to the live docs "latest"
- Changes to ``develop`` branch goes to the next release docs (that will become latest)

Docs in Conan 2.0 are completely different, written from scratch. I don't think looking for 1:1 correspondence of those typos is worth, but feel free to do a quick check.

PR #2850, opened 2022-12-12
Improve Android example to allow user interaction

danicriss:
In the previous example of Android integration via Gradle (Groovy) the call was blocking and the output was shown only when the call finished. This had several problems:
- any interaction attempts from Conan were not visible to the user
- user had no way to input anything if Conan asked them to
- the output was lost if the user cancelled

Examples scenarios are
- Conan locked and requiring an unlock command
- using a private repo, Conan could've asked for credentials

From a user perspective, the calls to Gradle just blocked with no visible output. Even worse, if the user finally decided to cancel, all output was lost. Mitigation was to run the Conan command manually to see what was going on

The proposed edit uses Gradle's recommended way to call a task (see https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Exec.html). This pipes the Conan output directly to stdout (`System.out` if you will). It also pipes stdin (`System.in`) which is not piped by default

As a bonus, added a showcase of how to integrate the `conanInstall` task into the Android build, so it'll be called before any NDK builds happen

All the above is a replica of a production PR after we've integrated Conan and had to walk several users (well, developers) through the same issues (described above). Tested in our (closed source, unfortunately) project


memsharded (commented):
This looks good, but I don't know enough about Android to review, lets ask someone else.

Thanks very much for contributing this.

Deishelon:
LGTM! 👍

prince-chrismc:
Doing some research on this one, I discovered a complete example from @ppetraki in a gist from last year https://gist.github.com/ppetraki/ca8630cc272edcd6a881ed82aa30b559 which seems to be in the same spirit as https://docs.gradle.org/7.6/userguide/building_cpp_projects.html which also makes use of tasks... but I am not a Gradle user either

I am inclined to give this a tick

PR #2855, opened 2022-12-17
Fix incorrect links in commands.rst

maxtruxa:
Links for docs on `conan config` and `conan graph` currently point to `conan cache`.


RubenRBS (approved):
Oops nice catch, thanks!

memsharded (approved):
Thanks!

PR #2886, opened 2023-01-04
docs: update usage of `self.info` for validate and package_id

prince-chrismc:
follow up to https://github.com/conan-io/conan-center-index/pull/13896#discussion_r1061050561

### DIFF
--- a/migrating_to_2.0/recipes.rst
+++ b/migrating_to_2.0/recipes.rst
@@ -97,7 +97,7 @@ Settings
           ...
 
           def validate(self):
-              if self.info.settings.os == "Macos":
+              if self.settings.os == "Macos":
                   raise ConanInvalidConfiguration("Macos not supported")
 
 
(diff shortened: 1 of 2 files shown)


memsharded (approved):

samuel-emrys:
Does the following comment also need revision? https://github.com/conan-io/docs/blame/1b0a2a38d3ac766fa5ee6dff70ea230b1ec9bcca/migrating_to_2.0/recipes.rst#L247-L248

And perhaps some additional explanation of why we use `self.info.{settings,options}` instead of `self.{settings,options}` in the `package_id(self)` method?

https://github.com/conan-io/docs/blame/1b0a2a38d3ac766fa5ee6dff70ea230b1ec9bcca/migrating_to_2.0/recipes.rst#L685-L702

(I used blame because it's hard to get links to rst files, apparently, and can't comment out of diff).

Otherwise looks pretty good.

prince-chrismc:
the first lines are still "correct" you should not be checking info in validate, though you still can if you need to check compatibility for some reasons... I choose not to modify those lines since I could not find a better wording without writing an easy 🙊 

As for when to use `self.info` the other docs https://docs.conan.io/en/latest/reference/conanfile/methods.html?highlight=package_id%20method#package-id and https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html?highlight=package_id%20method#defining-a-custom-package-id still are correct 

Specifically https://docs.conan.io/en/latest/reference/conanfile/methods.html?highlight=package_id%20method#self-info says why, maybe it needs more language and better examples but I am not sure how that plays with the 2.0 docs which I am not familiar with yet

I am sure there will be more PRs to improve this most recent change 🤞

PR #2887, opened 2023-01-05
Add tools.build:linker_scripts conf variable

jwillikers:
Documentation for conan-io/conan#12854

### DIFF
--- a/reference/conanfile/tools/cmake/cmaketoolchain.rst
+++ b/reference/conanfile/tools/cmake/cmaketoolchain.rst
@@ -366,7 +366,9 @@ conf
 - ``tools.build:sharedlinkflags`` list of extra linker flags that will be appended to ``CMAKE_SHARED_LINKER_FLAGS_INIT``.
 - ``tools.build:exelinkflags`` list of extra linker flags that will be appended to ``CMAKE_EXE_LINKER_FLAGS_INIT``.
 - ``tools.build:defines`` list of preprocessor definitions that will be used by ``add_definitions()``.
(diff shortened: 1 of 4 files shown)


reference/conanfile/tools/cmake/cmaketoolchain.rst
```diff
@@ -366,6 +366,7 @@ conf
 - ``tools.build:sharedlinkflags`` list of extra linker flags that will be appended to ``CMAKE_SHARED_LINKER_FLAGS_INIT``.
 - ``tools.build:exelinkflags`` list of extra linker flags that will be appended to ``CMAKE_EXE_LINKER_FLAGS_INIT``.
 - ``tools.build:defines`` list of preprocessor definitions that will be used by ``add_definitions()``.
+- ``tools.build:linker_scripts`` list of linker scripts, each of which will be prepended with ``-T`` and appended to ``CMAKE_EXE_LINKER_FLAGS_INIT``.
```
jcar87, line 369: I would suggest adding a further clarification that this is to be used when the linker understands the `-T` option and can parse linker scripts. I know it should be implied if someone uses this, but won't hurt, since on other platforms the native linker does not operate with linker scripts.

jwillikers: @jcar87 I was thinking the same thing, but I didn't want to put it in the `CMakeToolchain` stuff only, since it may apply to Meson and Autotools as well. Should I describe this caveat in all three places?

RubenRBS: Yes, please do :)

jwillikers: Done.

czoido (approved):

jcar87 (approved):
Thanks @jwillikers !!

PR #2904, opened 2023-01-14
Typo fix: mixed up profiles

yowidin:
Host and build profiles were mixed up in the android example


memsharded (approved):
You are right, thanks very much for the fix!
Docs will be released together with next release.

PR #2908, opened 2023-01-17
[conan v2] docs: fix hook headings

prince-chrismc:
Hiding all these from this page 
image


SSE4:
why hiding? I think it's useful for navigation to show the structure of the document

prince-chrismc:
It's on the wrong page though and doesn't follow the normal flow everywhere else :) this just keeps in on the right page

uilianries (approved):

PR #2920, opened 2023-01-25
[develop2] versioning

memsharded:
PR to examples in https://github.com/conan-io/examples2/pull/62


tutorial/versioning.rst
```diff
@@ -0,0 +1,35 @@
...
+Then, it will be introduced how ``requires`` with version ranges can 
+help to automate updating to the latest versions.
+
+There are some situations when recipes or source code are changed, but the version of the
+package is not increased. For those situations, Conan uses automatic ``revisions`` to 
+be able to provide traceability and reproducibility of those changes.
+
+Lockfiles are a common mechanism in other package managers to be able to reproduce the same
+dependency graph later in time, even when new versions or revisions of dependencies are uploaded.
+Conan also provide lockfiles to be able to guarantee this reproducibility.
```
czoido, line 20:
```suggestion
Conan also provides lockfiles to be able to guarantee this reproducibility.
```

tutorial/versioning.rst
```diff
@@ -0,0 +1,35 @@
...
+First, explicit version updates and how to define versions of packages is explained.
+
+Then, it will be introduced how ``requires`` with version ranges can 
+help to automate updating to the latest versions.
+
+There are some situations when recipes or source code are changed, but the version of the
+package is not increased. For those situations, Conan uses automatic ``revisions`` to 
+be able to provide traceability and reproducibility of those changes.
+
+Lockfiles are a common mechanism in other package managers to be able to reproduce the same
```
czoido, line 18:
```suggestion
Lockfiles are a common mechanism in package managers to be able to reproduce the same
```
Maybe remove "other"?

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
+.. _tutorial_versioning_version_ranges:
+
+Version ranges
+==============
+
+In the previous section, we ended with several version of the ``pkg`` package.
```
czoido, line 6:
```suggestion
In the previous section, we ended with several versions of the ``pkg`` package.
```

czoido, line 6: Where is that section? :thinking_face:

tutorial/versioning/versions.rst
```diff
@@ -0,0 +1,219 @@
+.. _tutorial_versioning_versions:
+
+Versions
+========
+
+Let's start with a very simple recipe:
```
czoido, line 6: Maybe we can do a summary of what we are going to cover in this section before beginning the example?

czoido: Maybe the structure could be like...

Versions
======

Write a little summary... that we will explain how to put versions inrecipes and after that to automating versions and then what comes next, etc.

Versions in recipes
-------------------

.....

Automating versions
-------------------
.....
.....

czoido (dismissed):
Besides what I already commented in the code, I was thinking that maybe we can do the tutorial using `conan new ...` and modifying the resulting recipe so that users have a real recipe that works, maybe newbies find confusing to show only a small portion of a recipe that does not exist? And also we have to take all this code to the examples2 repo to pass the ci

memsharded:
> Besides what I already commented in the code, I was thinking that maybe we can do the tutorial using conan new ... and modifying the resulting recipe so that users have a real recipe that works, 

I did it this way because it quickly became noisy and slow to have to do full builds to check just a version number, as I repeat the process for several versions. And it allowed the tutorial to be self contained. 

I can definitely put it in the examples repo, but don't you think that using a simple, empty recipe is good to illustrate version resolution. This is going to be a very intense topic in the following sections, including lockfiles, revisions, etc, having to build full packages for illustrating these things seems a bit heavy to me.

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
...
+The ordering of versions when necessary (for example to decide which is the latest version in a version range)
+is done by comparing individually each dot-separated entity in the version, from left to right. Digits will be
+compared numerically, so 2 < 11, and entries containing letters will be compared alphabetically (even if they
+also contain some numbers).
+
+Similarly to the semver specification, Conan can manage **prereleases** and **builds** in the form: 
+``VERSION-prerelease+build``.
+Conan will also order pre-releases and builds according to the same rules, and each one of them can also
+contain an arbitrary number of items, like ``1.2.3-pre.1.2.1+build.45.a``.
+Note that the semver standard do not apply any ordering to builds.
```
RubenRBS, line 115:
```suggestion
Note that the semver standard does not apply any ordering to builds.
```
Also, this is what semver oficially says. It would be nice to clarify what Conan actually ends up doing

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
...
+Also, lower bounds and upper bounds in isolation are permitted, though they are generally not recommended
+under normal versioning schemes, specially the lower bound only. ``requires = "pkg/[>=1.0 <2.0]"`` will 
+include versions like 1.0, 1.2.3 and 1.9, but will not include 0.3, 2.0 or 2.1 versions.
+
+
+The tilde ``~`` operator can be used to define an "approximately" equal version range. ``requires = "pkg/[~1]"``
+will include versions 1.3 and 1.8.1, but will exclude versions like 0.8 or 2.0. Likewise
+``requires = "pkg/[~2.5]"`` will include 2.5.0 and 2.5.3, but exclude 2.1, 2.7, 2.8.
+
+The caret ``^`` operator is very similar to the tilde, but allowing variability over the last defined digig.
```
RubenRBS, line 136:
```suggestion
The caret ``^`` operator is very similar to the tilde, but allowing variability over the last defined digit.
```

tutorial/versioning/versions.rst
```diff
@@ -0,0 +1,219 @@
...
+
+    from conan import ConanFile
+
+    class AppRecipe(ConanFile):
+        name = "app"
+        version = "1.0"
+        requires = "pkg/1.5"
+
+
+This process, while it achieves very good reproducibility and traceability, it can be a bit tedious if we are
```
RubenRBS, line 217:
```suggestion
This process, while it achieves very good reproducibility and traceability, can be a bit tedious if we are
```

reference/conanfile/attributes.rst
```diff
@@ -394,12 +394,22 @@ requires
...
-          2.8          # equivalent to =2.8
-          ~=3.0        # compatible, according to semver
-          >1.1 || 0.8  # conditions can be OR'ed
+    .. list-table::
+        :widths: 25 25 50
+        :header-rows: 1
+
+        * - Expression
+          - Versions in range
+          - Versions outside range
```
SSE4, line 403:
```suggestion
          - Versions outside of range
```

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
...
+
+    class pkgRecipe(ConanFile):
+        name = "app"
+        requires = "pkg/1.0"
+
+Let's create ``pkg/1.0`` and install ``app``, to see it requires ``pkg/1.0``:
+
+.. code-block:: bash
+
+    $ conan remove pkg* -c
```
SSE4, line 30:
```suggestion
    $ conan remove "pkg*" -c
```
it's better to quote. some shells (e.g. zsh) substitute it too early.

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
...
+        pkg/1.2
+
+
+Semantic versioning
+-------------------
+
+The semantic versioning specification or `semver <https://semver.org/>`_, specifies that packages should
+be versioned using always 3 dot-separated digits like ``MAJOR.MINOR.PATCH``, with very specific meanings for each digit.
+
+Conan extends the semver specification to any number of digits, and also allows to include letters in it.
```
SSE4, line 101: does it allow only latin letters, or any letters in general?

tutorial/versioning/version_ranges.rst
```diff
@@ -0,0 +1,147 @@
...
+This was done because during 1.X a lot of experience and feedback from users was gathered, and it became evident
+than in C++ the versioning scheme is often more complex, and users were demanding more flexibility, allowing
+versions like ``1.2.3.a.8`` if necessary.
+
+The ordering of versions when necessary (for example to decide which is the latest version in a version range)
+is done by comparing individually each dot-separated entity in the version, from left to right. Digits will be
+compared numerically, so 2 < 11, and entries containing letters will be compared alphabetically (even if they
```
SSE4, line 108: for letters, is case considered? `1.2.3.a.8` and `1.2.3.A.8`, are they treated as the same version or different?

memsharded: package references are all lowercase, the second is not allowed.

tutorial/versioning/version_ranges.rst
memsharded: [a-z][A-Z], only latin letters, yes.

SSE4: then just `[a-z]`, if only lowercase are allowed

memsharded: yes, right, only a-z

RubenRBS (approved):

tutorial/versioning/conflicts.rst
```diff
@@ -50,10 +50,210 @@ Resolving conflicts
 
 Of course, the most direct and straightforward way to solve such a conflict is
 going to the dependencies ``conanfile.py`` and upgrading their ``requirements()``
-so they point now the tha same version. However this might not be practical in
-some cases, or it might be even impossible to fix the dependencies. For that
-case there could be different strategies:
+so they point now the same version. However this might not be practical in
```
SSE4, line 53:
```suggestion
so they point now to the same version. However this might not be practical in
```

tutorial/versioning/conflicts.rst
```diff
@@ -0,0 +1,259 @@
...
+.. code-block:: bash
+
+    $ git clone https://github.com/conan-io/examples2.git
+    $ cd examples2/tutorial/versioning/conflicts/versions
+
+In this folder we have a small project, consisting in several packages: ``matrix`` (a math library),
+``engine/1.0`` video game engine that depends on ``matrix/1.0``, ``intro/1.0``, a package implementing
+the intro credits and functionality for the videogame that depends on ``matrix/1.1`` and finally the
+``game`` recipe that depends simultaneously on ``engine/1.0`` and ``intro/1.0``. All these packages
+are actually empty, but they are enough to produce the conflicts.
```
czoido, line 20: Having graphviz support in the docs could be good for cases like this

tutorial/versioning/revisions.rst
```diff
@@ -0,0 +1,253 @@
+Revisions
+=========
+
+This sections introduces how doing modifications to a given recipe or source code without explicitly
+creating new versions, will still internally track those changes with a mechanism called revisions.
+
+
+Creating different revisions
+----------------------------
```
czoido, line 9: Maybe we should specify that all these revisions are recipe revisions through all the tutorial until we get to the package revisions part?
```suggestion
Creating different recipe revisions
----------------------------------
```

PR #2923, opened 2023-01-26
Fix name regex to be truly 50 characters only.

douppc:
## Abstract

The previous text of this help documentation provides a regex that actually allows a name of size 51 characters instead of what the rest of documentation claims is 50 characters. This new regex checks that it is appropriately 50 characters.

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -11,7 +11,7 @@ Attributes
 name
 ----
 This is a string, with a minimum of 2 and a maximum of 50 characters (though shorter names are recommended), that defines the package name. It will be the ``<pkgName>/version@user/channel`` of the package reference.
-It should match the following regex ``^[a-zA-Z0-9_][a-zA-Z0-9_\+\.-]{1,50}$``, so start with alphanumeric or underscore, then alphanumeric, underscore, +, ., - characters.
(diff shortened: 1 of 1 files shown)


reference/conanfile/attributes.rst
```diff
@@ -11,7 +11,7 @@ Attributes
 name
 ----
 This is a string, with a minimum of 2 and a maximum of 50 characters (though shorter names are recommended), that defines the package name. It will be the ``<pkgName>/version@user/channel`` of the package reference.
-It should match the following regex ``^[a-zA-Z0-9_][a-zA-Z0-9_\+\.-]{1,50}$``, so start with alphanumeric or underscore, then alphanumeric, underscore, +, ., - characters.
+It should match the following regex ``^[a-zA-Z0-9_][a-zA-Z0-9_\+\.-]{1,49}$``, so start with alphanumeric or underscore, then alphanumeric, underscore, +, ., - characters.
```
memsharded, line 14: Hi @douppc 
Thanks for contributing this. The actual code is:
```python
 _max_chars = 51
    _min_chars = 2
    _validation_pattern = re.compile("^[a-zA-Z0-9_][a-zA-Z0-9_\+\.-]{%s,%s}$"
                                     % (_min_chars - 1, _max_chars - 1))

    _validation_revision_pattern = re.compile("^[a-zA-Z0-9]{1,%s}$" % _max_chars)
```

So the regex is correct. However I don't think it is worth arguing over this, Conan 2.0 is almost there, and it relax a bit this assumption.

PR #2926, opened 2023-01-27
Mark all the old generators as deprecated

prince-chrismc:


SSE4:
can you specify a replacement for each deprecated generator? it would be easier for reader, than trying to find information in long migration guide.

memsharded (approved):

PR #2927, opened 2023-01-27
Update getting_started tutorial

alessiosacco:
The conan CLI has changed and the command is not valid anymore. The command has been verified against version 2.0.8b


memsharded (changes requested):
hi @alessiosacco 

Thanks for contributing.
But I am afraid you are mixing branches, that syntax is Conan 2.0 but you are sending a PR for the 1.X docs. You can find it in the ``develop2`` branch (and in https://docs.conan.io/en/2.0)

alessiosacco:
You are right, I was fooled by the documentation UI. https://docs.conan.io/en/latest/creating_packages/getting_started.html points to the documentation of 1.57, even though in the version selector latest is right next to 2.0.0. I expected that clicking on latest would bring me to the documentation of 2.0.0, but this is not the case. I will close this PR since the fix is not necessary.

SSE4:
@memsharded is it possible to fix `latest` to point to 2.x instead of 1.x?

memsharded:
> @memsharded is it possible to fix latest to point to 2.x instead of 1.x?

No, that can only happen when 2.0 is finally released

PR #2945, opened 2023-02-02
add note about `conan list --format=html`

SSE4:
related to: 
https://github.com/conan-io/conan/issues/13021
https://github.com/conan-io/conan/issues/13022

it needs some sort of disclaimer that `--format html` may work differently from `--format json` in some circumstances, e.g. showing a different amount of packages.

### DIFF
--- a/reference/commands/list.rst
+++ b/reference/commands/list.rst
@@ -29,7 +29,7 @@ conan list
 .. warning::
 
     In some circumstances, `--format html` might produce semantically different results from `--format json`, like
-    showing a different amount of packages. Use it on your own risk.
+    showing a different amount of packages (there is a bug `GitHub issue #13022 <https://github.com/conan-io/conan/issues/13022>). Use it on your own risk.
 
 
 The ``conan list`` command can list recipes and packages from the local cache, from the
--- a/reference/commands/list.rst
+++ b/reference/commands/list.rst
@@ -28,9 +28,10 @@ conan list
 
 .. warning::
 
-    In some circumstances, `--format html` might produce semantically different results from `--format json`, like
-    showing a different amount of packages (there is a bug `GitHub issue #13022 <https://github.com/conan-io/conan/issues/13022>`_). Use it on your own risk.
-
+    The html formatter (--format=html) is not fully implemented yet, so it will create an empty html file
+    with no packages information. This is `a reported bug
+    <https://github.com/conan-io/conan/issues/13022>`_ that happens in Conan <= 2.0-beta9
+    and is planned to be solved in 2.0-beta10
 
 The ``conan list`` command can list recipes and packages from the local cache, from the
 specified remotes or from both. This command uses a *reference pattern* as input. The
--- a/reference/commands/list.rst
+++ b/reference/commands/list.rst
@@ -29,7 +29,7 @@ conan list
 .. warning::
 
     In some circumstances, `--format html` might produce semantically different results from `--format json`, like
-    showing a different amount of packages (there is a bug `GitHub issue #13022 <https://github.com/conan-io/conan/issues/13022>). Use it on your own risk.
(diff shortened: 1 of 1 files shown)


reference/commands/list.rst
```diff
@@ -26,6 +26,11 @@ conan list
                             Remote names. Accepts wildcards
       -c, --cache           Search in the local cache
 
+.. warning::
+
+    In some circumstances, `--format html` might produce semantically different results from `--format json`, like
+    showing a different amount of packages. Use it on your own risk.
```
franramirez688, line 32: @SSE4 The `--format html` needs to be fixed, but it's not a problem about showing a different amount packages at all, the problem is that it's not showing anything because it's not working.

franramirez688 (changes requested):

reference/commands/list.rst
SSE4: that's exactly that I mean - it may show a different amount, e.g. zero packages, while `--format json` may show more than zero packages. so `not showing anything` falls into that category.

SSE4:
@czoido can we merge this one? right now there is no indication `--format html` may not work correctly in some cases.

SSE4:
> I am afraid this is not accurate. `--format=html` should show the same results as `json`. If not, it is a bug, and should be fixed. In fact, html is basically broken now, so yes, it is a bug. It can be added as such, linked to the github issue. But it is not a warning that they could display different results.

i don't see any contradiction: yes, it's a bug. yes, it should be fixed. yes, `--format=html` is broken in beta9. yes, it can display different result from `--format=json`.
why warning cannot be added? right now there is no way for user to be aware of that.
if warning cannot be added as such for some reason (I still don't get why documentation cannot reflect the reality - IMO it should describe the actual software behavior), please fix it and release beta 9.1 ASAP.

memsharded:
> why warning cannot be added? right now there is no way for user to be aware of that.

A warning CAN be added, saying there is a bug, linking the bug. The request for changes is because the current warning says the output might be different, but not because it is a bug, so users could think that it is expected that the output is different.

SSE4:
> A warning CAN be added, saying there is a bug, linking the bug. The request for changes is because the current warning says the output might be different, but not because it is a bug, so users could think that it is expected that the output is different.

okay, I've added note that there is a bug and link to the bug.

reference/commands/list.rst
```diff
@@ -26,6 +26,11 @@ conan list
                             Remote names. Accepts wildcards
       -c, --cache           Search in the local cache
 
+.. warning::
+
+    In some circumstances, `--format html` might produce semantically different results from `--format json`, like
+    showing a different amount of packages (there is a bug `GitHub issue #13022 <https://github.com/conan-io/conan/issues/13022>`_). Use it on your own risk.
```
czoido, line 32: I think it would be more accurate to say that is not implemented yet in <=2.0-beta9 and then remove the warning for the next release, because at least a very basic output will be there. I think it's a bit misleading to say that it may produce different results when the reality is that is not working at all because it has not been implemented yet.

SSE4: well, it's kinda working, as it produces an output file, and it's valid .html file that could be opened. e.g. there is no error message, crash, etc. it just doesn't include any data.
that's why it's so confusing, and I'd preferred if it just raised an error immediately for any `--format html` input instead.

czoido: What about something similar to this?

```
.. warning::

    The html formatter (--format=html) is not fully implemented yet, so it will create an empty html file
    with no packages information. This is `a reported bug
    <https://github.com/conan-io/conan/issues/13022>`_ that happens in Conan <= 2.0-beta9
    and is planned to be solved in 2.0-beta10
```

Besides that, this should be pointed to the _release/2.0.0-beta_ branch to be published in the docs or it will never be published if we fix it in the next release

SSE4: okay, I've changed a message and base.

PR #2948, opened 2023-02-02
[conan2] how to read docs

memsharded:


introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
 But an essential part of Conan is that many of those users will contribute back, creating an amazing and helpful community:
 
-- The https://github.com/conan-io/conan project has around 6K stars in Github and counts with contributions of almost 300 different users (this is just the client tool).
+- The https://github.com/conan-io/conan project has around 6.5K stars in Github and counts with contributions from more than 300 different users (this is just the client tool).
 - Many other users contribute recipes for ConanCenter via the https://github.com/conan-io/conan-center-index repo, creating packages for popular Open Source libraries, contributing many thousands of Pull Requests per year.
 - More than two thousands Conan users hang around the `CppLang Slack #conan channel <https://cppalliance.org/slack/>`_, and help responding to questions, discussing problems and approaches, making it one of the most active channels in the whole CppLang slack.
 - There is a Conan channel in `#include<cpp> discord <https://www.includecpp.org/discord/>`_.
 
 
+How to read this documentation
```
prince-chrismc, line 115:
```suggestion
Navigating the documentation
```

introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+------------------------------
+
+This documentation has very different sections:
+
+- The **tutorial** is an actual hands-on tutorial, with examples and real code, intended to be played sequentially from beginning to end,
+  running the exercises in your own computer. There is a conducting thread and exercises might depend on some previous explanations and code.
```
prince-chrismc, line 121: What is a "conducting thread" ?

introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+  This is the recommended approach for learning Conan.
+- The **examples** also contains hands-on, fully operational examples with code, aimed to explain some very specific feature, tool or behavior.
+  They do not have a conducting thread, they should be navigated by topic.
+- The **reference** is the ground truth for the interfaces of every public command, class, method, helper, API and configuration file that can
```
prince-chrismc, line 125:
```suggestion
- The **reference** is the source of truth for the interfaces of every public command, class, method, helper, API and configuration file that can
```

introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+  be used. It is not designed to be read fully, but to check for individual items when necessary.
+- The **knowledge** base contains things like the FAQ, a very important section about genral guidelines, good practices and bad practices, videos
+  from conference talks, etc.
+
+Features in this documentation might be labeled as:
+
+- **experimental**: This feature is released and can be used, but it is under active development and the interfaces, APIs or behavior might change
+  as a result of evolution, and this will not be considered breaking. If you are interested in these features you are encouraged to try them
+  and give feedback, because that is exactly what allows to stabilize them.
+- **preview**: The feature is release in preview mode, which means it aims to be as final and stable as possible. Users are encouraged to use them,
```
prince-chrismc, line 135: English is a touch confusing here

```suggestion
- **preview**: When a feature is released in preview mode, this means it aims to be as final and stable as possible. Users are encouraged to use them,
```

introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+  and the maintainers team will try not to break them unless necessary. But if necessary, they might change and break.
+- **deprecated**: This feature should no longer be used, and it will be fully removed in next major release. Other alternatives or approaches should
+  be used instead of it, and if using it, migrating to the other alternatives should be done as soon as possible. They will not be maintained or
+  get fixes.
+
+
```
prince-chrismc, line 141: I'd love a summary table for this 


| Feature Life Cycle | Breaking Changes | Encouraged Usage | Fixes |
|-------------------|--------------------|--------------------|------|
|**experimental**| ✅ | ⛔ | ✅ |
|**preview**| ✅ | ⛔ | ✅ |
|"stable"| ✅ | ⛔ | ✅ |
|**deprecated**| ✅ | ⛔ | ✅ |

introduction.rst
```diff
@@ -105,12 +106,40 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+- **experimental**: This feature is released and can be used, but it is under active development and the interfaces, APIs or behavior might change
+  as a result of evolution, and this will not be considered breaking. If you are interested in these features you are encouraged to try them
+  and give feedback, because that is exactly what allows to stabilize them.
+- **preview**: The feature is release in preview mode, which means it aims to be as final and stable as possible. Users are encouraged to use them,
+  and the maintainers team will try not to break them unless necessary. But if necessary, they might change and break.
+- **deprecated**: This feature should no longer be used, and it will be fully removed in next major release. Other alternatives or approaches should
+  be used instead of it, and if using it, migrating to the other alternatives should be done as soon as possible. They will not be maintained or
+  get fixes.
+
```
prince-chrismc, line 140: What about no labels? :)

prince-chrismc (commented):
Some quick feedback to improve the clarity :)

introduction.rst
memsharded: Yes, sorry, I don't know how to say it in english.
I want to say this is a story, the tutorial implements developers stories, with incremental complexity, in a natural and continuous way, like a section is continuation of the previous one, and there is like a guiding narrative over the whole tutorial.

introduction.rst
memsharded: We really need the labels, it is really very painful not able to release new features and get feedback, or not being able to ever change anything.

introduction.rst
memsharded: We should probably be careful about this, I don't love the "Encouraged usage" with a forbidden, even for experimental. We would encourage usage by some motivated users, but we don't want them to avoid it completely, otherwise we will never get real feedback. I like the idea of the table, but the names and the icons should be considered.

prince-chrismc: 💯 Agree, it deserves more thought, those where the first ones that came to mind :)

introduction.rst
prince-chrismc: Oops, that was not clear! 

You did not document what the meaning of "no label" for a feature should imply... I think users would knowing it has meaning rather then being just undocumented.

> If no label is noted on a feature then is should be considered as stable and any braking change will be delayed until a major release or alternative method is provided 

Just so there's something to reference back to

introduction.rst
prince-chrismc: | Feature Life Cycle | Breaking Changes | Usage | Fixes |
|-------------------|--------------------|--------------------|------|
|**experimental**| Likely | To provide feedback | Likely |
|**preview**| only if necessary | ??? | Yes |
|"stable"| No | Yes | Pfft. our code is perfect 🤩  |
|**deprecated**| No | Migrate away | No |

introduction.rst
memsharded: Oh, I see what you mean. Yes, I will add some clarification of that.

introduction.rst
```diff
@@ -105,12 +106,43 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
 
+Navigating the documentation
+------------------------------
+
+This documentation has very different sections:
+
+- The **tutorial** is an actual hands-on tutorial, with examples and real code, intended to be played sequentially from beginning to end,
+  running the exercises in your own computer. There is a conducting thread and exercises might depend on some previous explanations and code.
+  This is the recommended approach for learning Conan.
+- The **examples** also contains hands-on, fully operational examples with code, aimed to explain some very specific feature, tool or behavior.
```
SSE4, line 123:
```suggestion
- The **examples** also contain hands-on, fully operational examples with code, aimed to explain some very specific feature, tool or behavior.
```

introduction.rst
```diff
@@ -105,12 +106,43 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+  They do not have a conducting thread, they should be navigated by topic.
+- The **reference** is the source of truth for the interfaces of every public command, class, method, helper, API and configuration file that can
+  be used. It is not designed to be read fully, but to check for individual items when necessary.
+- The **knowledge** base contains things like the FAQ, a very important section about genral guidelines, good practices and bad practices, videos
```
SSE4, line 127:
```suggestion
- The **knowledge** base contains things like the FAQ, a very important section about general guidelines, good practices and bad practices, videos
```

introduction.rst
SSE4: I suppose it needs to explain what happens if breaking changes are released.
e.g., deprecated can only be removed if major version is bumped.
if experimental or preview API is changed in a breaking way, major version has to be bumped.
otherwise, if breaking API change is detected in minor or patch release, it should be immediately reverted.
according to https://semver.org/, more or less.

introduction.rst
SSE4: no label should probably be replaced with `stable` label.
it should be documented as never changing in a breaking way, and staying forever.

introduction.rst
```diff
@@ -105,12 +106,43 @@ Conan is being used in production by thousands of companies like TomTom, Audi, R
...
+  from conference talks, etc.
+
+Features in this documentation might be labeled as:
+
+- **experimental**: This feature is released and can be used, but it is under active development and the interfaces, APIs or behavior might change
+  as a result of evolution, and this will not be considered breaking. If you are interested in these features you are encouraged to try them
+  and give feedback, because that is exactly what allows to stabilize them.
+- **preview**: When a feature is released in preview mode, this means it aims to be as final and stable as possible. Users are encouraged to use them,
```
SSE4, line 135: `final and stable as possible` doesn't play well together with `if necessary, they might change and break`. second part basically means `unstable and subject to change`. it's either `stable` or `experiment`, and practically, right now definition of `preview` is not actually different from `experimental` (maybe, just with probability of changes, like `preview` has 1% chance to change in a breaking way, while `experimental` has 99% change to break, but both may break your code, ultimately).

introduction.rst
memsharded: No, we don't want to have to add a "stable" note in every page of the several hundreds pages of the docs. The stable label is the implicit, default one.
And stable features can indeed be declared "deprecated" (not removed, but deprecated)

introduction.rst
SSE4: that seems to be violation of [semver](https://semver.org/). for versions beyond 1.0.0 public API should be considered stable. and if public API changes in a breaking way, it should result into major bump.

introduction.rst
SSE4: yes, it could say `treat everything as "stable", if not specified otherwise`. and then goes definition of "stable".
I don't mean to add "stable" label explicitly everywhere. I just mean to describe default and "stable" here, as something that doesn't change in a breaking way, ever.

introduction.rst
memsharded: Not for experimental and preview, that is exactly what experimental and preview features are for. They do not form part of the public, stable API. Otherwise we couldn't evolve anything and become to a halt. Or doing a major release every month, which doesn't make sense.

SSE4: i am just saying to increment a version, if you break public API in an incompatible way.
to let your consumers know about that, and not to break their expectations about version number.
it's kinda annoying to see recipe has stopped to work in minor update - it should never happen in practice.

memsharded: If we had to increment a version every time a change is needed in some experimental feature, we would already be in Conan 238.1.3. Which doesn't make sense at all, and it is the whole reason to introduce **experimental** and **preview** features. We have not invented this, this is standard practice in SW products.

prince-chrismc: C++ expectations vs other ecosystems differ here so I think for C++ developers they'll understand the nuances here and be acceptable

introduction.rst
prince-chrismc: Maybe this?

```suggestion
  running the exercises in your own computer. There is a "narrative" to this section and the exercises might depend on some previous explanations and code - building on the previous example.
```

czoido (approved):

prince-chrismc (approved):

PR #2957, opened 2023-02-08
Add some context for conan config home command

RubenRBS:


reference/commands/config.rst
```diff
@@ -19,6 +19,11 @@ conan config home
     --logger    Show the output with log format, with time, type and message.
 
 
+The ``conan config home`` command returns the path of the Conan home folder.
+This folder contains the Conan cache under ``p/``, Conan profiles under ``profiles/``,
+Conan extensions under ``extensions/``, and the Conan config files, among others.
+
+
```
franramirez688, line 26: Perhaps, I'd enumerate all the folders/files. Something like this:

```
The ``conan config home`` command returns the path of the Conan home folder. It will normally have the following tree structure:

* ``extensions/``: Conan extensions folder.
* ``p/``: Conan data folder.
* ``profiles/``: Conan profiles folder.
* ``remotes.json``: Remotes file.
* ``settings.yml``: Conan settings file.
* ``global.conf``: Conan configuration file.
* ``version.txt``: Conan version file.
```
We don't have `:ref:` links for all of them but we can add them later. Wdyt?

franramirez688 (approved):

reference/commands/config.rst
memsharded: Too much for this command ref? I think a link to the "config files" section in the docs would be enough.

memsharded (approved):

reference/commands/config.rst
franramirez688: Then, let's keep it as:

```
The ``conan config home`` command returns the path of the Conan home folder.
```

I can add the `:ref:` to config files in my opened PRs.

RubenRBS: I've added the text, but I don't quite know what ref to add exactly

PR #2958, opened 2023-02-08
Feature/develop2 python requires

memsharded:


reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
\ No newline at end of file
+===============
+
+
+Introduction
+------------
+
+The ``python_requires`` feature is a very convenient way to share files and code between
+different recipes. A python requires is a special recipe that do not create packages and
+it is just intended to be reused by other recipes.
```
RubenRBS, line 12:
```suggestion
different recipes. A python requires is a special recipe that does not create packages and
is just intended to be reused by other recipes.
```

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+        return 234
+
+    class Pkg(ConanFile):
+        name = "pyreq"
+        version = "0.1"
+        package_type = "python-require"
+
+     
+And then we will make it available to other packages with ``conan create .``. Note that a ``python-require``
+package do not create binaries, it is just the recipe part.
```
RubenRBS, line 32:
```suggestion
package does not create binaries, it is just the recipe part.
```

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+        exports = "*"
+
+
+.. code-block:: bash
+
+    $ echo "Hello" > myfile.txt
+    $ conan create .
+
+
+Now the python-require has been created, we can access its path (the place where *myfile.txt* is) with the
```
RubenRBS, line 179:
```suggestion
Now that the python-require has been created, we can access its path (the place where *myfile.txt* is) with the
```

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+
+        def build(self):
+            pyreq_path = self.python_requires["pyreq"].path
+            myfile_path = os.path.join(pyreq_path, "myfile.txt")
+            content = load(self, myfile_path)  # content = "Hello"
+            self.output.info(content)
+            # we could also copy the file, instead of reading it
+
+
+Note that only ``exports`` work for this case, but not ``exports_sources``.
```
RubenRBS, line 200:
```suggestion
Note that only ``exports`` works for this case, but not ``exports_sources``.
```

RubenRBS (approved):
👍

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
----------------
\ No newline at end of file
+===============
+
+
+Introduction
+------------
+
+The ``python_requires`` feature is a very convenient way to share files and code between
+different recipes. A python requires is a special recipe that do not create packages and
```
SSE4, line 11:
```suggestion
different recipes. A python require is a special recipe that does not create packages and
```

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+        package_type = "python-require"
+
+     
+And then we will make it available to other packages with ``conan create .``. Note that a ``python-require``
+package do not create binaries, it is just the recipe part.
+
+.. code-block:: bash
+
+    $ conan create .
+    # It will only export the recipe, but NOT create binaries
```
SSE4, line 37:
```suggestion
    # It will only export the recipe, but will NOT create binaries
```

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+A common use case would be to declare a base class with methods we want to reuse in several
+recipes via inheritance. We'd write this base class in a python-requires package: 
+
+.. code-block:: python
+
+    from conan import ConanFile
+
+    class MyBase:
+        def source(self):
+            self.output.info("My cool source!")
```
SSE4, line 91: please add spaces/newlines between functions, it's hard to read like that

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+
+.. code-block:: python
+    
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+        version = "0.1"
+        python_requires = "pyreq/0.1"
+        python_requires_extend = "pyreq.MyBase"
```
SSE4, line 134: can I use `python_requires_extend = "pyreq.MyBase1", "myreq.MyBase2"`?

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+different recipes. A python requires is a special recipe that do not create packages and
+it is just intended to be reused by other recipes.
+
+A very simple recipe that we want to reuse could be:
+
+.. code-block:: python
+    
+    from conan import ConanFile
+
+    myvar = 123
```
SSE4, line 20: that's the scope/life cycle of this `global` (module) variable?
e.g. if one recipe does:
```
self.python_requires["pyreq"].module.myvar = 777
```
will other recipe see this change?

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
+        version = "1.0"
+        package_type = "python-require"
+
+
+.. code-block:: python
+    :caption: test_package/conanfile.py
+
+    from conan import ConanFile
+
+    class Tool(ConanFile):
```
SSE4, line 227: it probably needs an explicit `test_type` here (and what's the correct one for python requires?)

reference/extensions/python_requires.rst
```diff
@@ -1,4 +1,304 @@
...
 Python requires
----------------
\ No newline at end of file
+===============
+
+
+Introduction
+------------
+
+The ``python_requires`` feature is a very convenient way to share files and code between
```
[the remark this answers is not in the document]
memsharded: yes, files, any file.

reference/extensions/python_requires.rst
memsharded: Yes, but not recommended, read below

reference/extensions/python_requires.rst
memsharded: Yes

reference/extensions/python_requires.rst
memsharded: test_type has been removed in 2.0, not used anymore

SSE4: but it's not automatically adding requirements in 2.0, right?

memsharded: No, it has to be explicit (except ``python-requires``)

SSE4: why except python requires?

czoido (approved):

PR #2960, opened 2023-02-10
[watchdog] Improved `auto.sh`

franramirez688:
* Avoid launching `make html` till any change is done.
* Added `--drop` to avoid multiple instances of `watchdog` being launched (don't try to change between develop2 -> develop while the watchdog process runs in parallel without this parameter...)


RubenRBS (approved):
Nice work, sorry that I forgot when you asked me about it! Nice idea adding the `--drop`

PR #2966, opened 2023-02-14
[develop2] attributes no autodoc

memsharded:


czoido (approved):

reference/conanfile/attributes.rst
```diff
@@ -712,6 +733,18 @@ attribute.
...
+and regular packages in the cache. It is only available at the ``layout()`` method.
+
+- ``self.cpp.package``: For a regular package being used from the Conan cache. Same as
+  declaring ``self.cpp_info`` at the ``package_info()`` method.
+- ``self.cpp.source``: For "editable" packages, to describe the artifacts under
+  ``self.source_folder``
+- ``self.cpp.build``: For "editable" packages, to describe the artifacts under
+  ``self.build_folder``.
+
+
```
czoido, line 747: Maybe we can omit this and just redirect to the layout docs?

memsharded: Yes, most likely, this first round was mostly to remove the autodocs.
I changed a few things from the first name, version, etc. but most of the other attributes are left intact as they were.

PR #2972, opened 2023-02-16
Start troubleshooting and add missing binary error help

czoido:
For this: https://github.com/conan-io/conan/pull/13135/files

### DIFF
--- a/knowledge/faq.rst
+++ b/knowledge/faq.rst
@@ -52,7 +52,7 @@ By default, Conan doesn't build packages from sources. There are several possibi
 overcome this error:
 
 - You can try to build the package for your settings from sources, indicating some build
-  policy as argument, like :command:`--build zlib` or :command:`--build missing`. If the
+  policy as argument, like :command:`--build zlib*` or :command:`--build missing`. If the
   package recipe and the source code work for your settings you will have your binaries
   built locally and ready for use.
 
--- a/knowledge/faq.rst
+++ b/knowledge/faq.rst
@@ -8,5 +8,57 @@ FAQ
     There is a great community behind Conan with users helping each other in `Cpplang Slack`_.
     Please join us in the ``#conan`` channel!
 
+Troubleshooting
+---------------
+
+ERROR: Missing prebuilt package
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When installing packages (with :command:`conan install` or :command:`conan create`) it is possible
+that you get an error like the following one:
+
+.. code-block:: text
+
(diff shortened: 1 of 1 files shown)


knowledge/faq.rst
```diff
@@ -8,5 +8,57 @@ FAQ
...
+    shared=False
+
+    ERROR: Missing prebuilt package for 'zlib/1.2.11'
+    Check the available packages using 'conan list zlib/1.2.11:* -r=remote'
+    or try to build locally from sources using the '--build=zlib/1.2.11' argument
+
+    More Info at 'https://docs.conan.io/en/2/knowledge/faq.html#error-missing-prebuilt-package'
+
+This means that the package recipe ``zlib/1.2.11`` exists, but for some reason there is no
+precompiled package for your current settings. Maybe the package creator didn't build and
```
SSE4, line 44:
```suggestion
precompiled package for your current settings or options. Maybe the package creator didn't build and
```

knowledge/faq.rst
```diff
@@ -8,5 +8,57 @@ FAQ
...
+    compiler.cppstd=gnu11
+    compiler.libcxx=libc++
+    compiler.version=14
+    os=Macos
+    [options]
+    fPIC=True
+    shared=False
+
+    ERROR: Missing prebuilt package for 'zlib/1.2.11'
+    Check the available packages using 'conan list zlib/1.2.11:* -r=remote'
```
SSE4, line 38: I am missing the point of checking for available packages here. if we already know, it's not available, what does it benefit? e.g. I need package for apple-clang 14, and there is a package for apple-clang 11, may I somehow force usage of existing package?

czoido: I guess that maybe you don't care about that setting and you are ok using another setting that you have binaries available for instead of building a new one for your configuration

SSE4: yes, but then you would need to run it via `-s:h zlib:compiler.version=11`, right?
I suppose it's not that obvious and deserves some explanation

knowledge/faq.rst
```diff
@@ -8,5 +8,57 @@ FAQ
...
+providing packages for some platforms or compilers. E.g. the package creator built
+packages from the recipe for apple-clang 11, but you are using apple-clang 14.
+Also you may want to check your `package ID mode` as it may
+have an influence on the packages available for it.
+
+By default, Conan doesn't build packages from sources. There are several possibilities to
+overcome this error:
+
+- You can try to build the package for your settings from sources, indicating some build
+  policy as argument, like :command:`--build zlib` or :command:`--build missing`. If the
```
memsharded, line 55:
```suggestion
  policy as argument, like :command:`--build zlib*` or :command:`--build missing`. If the
```

memsharded (approved):

PR #2981, opened 2023-02-16
Create a new 2 conan docs slug

czoido:
This will create a new slug: https://docs.conan.io/en/2/ that after releasing will be the destination of the main redirect of the docs.


memsharded (commented):
we will drop "en" in 2.0

RubenRBS (approved):

PR #2995, opened 2023-02-20
Add sentence about deployers in extensions

RubenRBS:
Forgot to add it when I first added the deployers

### DIFF
--- a/reference/extensions.rst
+++ b/reference/extensions.rst
@@ -17,6 +17,7 @@ Conan can be extended in a few ways, with custom user code:
   parallelization tools
 - The package signing extension allows to sign and verify packages at upload and install time
   respectively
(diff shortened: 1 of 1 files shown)


reference/extensions.rst
```diff
@@ -17,6 +17,7 @@ Conan can be extended in a few ways, with custom user code:
   parallelization tools
 - The package signing extension allows to sign and verify packages at upload and install time
   respectively
+- Deployers, a mechanism to facilitate copying files form one folder, usually the Conan cache, to user folders
```
memsharded, line 20:
```suggestion
- Deployers, a mechanism to facilitate copying files from one folder, usually the Conan cache, to user folders
```

RubenRBS: 🫠

PR #2997, opened 2023-02-20
Update commands `--help` output

RubenRBS:
This includes changes from https://github.com/conan-io/conan/pull/13179

Also fixes some of the typos in the commands reference (Although I plan to improve the actual wording tomorrow, so stay tuned for that)

The most contentious thing this PR does is to get all `--help` output from an 80-char terminal window. I've done this so that the user does not need to scroll to see the whole output, which could happen for long lines. I've put forth the idea to remove the `max-width` limit on the docs content so that the pages could fill the whole screen and then we wouldn't need to squeeze the output like this, but meanwhile I think that this reads ok, but I'm open to using a different one if anyone feels too strongly against this :)


franramirez688 (approved):

RubenRBS:
Closing in favour of #2999

PR #3002, opened 2023-02-21
New cppstd for apple-clang

czoido:
Docs for https://github.com/conan-io/conan/pull/13185

### DIFF
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -41,38 +41,53 @@ and edited later. If you don't specify a *name*, the command will create the ``d
     :caption: *Creating the Conan default profile*
 
     $ conan profile detect
-    Found apple-clang 12.0
+    apple-clang>=13, using the major as version
     Detected profile:
     [settings]
     arch=x86_64
     build_type=Release
     compiler=apple-clang
-    compiler.cppstd=gnu98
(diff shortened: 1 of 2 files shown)


reference/config_files/profiles.rst
```diff
@@ -41,38 +41,53 @@ and edited later. If you don't specify a *name*, the command will create the ``d
...
 
+
+.. note:: **A note about the detected C++ standard by Conan**
+
+    Conan will always set the default C++ standard as the one that the detected compiler
+    version uses by default, except for the case of macOS using apple-clang. In this case,
+    for apple-clang>=11, it sets ``compiler.cppstd=gnu17``. If you want to use a different
+    C++ standard, you can edit the default profile file directly. First, get the location
+    of the default profile using:
+
```
czoido, line 69:
```suggestion
    C++ standard, you can edit the default profile file directly.

```

SSE4:
why exception is needed? if non-default C++ standard is needed, it could be always specified explicitly in profile.
otherwise, I think conan v2 should avoid these sophisticated workarounds and just return whatever compiler reports as is, without any adjustments. otherwise, it would cause a lot of reports like "my compiler default is X but conan reports Y", which is a huge source of confusion.

PR #3017, opened 2023-02-23
Fix #3015

### LINKED ISSUE #3015: [bug] Conan v2 still points to use old self.copy
Documentation still uses `self.copy` instead of pointing to use `conan.tools.files.copy`

https://docs.conan.io/2.0/tutorial/creating_packages/create_your_first_package.html

```
% find . -name "*.rst" -exec grep self.copy {} + 
./tutorial/creating_packages/create_your_first_package.rst:  existing CMake install functionality (if the CMakeLists.txt didn't implement it, it is easy to write ``self.copy()``
```

czoido:
Closes: https://github.com/conan-io/docs/issues/3015


uilianries (approved):
LGTM

PR #3020, opened 2023-02-23
document required_version

memsharded:


czoido (approved):

reference/conanfile/attributes.rst
```diff
@@ -101,3 +101,24 @@ Controls when the current package built binaries are uploaded or not
 
         class Pkg(ConanFile):
             upload_policy = "skip"
+
+
+required_conan_version
```
prince-chrismc, line 106: It's not actually an attribute of the conanfile class? shouldn't this be one level higher?

memsharded: Yeah, I know, but I think as this is the only "global" variable in conanfile.py, it is not worth a dedicated section, it is an "attribute" of the conanfile.py module

prince-chrismc: 👍

prince-chrismc (approved):

PR #3032, opened 2023-02-28
Avoid using pkg* unified patterns in documentation, as they might introduce ambiguities.

oxygene:
Avoid using `pkg*` unified patterns in documentation, as they might introduce ambiguities.

There are packages with common prefix out there (or might be introduced in future). Using `pkg*` e.g. for options could match more than intended by the user, e.g. when a new library `pkg-ng` is added to the project later.

Prefer using `pkg/*` instead.

See discussion at https://github.com/conan-io/conan/issues/13233#issuecomment-1447664410


memsharded (approved):
Thanks for contributing this.

Lets merge it. It might be worth checking the 2.0 docs for possible occurrences of this.

PR #3057, opened 2023-03-06
Use rm_safe for fpic in configure()

RubenRBS:
Found this while looking into the docs for https://github.com/conan-io/conan/pull/13328


memsharded (approved):
Looks good, but probably to be targeted to release/2.0 for immediate publication?

PR #3112, opened 2023-03-15
docs for SCM capture

memsharded:
Close https://github.com/conan-io/conan/issues/13338


examples/tools/scm/git/capture_scm/git_capture_scm.rst
```diff
@@ -0,0 +1,146 @@
...
+
+Capturing Git scm information
+=============================
+
+There are 2 main strategies to handle source code in recipes:
+
+- **Third-party code**: When the ``conanfile.py`` recipe is packaging third party code, like an open source library, it is typically better to use the ``source()`` method to download or clone the sources of that library. This is the approach followed by the ``conan-center-index`` repository for ConanCenter.
+- **Your own code**: When the ``conanfile.py`` recipe is packaging your own code, it is typically better to have the ``conanfile.py`` in the same repository as the sources. Then, there are 2 alternatives for achieving reproducibility:
+
+  - Using the ``exports_sources`` (or ``export_source()`` method) to capture a copy of the sources together with the recipe in the Conan package. This is very simple and pragmatic and would be recommended for majority of cases.
```
czoido, line 11:
```suggestion
  - Using the ``exports_sources`` (or ``export_source()`` method) to capture a copy of the sources together with the recipe in the Conan package. This is very simple and pragmatic and would be recommended for the majority of cases.
```

examples/tools/scm/git/capture_scm/git_capture_scm.rst
```diff
@@ -0,0 +1,146 @@
...
+  - For cases when it is not possible to store the sources in Conan recipe, for example when the package is to be consumed for someone that shouldn't have access to the source code at all, then the current **scm capture** method would be the way.
```
czoido, line 12:
```suggestion
  - For cases when it is not possible to store the sources beside the Conan recipe, for example when the package is to be consumed for someone that shouldn't have access to the source code at all, then the current **scm capture** method would be the way.
```
Something like that may sound more clear?

igagis (approved):

examples/tools/scm/git/capture_scm/git_capture_scm.rst
```diff
@@ -0,0 +1,146 @@
...
+    This revision will not be buildable in other computer
+    hello/0.1: Obtained URL: <local-path>/capture_scm and 8e8764c40bebabbe3ec57f9a0816a2c8e691f559
+    hello/0.1: Copied 1 '.yml' file: conandata.yml
+    hello/0.1: Copied 1 '.py' file: conanfile.py
+    ...
+    hello/0.1: Cloning sources from: {'commit': '8e8764c40bebabbe3ec57f9a0816a2c8e691f559', 'url': '<local-path>/capture_scm'}
+
+Let's explain step by step what is happening:
+
+- When the recipe is exported to the Conan cache, the ``export()`` method executes, running ``scm_url, scm_commit = git.get_url_and_commit()``.
```
czoido, line 90: I think we should add links to the reference for all the Git methods used, like `get_url_and_commit()` I would add one here and maybe a link in the end of the tutorial as a "see also" or "read more" section

examples/tools/scm/git/capture_scm/git_capture_scm.rst
```diff
@@ -0,0 +1,146 @@
...
+- The ``export()`` method stores the ``url`` and ``commit`` information in the ``conandata.yml`` for future reproducibility.
+- When the package needs to be built from sources and it calls the ``source()`` method, it recovers the information from the ``conandata.yml`` file, and calls ``git.clone()`` with it to retrive the sources. In this case, it will be cloning from the local checkout in ``<local-path>/capture_scm``, but if it had a remote defined, it will clone from it.
+
+
+.. warning::
+
+    To achieve reproducibility, it is very important for this **scm capture** technique that the current checkout is not dirty. If it was dirty, it would be impossible to guarantee future reproducibility of the build, so ``git.get_url_and_commit()`` can raise errors, and require to commit changes. If more than 1 commit is necessary, it would be recommended to squash those commits before pushing changes to upstream repositories.
```
czoido, line 99: Maybe we can do a second `conan create` to illustrate the fail this or add a `.gitignore` to ignore all the files you don't want to commit?

RubenRBS (approved):

czoido (approved):

PR #3145, opened 2023-04-03
Add documentation for `conan cache clean`

RubenRBS:
Docs for https://github.com/conan-io/conan/pull/13581

### DIFF
--- a/reference/commands/cache.rst
+++ b/reference/commands/cache.rst
@@ -158,9 +158,8 @@ conan cache clean
 .. code-block:: text
 
     $ conan cache clean -h
-    usage: conan cache clean [-h] [-v [V]] [-s] [-b] [-d]
-                             [-p PACKAGE_QUERY]
-                             pattern
+    usage: conan cache clean [-h] [-v [V]] [-s] [-b] [-d] [-t] [-p PACKAGE_QUERY]
+                             [pattern]
 
     Remove non-critical folders from the cache, like source, build and/or download
     (.tgz store) ones.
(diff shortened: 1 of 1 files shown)


reference/commands/cache.rst
```diff
@@ -158,26 +158,53 @@ conan cache clean
...
+                             [pattern]
 
     Remove non-critical folders from the cache, like source, build and/or download
     (.tgz store) ones.
 
     positional arguments:
       pattern               Selection pattern for references to clean
 
-    optional arguments:
+    options:
```
czoido, line 170: But that's not the real output of the command, is it?

memsharded: I think it depends on the OS Python distro, the message is different

reference/commands/cache.rst
```diff
@@ -158,30 +158,58 @@ conan cache clean
...
+  .. code-block:: text
+
+      $ conan cache clean * --temp
+
+
+- Remove the download folders for every ``zlib`` recipe
+
+  .. code-block:: text
+
+      $ conan cache clean zlib* --download
```
czoido, line 210: Maybe it's good to add an example combining several arguments ?

reference/commands/cache.rst
```diff
@@ -158,30 +158,58 @@ conan cache clean
...
+
+
+**Examples**:
+
+
+- Remove all non-critical files:
+
+  .. code-block:: text
+
+      $ conan cache clean *
```
SSE4, line 196: doesn't work for me:
```
ERROR: Define at least one argument among [--source, --build, --download]
```

SSE4 (changes requested):
I think it's better to quote wildcard arguments, as certain shells will expand them too early (zsh)

reference/commands/cache.rst
```diff
@@ -158,30 +158,58 @@ conan cache clean
...
-
       -s, --source          Clean source folders
       -b, --build           Clean build folders
       -d, --download        Clean download folders
+      -t, --temp            Clean temporary folders
       -p PACKAGE_QUERY, --package-query PACKAGE_QUERY
                             Remove only the packages matching a specific query,
                             e.g., os=Windows AND (arch=x86 OR compiler=gcc)
 
+This command will remove non-critical files that Conan generates in its execution.
```
SSE4, line 184: what are critical and non-critical files? it's not obvious

reference/commands/cache.rst
RubenRBS: Are you on the `release/2.0` branch? Works fine for me

reference/commands/cache.rst
RubenRBS: Good point, let me clarify those :)

czoido (approved):

PR #3146, opened 2023-04-03
Add docs for tools.android:cmake_legacy_toolchain

jcar87:
Docs for: https://github.com/conan-io/conan/pull/13459

### DIFF
--- a/reference/commands/config.rst
+++ b/reference/commands/config.rst
@@ -209,7 +209,7 @@ Displays all the Conan built-in configurations. There are 2 groups:
     core:non_interactive: Disable interactive user input, raises error if input necessary
     core:required_conan_version: Raise if current version does not match the defined range.
     tools.android:ndk_path: Argument for the CMAKE_ANDROID_NDK
-    tools.android:cmake_legacy_toolchain: Value for ANDROID_USE_LEGACY_TOOLCHAIN_FILE in CMakeToolchain
+    tools.android:cmake_legacy_toolchain: Boolean value for ANDROID_USE_LEGACY_TOOLCHAIN_FILE in CMakeToolchain. Not defined if not given a value.
     tools.apple:enable_arc: (boolean) Enable/Disable ARC Apple Clang flags
     tools.apple:enable_bitcode: (boolean) Enable/Disable Bitcode Apple Clang flags
     tools.apple:enable_visibility: (boolean) Enable/Disable Visibility Apple Clang flags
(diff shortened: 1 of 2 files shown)


reference/commands/config.rst
```diff
@@ -209,6 +209,7 @@ Displays all the Conan built-in configurations. There are 2 groups:
     core:non_interactive: Disable interactive user input, raises error if input necessary
     core:required_conan_version: Raise if current version does not match the defined range.
     tools.android:ndk_path: Argument for the CMAKE_ANDROID_NDK
+    tools.android:cmake_legacy_toolchain: Value for ANDROID_USE_LEGACY_TOOLCHAIN_FILE in CMakeToolchain
```
SSE4, line 212:
```suggestion
    tools.android:cmake_legacy_toolchain: (boolean) Value for ANDROID_USE_LEGACY_TOOLCHAIN_FILE in CMakeToolchain
```

reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -451,6 +451,8 @@ CMakeToolchain is affected by these ``[conf]`` variables:
 - **tools.cmake.cmaketoolchain:toolchain_file** user toolchain file to replace the ``conan_toolchain.cmake`` one.
 - **tools.cmake.cmaketoolchain:user_toolchain** list of user toolchains to be included from the ``conan_toolchain.cmake`` file.
 - **tools.android:ndk_path** value for ``ANDROID_NDK_PATH``.
+- **tools.android:cmake_legacy_toolchain**: value for ``ANDROID_USE_LEGACY_TOOLCHAIN_FILE``. It may be useful to set this to ``False`` if compiler flags
```
SSE4, line 454: is it `True` by default? (if you're saying setting it to `False` is useful, I think you imply default True).
if so, why something considered legacy is activated by default?

SSE4:
is there an official upstream documentation for that variable? I struggle to find one. if so, can we add a link here?

SSE4 (changes requested):
for conan user it's kinda non-trivial to understand here what's the difference between "legacy" and "normal" (I assume, or how is it called) toolchain file, and which one is used by default in conan, and when to use each of them. I would be grateful if you could explain these details a little bit more, for newcomers.

memsharded (approved):

jcar87:
> is there an official upstream documentation for that variable? I struggle to find one. if so, can we add a link here?

The changelog for https://github.com/android/ndk/wiki/Changelog-r23#r23c - where the flag can be used to work around this bug: https://github.com/android/ndk/issues/1693

Versions of Android NDK >= r23 can use CMake's built-in integration with CMake  >= 3.21. _Some_ versions of the Android NDK however still revert back to the older implementation due to this particular thing. In practice, a user may only want to define this if they are having troubles with CMAKE_CXX_FLAGS and similar - as the legacy CMake toolchain file has issues with that. I believe our documentation clearly states that as the motivation - I will add some clarifications otherwise.


If you feel it's insufficiently documented by Android, please raise an issue in https://github.com/android/ndk.

czoido (approved):

PR #3156, opened 2023-04-05
Add tool_requires definition to conan new

RubenRBS:


reference/commands/new.rst
```diff
@@ -50,7 +50,7 @@ The available templates are:
   Creates a simple recipe with some example code and helpful comments,
   and is a good starting point to avoid writing boilerplate code.
 
-  Its variables are: [name], [version], [description], [require, require...]
+  Its variables are: [name], [version], [description], [requires, requires...], [tool_requires, tool_requires, ...]
```
SSE4, line 53: why is it doubled here?

RubenRBS: So users know that it is a list that they can pass multiple values multiple times, but if you had to ask, then maybe it's not as obvious as I had hoped for? :/

SSE4: maybe it needs to be `[requires1, requires2...requiresN]` then?
otherwise, it looks like I need to pass the same thing twice

czoido, line 53:
```suggestion
  Its variables are: name, version, description, [requires, requires...], [tool_requires, tool_requires, ...]
```

RubenRBS: Locally, I can run:

```
~/c/s/basic $ conan new basic
File saved: conanfile.py
~/c/s/basic $ cat conanfile.py
from conan import ConanFile

class BasicConanfile(ConanFile):
    name = "pkg"
    version = "1.0"
    description = "A basic recipe"
    license = "<Your project license goes here>"
    homepage = "<Your project homepage goes here>"
    ...
```

PR #3157, opened 2023-04-05
Add documentation about using --package-query to remove only specific packages

RubenRBS:
For https://github.com/conan-io/conan/issues/13595

### DIFF
--- a/reference/commands/remove.rst
+++ b/reference/commands/remove.rst
@@ -102,6 +102,9 @@ Note that you can filter which packages will be removed using the ``--package-qu
     # Removes all the zlib/1.2.11 packages built with Clang compiler
 
 
+You can query packages by both their settings and options, including custom ones.
+
+
 All the above commands, by default, operate in the Conan cache.
 To remove artifacts from a server, use the ``-r=myremote`` argument:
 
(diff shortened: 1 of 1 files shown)


memsharded (approved):

reference/commands/remove.rst
```diff
@@ -94,7 +94,15 @@ pattern including the ``:`` separator of the ``package_id``:
     # in normal situations
 
 
-All the above commands, by default operate in the Conan cache.
+Note that you can filter which packages will be removed using the ``--package-query`` argument:
```
SSE4, line 97: can we fully document package query syntax somewhere and provide a link here?
like operators (`AND`, `NOT`, `OR`, `XOR`, `=`, `!=`, etc.), patterns (can I use `build_type=Rel*`?)
also, it's a bit unclear, if it can filter only by settings, but not by options?

RubenRBS: Good idea! Thanks for the suggestion :)

RubenRBS: I've clarified that both settings and options work, and will work on adding a syntax reference in another PR

PR #3164, opened 2023-04-10
Add docu about traits inferring based on package_type value

RubenRBS:
Had the question about what traits were inferred and realized the only docs were the actual implementation, so I've translated it to some docu

### DIFF
--- a/reference/conanfile/methods/requirements.rst
+++ b/reference/conanfile/methods/requirements.rst
@@ -106,7 +106,7 @@ force
 
 This ``requires`` will force its version in the dependency graph upstream, overriding
 other existing versions even of transitive dependencies, and also solving potential
-existing conflicts. The downstream consunmers ``force`` trait always have higher priority.
+existing conflicts. The downstream consumer's ``force`` traits always have higher priority.
 
 override
 ~~~~~~~~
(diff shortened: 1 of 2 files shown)


reference/conanfile/methods/requirements.rst
```diff
@@ -120,4 +120,22 @@ direct
...
+.. _reference_conanfile_package_type_trait_inferring:
+
+package_type trait inferring
+============================
+
+Some traits are automatically inferred based on the value of the ``package:type`` if not explicitly set by the recipe.
+
+ * ``application``: ``headers=False``, ``libs=False``, ``run=True``
+ * ``shared-library``: ``run=True``
+ * ``static-library``: ``run=False``
```
memsharded, line 135: shared and static libraries will also have ``headers=True`` and ``libs=True`` by default. 
Maybe the missing piece here is also the default traits of ``requires``, ``tool_requires`` and ``test_requires``?

memsharded (commented):
Good addition, but maybe it needs a bit more of additional information to make full sense to readers?

memsharded (approved):

czoido (approved):

PR #3178, opened 2023-04-18
Remove space in requirements options

gummif:
With whitespace it fails silently in conan 2.0.4 (but fails with `option ' xxx' doesn't exist` in conan 1).


memsharded:
I have changed the base branch to ``release/2.0`` so it is published immediately after merge.

PR #3180, opened 2023-04-19
[profile] Add profile_name attribute to Jinja context

uilianries:
Add the `profile_name` based on the original need, creating a generic profile that uses its own file name to define its settings.

Related to https://github.com/conan-io/conan/pull/13721

### DIFF
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -520,6 +520,20 @@ Some of the capabilities of the profile templates are:
      [conf]
      tools.cmake.cmaketoolchain:toolchain_file = {{ os.path.join(profile_dir, "toolchain.cmake") }}
 
+- Getting settings from a filename, including referencing the current profile name. For
+  example, defining a generic profile which is configured according to its file name.
+  The variable ``profile_name`` pointing to the current profile file name is added to the context.
+
+  .. code-block:: jinja
+     :caption: *Linux-x86_64-gcc-12*
+
+     {% set os, arch, compiler, compiler_version = profile_name.split('-') %}
+     [settings]
+     os={{ os }}
+     arch={{ arch }}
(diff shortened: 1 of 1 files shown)


reference/config_files/profiles.rst
```diff
@@ -520,6 +520,20 @@ Some of the capabilities of the profile templates are:
      [conf]
      tools.cmake.cmaketoolchain:toolchain_file = {{ os.path.join(profile_dir, "toolchain.cmake") }}
 
+- Getting settings from a filename, including referencing the current profile name. For
+  example, defining a generic profile which is configured according to its file name.
+  The variable ``profile_name`` pointing to the current profile file name is added to the context.
+
+  .. code-block:: jinja
+     :caption: *Linux-x86_64-gcc-12.jinja*
```
uilianries, line 528:
```suggestion
     :caption: *Linux-x86_64-gcc-12*
```

uilianries:
@memsharded I needed to update the parse, as it will include `.jinja` as part of the `profile_name` and need to exclude it for parsing the profile settings. Otherwise, the `compiler.version` would be `12.jinja`

reference/config_files/profiles.rst
memsharded: But why not drop the ``jinja`` extension?

uilianries: done, extension has been removed.

RubenRBS (approved):

memsharded:
This cannot be merged, because it goes to ``release/2.0``, but belongs to a feature not yet released (2.0.5), so we need to wait for merge (or please target the PR to ``develop2`` instead)

uilianries:
@memsharded fixed the base branch to develop2. Of course, I needed to use forced push to correct the base. Please, review it again.

PR #3212, opened 2023-05-16
Added `possible_values()`

franramirez688:
Docs for https://github.com/conan-io/conan/pull/13796

### DIFF
--- a/reference/conanfile/attributes/binary_model.inc
+++ b/reference/conanfile/attributes/binary_model.inc
@@ -92,7 +92,7 @@ method:
 
 
 This method is layer-based so doing ``self.settings.possible_values()`` returns the whole :ref:`reference_config_files_settings_yml`
-as a Python dict-like object, and, doing ``self.settings.compiler.version.possible_values()`` for instance, it returns
+as a Python dict-like object, and doing ``self.settings.compiler.version.possible_values()`` for instance returns
 the available versions for the compiler used by the consumer.
 
 If you want to do a safe deletion of settings, you could use the ``rm_safe()`` method.
--- a/reference/conanfile/attributes/binary_model.inc
+++ b/reference/conanfile/attributes/binary_model.inc
@@ -74,8 +74,26 @@ If you want to do a safe check of settings values, you could use the ``get_safe(
         # Will be the default version if the return is None
         build_type = self.settings.get_safe("build_type", default="Release")
 
-The ``get_safe()`` method will return ``None`` if that setting or sub-setting doesn't
+The ``get_safe()`` method returns ``None`` if that setting or sub-setting doesn't
 exist and there is no default value assigned.
+
+It's also feasible to check the possible values defined for any setting/sub-setting using the ``possible_values()``
+method:
+
+.. code-block:: python
+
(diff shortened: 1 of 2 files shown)


reference/conanfile/attributes/binary_model.inc
```diff
@@ -74,9 +74,27 @@ If you want to do a safe check of settings values, you could use the ``get_safe(
...
+    def generate(self):
+        definition = self.settings.compiler.version.possible_values()
+        self.output.info("HOST: " + ", ".join(definition))
+        definition = self.settings_build.compiler.version.possible_values()
+        self.output.info("BUILD: " + ", ".join(definition))
+        cppstds = self.settings.compiler.cppstd.possible_values()
+        self.output.info("CPPSTDS: " + str(cppstds))
+
+
+This method is layer-based so doing ``self.settings.possible_values()`` returns the whole :ref:`reference_config_files_settings_yml`
```
czoido, line 94: I think we should better not use that 'layer-based' expression and maybe extending the example on how it works a bit

RubenRBS: Agree

RubenRBS (approved):

reference/conanfile/attributes/binary_model.inc
franramirez688: Now, it's explained according to the example above.

reference/conanfile/attributes/binary_model.inc
```diff
@@ -74,9 +74,31 @@ If you want to do a safe check of settings values, you could use the ``get_safe(
...
+        # Print the available compiler versions given your host profile
+        compiler_versions = self.settings.compiler.version.possible_values()
+        self.output.info(f"[HOST] Versions for {str(self.settings.compiler)}:  {', '.join(compiler_versions)}")
+        # Print the available compiler versions given your build profile
+        compiler_versions = self.settings_build.compiler.version.possible_values()
+        self.output.info(f"[BUILD] Versions for {str(self.settings.compiler)}:  {', '.join(compiler_versions)}")
+
+
+As you can see above, this method is layer-based so doing ``self.settings.possible_values()`` returns the
+whole :ref:`reference_config_files_settings_yml` as a Python dict-like object, and doing
```
czoido, line 98:
```suggestion
As you can see above, doing ``self.settings.possible_values()`` returns the
whole :ref:`reference_config_files_settings_yml` as a Python dict-like object, and doing
```

reference/conanfile/attributes/binary_model.inc
```diff
@@ -74,9 +74,31 @@ If you want to do a safe check of settings values, you could use the ``get_safe(
         # Will be the default version if the return is None
         build_type = self.settings.get_safe("build_type", default="Release")
 
-The ``get_safe()`` method will return ``None`` if that setting or sub-setting doesn't
+The ``get_safe()`` method returns ``None`` if that setting or sub-setting doesn't
```
czoido, line 77: What about adding some link here in the introduction to the settings.yml and mentioning that these values come from that file?

czoido (approved):

PR #3217, opened 2023-05-16
git line endings FAQ

memsharded:
Close https://github.com/conan-io/conan/issues/13616


knowledge/faq.rst
```diff
@@ -109,3 +109,18 @@ This error can happen, if there are no or false authentication credentials in th
     http.client.HTTPConnection.debuglevel = 1
     
 One source of error can be the ``.netrc`` file, which is `honored by the requests library <https://requests.readthedocs.io/en/latest/user/quickstart/#custom-headers>`_.
+
+
+ERROR: Obtaining different revisions in Linux and Windows
```
SSE4, line 114: is it actually an error in client? I think that's just having two different revisions, and conan 2.0 can handle that just fine.

memsharded: It is not an actual error message, but a common situation that users perceive as an error, so better have an entry here in the FAQ

RubenRBS (approved):

PR #3228, opened 2023-05-23
fix: Improve custom commands formatters and add outputs

danimtb:
Some additional info I found missing while I was developing custom commands for the build info

### DIFF
--- a/reference/extensions/custom_commands.rst
+++ b/reference/extensions/custom_commands.rst
@@ -6,18 +6,17 @@ Custom commands
 It's possible to create your own Conan commands to solve self-needs thanks to Python and Conan public API powers altogether.
 
 Location and naming
---------------------
+-------------------
 
 All the custom commands must be located in ``[YOUR_CONAN_HOME]/extensions/commands/`` folder. If you don't know where
-``[YOUR_CONAN_HOME]`` is located, you can run :command:`conan config home` to check it.
+``[YOUR_CONAN_HOME]`` is located, you can run :command:`conan config home` to check it. If _commands_ sub-directory is not created yet, you will have to create it.
 
-If _commands_ sub-directory is not created yet, you will have to create it. Those custom commands files must be Python
-files and start with the prefix ``cmd_[your_command_name].py``. The call to the custom commands is like any other
-existing Conan one: :command:`conan your_command_name`.
+Those custom commands files must be Python files and start with the prefix ``cmd_[your_command_name].py``.
+The call to the custom commands is like any other existing Conan one: :command:`conan your_command_name`.
 
 
 Scoping
-++++++++++
++++++++
 
 It's possible to have another folder layer to group some commands under the same topic.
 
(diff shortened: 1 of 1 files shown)


reference/extensions/custom_commands.rst
```diff
@@ -63,27 +61,31 @@ Main decorator to declare a function as a new Conan command. Where the parameter
...
+    from conan.api.output import cli_out_write, ConanOutput
     from conan.cli.command import conan_command
 
+
     def output_json(msg):
-        return json.dumps({"greet": msg})
+        cli_out_write(cli_json.dumps({"greet": msg}))
+
+    def output_text(msg):
+        ConanOutput().info(msg)
```
memsharded, line 72: output text shouldn't use ``ConanOutput()``, that goes to sys.stderr, it should probably use ``cli_out_write()`` too

reference/extensions/custom_commands.rst
```diff
@@ -123,22 +125,27 @@ The command call looks like :command:`conan hello moon`.
...
         conan_api = result["conan_api"]
         ...
+        cli_out_write(cli_json.dumps(...))
 
     def format_graph_info(result):
         graph = result["graph"]
         field_filter = result["field_filter"]
         package_filter = result["package_filter"]
         ...
+        ConanOuptup().info("Conan info command output:")
```
memsharded, line 147: Typo.
```suggestion
        ConanOutput().info("Conan info command output:")
```

reference/extensions/custom_commands.rst
danimtb: but `ConanOutput()` is kind of "non-parseable" user output right? Why should it use `cli_out_write()`? This is what was confusing to me and what I'd like to clarify with this PR

memsharded: If it is a "formatter", it is intended for some kind of "final message" result. Even if in the text format is not machine readable, it should go to ``stdout``, not to ``stderr``, because it is the output of the command.

If that is not the case, and it is just informational message, it can be put in the command itself, and not in the formatter.

danimtb: updated!

memsharded: I still see the ``ConanOutput().info()`` and not ``cli_out_write()``, did you forget to push changes?

reference/extensions/custom_commands.rst
```diff
@@ -168,34 +170,53 @@ When there are sub-commands, the base command cannot define arguments, only the
...
+        cli_out_write(f"Conan info result:\n\n{graph_info_result}")
+        ...
 
     @conan_subcommand(formatters={"text": format_graph_info,
                                   "html": format_graph_html,
                                   "json": format_graph_json,
                                   "dot": format_graph_dot})
     def graph_info(conan_api, parser, subparser, *args):
         ...
+        ConanOuptup().info("Conan info command output:")
```
memsharded, line 201: Typo still here? I think I already reviewed this

```suggestion
        ConanOutput().info("Conan info command output:")
```

PR #3229, opened 2023-05-24
Add cheat sheet for Conan 2.0

prince-chrismc:
I will prep the blog PR to announce this as well :) but this can be merged if it looks good!

image

image


images/cheatsheet/tmp.txt
```diff
@@ -0,0 +1 @@
+temp
```
memsharded, line 1: Is this file necessary?

knowledge/cheatsheet.rst
```diff
@@ -0,0 +1,16 @@
+***********
+Cheat sheet
+***********
+
+JFrog has created the following visual cheat sheet for basic Conan commands and
+concepts which users can print out and use as a handy reference. It is available
```
memsharded, line 6:
```suggestion
This is a visual cheat sheet for basic Conan commands and
concepts which users can print out and use as a handy reference. It is available
```
Everything in these docs is created by JFrog, this reads a bit weird

memsharded (commented):
Looks good, please check minor comments.

SSE4:
usually, it's not good idea to store binary files like .pdf in .git repositories.
@prince-chrismc how was it generated exactly? is it possible to render PDF and PNG via some script during CI build?

also, it would make it easier to edit this cheat sheet if we need.

SSE4:
some comments (can't comment inline):
- typo `eg.` -> `e.g.`
- bad visual line break for `package_inf o()`
- build requires `CMake >= 3.23`, how do I do it for an older version? IIRC tribe set CMake 3.15+ as requirement

images/cheatsheet/tmp.txt
prince-chrismc: I needed that to uplaod the files :) deleted

memsharded (approved):

PR #3281, opened 2023-06-25
Fixes #14166: Add strip docs to cmake.install

sagi-ottopia:
Fixes #14166: Add strip docs to cmake.install

### DIFF
--- a/reference/build_helpers/cmake.rst
+++ b/reference/build_helpers/cmake.rst
@@ -357,8 +357,9 @@ install()
 Installs `CMake` project with the given parameters.
 
 Parameters:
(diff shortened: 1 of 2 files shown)


memsharded (changes requested):
This is outdated now to https://github.com/conan-io/conan/pull/14167

Thanks for contributing the docs too!

sagi-ottopia:
Hi, I would like to update the docs according to the change that I made.
Where do you think is the most appropriate docs/section for it ?
Sagi

memsharded:
https://docs.conan.io/1/reference/conanfile/tools/cmake/cmake.html#install would be good.

I have just merged develop -> release/2.0, so this will be in next 2.0.8 too

memsharded (approved):

PR #3283, opened 2023-06-27
Feature/metadata

memsharded:


reference/extensions/metadata.rst
```diff
@@ -0,0 +1,279 @@
...
+
+
+.. note::
+
+    **Best practices**
+
+    - If you are constantly downloading metadata in many of your ``install`` operations, that might indicate an abuse of the system. Metadata retrieval should be way less common than ``install`` operations.
+    - Package usage shouldn't rely on metadata, if normal consumption of a package makes the metadata mandatory, then this metadata should be part of the package instead.
+    - If there are many small files, it might be necessary to zip those files to avoid excessive upload and download time.
+    - As metadata will not be downloaded by default, it will not be promoted by default in ``conan download`` + ``conan upload`` operations. The recommended way to run a server-server promotion (copy) is using the server tools, and guaranteeing that the metadata folders are copied too.
```
stackfun, line 273: Will metadata files also integrate with Artifactory's BuildInfo feature?

Might get tricky if metadata files are added later.

memsharded: Not at this moment. The generation of "modern" BuildInfo has just landed in the Conan extensions https://github.com/conan-io/conan-extensions repository, and will likely require some maturing too. Metadata is a completely new feature, that changes some core management of artifacts by Conan. It is just too early to try to make both things play together. It might not be even possible, as you say, because the capability of being able to add and remove metadata files goes against the declaration stored in a BuildInfo. In other words, the "BuildInfo" is information of built artifacts, but metadata are not build artifacts, but that, metadata, and not covered by BuildInfo. We will see about that later down the road.

memsharded:
Superseded by https://github.com/conan-io/docs/pull/3425

PR #3288, opened 2023-07-08
Backup sources docs, second try

RubenRBS:
After some tries (No lost work though, they might end up as part of the blog post!), I've decided on something simpler that gives a quick overview of the feature.

Once the blog is released, we can add a Read further link to it


reference.rst
```diff
@@ -16,3 +16,4 @@ Reference
    reference/extensions
    reference/environment
    reference/conan_server
+   reference/backup_sources
```
memsharded, line 19: Doesn't feel the best place for this

reference/backup_sources.rst
```diff
@@ -0,0 +1,59 @@
+.. _conan_backup_sources:
+
+Backup sources
+==============
+
+The *backup sources* feature is intended for storing the downloaded recipe sources in a secondary location,
+allowing future reproducibility of your builds even in the case where the original mirrors are no longer accessible.
```
memsharded, line 7:
```suggestion
allowing future reproducibility of your builds even in the case where the original download URLs are no longer accessible.
```

reference/backup_sources.rst
```diff
@@ -0,0 +1,59 @@
...
+when a ``sha256`` signature is provided.
+
+
+Configuration
+-------------
+This feature is controlled by a few configuration items:
+
+core.sources:download_cache
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Local path to store the sources backup to.
```
memsharded, line 19: (optional), if not defined it will store things in the ``.conan2/`` folder

reference/backup_sources.rst
```diff
@@ -0,0 +1,59 @@
...
+List of URLs that Conan will try to download the sources from, where ``origin`` represents the original URL.
+This allows to control the fetch order, either ``["origin", "https://your.backup/remote/"]``
+to look into and fetch from your backup remote only if and when the original source is not present,
+or ``["https://your.backup/remote/", "origin"]`` to prefer your remote ahead of the recipes.
+(*Note that being a list, multiple remotes are also possible*)
+
+core.sources:upload_url
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+URL of the remote to upload the backups when calling ``conan upload``,
+which might or might be not different from any of the URLs defined for download.
```
memsharded, line 32:
```suggestion
which might or might not be different from any of the URLs defined for download.
```

reference/backup_sources.rst
```diff
@@ -0,0 +1,59 @@
...
+when a ``sha256`` signature is provided.
+
+
+Configuration
+-------------
+This feature is controlled by a few configuration items:
+
+core.sources:download_cache
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Local path to store the sources backup to.
```
RubenRBS, line 19:
```suggestion
(Optional) Local path to store the sources backup to. If not set, the default Conan home cache path will be used.
```

reference/backup_sources.rst
```diff
@@ -0,0 +1,59 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+List of URLs that Conan will try to download the sources from, where ``origin`` represents the original URL.
+This allows to control the fetch order, either ``["origin", "https://your.backup/remote/"]``
+to look into and fetch from your backup remote only if and when the original source is not present,
+or ``["https://your.backup/remote/", "origin"]`` to prefer your remote ahead of the recipes.
+(*Note that being a list, multiple remotes are also possible*)
+
+core.sources:upload_url
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+URL of the remote to upload the backups when calling ``conan upload``,
```
RubenRBS, line 31:
```suggestion
(Optional) URL of the remote to upload the backups when calling ``conan upload``,
```

reference.rst
RubenRBS: Yep, I also know we talked about this. I've moved it to a new "Other" section which I _think_ was the call we made, but let me know otherwise :)

reference/other.rst
```diff
@@ -0,0 +1,14 @@
...
+
+These are other features that Conan also implements:
+
+
+Contents:
+
+.. toctree::
+   :maxdepth: 1
+
+   other/backup_sources
```
memsharded, line 14: I don't love the ``other`` section, I think we need to look for a better naming (I have tried a bit without success...)

memsharded: This feature should definitely have an entry in https://docs.conan.io/2/tutorial/other_features.html, but we still need to find a place for it.

memsharded (commented):
We will have the "Tutorial - DevOps" section, that will be a good place to write an introduction to this feature that is not-reference.

That means that the reference could be made part of the ``conf`` reference, in an structured way, but part of ``conf``

RubenRBS:
Closing in favour of #3365

PR #3291, opened 2023-07-11
added example with cmake module inside tool_requires

memsharded:
Docs for https://github.com/conan-io/conan/pull/14264

Can go directly to release/2.0, nothing really new from the release.

### DIFF
--- a/examples/graph/tool_requires/use_cmake_modules.rst
+++ b/examples/graph/tool_requires/use_cmake_modules.rst
@@ -1,0 +1,93 @@
+Use cmake modules inside a ``tool_requires`` transparently
+==========================================================
+
+When we want to reuse some ``.cmake`` scripts that are inside another Conan package
+there are several possible different scenarios, like if the ``.cmake`` scripts are inside
+a regular ``requires`` or a ``tool_requires``.
+
+Also, it is possible to want 2 different approaches:
+
+- The consumer of the scripts can do a explicit ``include(MyScript)`` in their CMakeLists.txt. This approach is nicely explicit and simpler to setup, just define ``self.cpp_info.builddirs`` in the recipe, and consumers with ``CMakeToolchain`` will automatically be able to do the ``include()`` and use the functionality. See the :ref:`example here<consume_cmake_macro>` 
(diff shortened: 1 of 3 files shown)


examples/graph/tool_requires/use_cmake_modules.rst
```diff
@@ -0,0 +1,93 @@
...
+.. code-block:: cmake
+    :caption: myfunctions/myfunction.cmake
+
+    function(myfunction)
+        message("Hello myfunction!!!!")
+    endfunction()
+
+We can do a ``cd myfunctions && conan create .`` and that will create the ``myfunctions/1.0`` package containing the cmake script.
+
+Then, a consumer package will be like:
```
RubenRBS, line 45:
```suggestion
Then, a consumer package will look like:
```

examples/graph/tool_requires/use_cmake_modules.rst
```diff
@@ -0,0 +1,93 @@
...
+And a ``myfunction.cmake`` file in:
+
+.. code-block:: cmake
+    :caption: myfunctions/myfunction.cmake
+
+    function(myfunction)
+        message("Hello myfunction!!!!")
+    endfunction()
+
+We can do a ``cd myfunctions && conan create .`` and that will create the ``myfunctions/1.0`` package containing the cmake script.
```
RubenRBS, line 43:
```suggestion
We can do a ``cd myfunctions && conan create .`` which will create the ``myfunctions/1.0`` package containing the cmake script.
```

RubenRBS (approved):

czoido (approved):
I think this should have an analogous PR in the examples2 repo, opening and issue there so we don't forget to add the code and modify the example.

memsharded:
> I think this should have an analogous PR in the examples2 repo, opening and issue there so we don't forget to add the code and modify the example.

I am not fully sure about this (the usage of resources). This is extracted exactly from a test in our test suite. Adding it to examples code seems duplicating the CI for it? It is that necessary? I know I always face the same when doing examples from Conan existing tests... I am good with adding them to the examples2 repo, just something doesn't seem great while doing it.

PR #3313, opened 2023-07-28
Update extend_own_cmake_presets.rst to specify correct directory

orgarten:
The link to the sources for this part pointed to the `local_flow_cmake_presets` directory which isn't the correct one here.


memsharded (approved):
Thanks very much for fixing this!

memsharded:
I have changed the branch to release/2.0 so it is published immediately after merge.

PR #3315, opened 2023-08-03
Update commands.rst

brosmar:
I think it should be "build" not "install"


RubenRBS:
Hi @brosmar thanks a lot for your contribution, great catch! Please remember to sign the CLA so that we can merge the PR :)

PR #3317, opened 2023-08-04
clarify tutorial immutable sources

memsharded:


knowledge/guidelines.rst
```diff
@@ -70,3 +70,4 @@ Forbidden practices
   like ``conan cache path`` to get information of the current folders. The Conan cache might 
   be checked while debugging, as read-only, but it is not allowed to edit, modify or delete 
   artifacts or files from the Conan cache by any other means that Conan command line or public API.
+- **Sources used in recipes must be immutable**. It is not allowed to use moving targets like a ``git`` branch or a download of a file that is continuously rewritten in the server. ``git`` checkouts must be of an immutable tag or a commit, and ``download()/get()`` must use checksums to verify the server files doesn't change. Not using immutable sources will be undefined behavior.
```
jcar87, line 73: This one is clear - although I wonder if we should add further clarification, either "for a given version of a recipe" or "once a recipe it's exported to the Conan cache, it is expected that the sources are immutable, that is, that retrieving the sources in the future will always retrieve the _same_ sources." and then the test (moving targets, etc). 

In case this isn't clear and/or users are using the same recipe to support multiple versions, but in a "correct" way.

tutorial/creating_packages/create_your_first_package.rst
```diff
@@ -313,6 +313,9 @@ located in the user home folder under the ``.conan2`` folder. Conan will use the
 You already used the :command:`conan list` command to list the recipes and binaries stored
 in the local cache. 
 
+An **important** note: the Conan cache is read-only. Do not modify, add, remove or change
```
jcar87, line 316: Wording suggestion: 
"the contents of the Conan cache are private to the Conan client - modifying, adding, removing or changing files inside the Conan cache is undefined behaviour likely to cause breakages."

(pedantic: it is "read-only" from the users point of view, but obviously not for the filesystem, in case users think filesystem)

tutorial/creating_packages/handle_sources_in_packages.rst
```diff
@@ -61,6 +61,8 @@ Let's have a look at the changes in the *conanfile.py*:
         default_options = {"shared": False, "fPIC": True}
 
         def source(self):
+            # Please, be aware that using the head of the branch instead of an immutable tag
+            # or commit is not a bad practice and not allowed by Conan
```
jcar87, line 65: I'd personally go for "strongly discouraged, unsupported by Conan and likely to cause issues"

tutorial/creating_packages/handle_sources_in_packages.rst
```diff
@@ -109,6 +113,15 @@ it. Note that we pass the ``strip_root=True`` argument so that if all the unzipp
 contents are in a single folder, all the contents are moved to the parent folder (check
 the :ref:`conan.tools.files.unzip()<conan_tools_files_unzip>` reference for more details).
 
+.. warning::
+
+    Using not immutable source origins, like the HEAD branch, or any other branch that is
```
jcar87, line 118: same as earlier, I'd perhaps start the sentence with the expectation, e.g: "it is expected that retrieving the sources in the future produces the same results. Using mutable source origins, like a moving reference in git (e.g HEAD branch), or the URL to a file where the contents may change over time, is strongly discouraged and not supported. Not following this practice will result in undefined behavior likely to cause breakages"

(but otherwise this is good)

jcar87 (approved):
LGTM if we want to merge this, otherwise some very minor comments.

knowledge/guidelines.rst
memsharded: Changed.

tutorial/creating_packages/create_your_first_package.rst
memsharded: Makes sense, changing it.

Issue #3318, opened 2023-08-07
Improve README with branch description


uilianries:
Now with Conan 1.x and 2.x under development, we have several branches like `develop2`, `master`, `develop`, `release/2.0` ... but there is no hint about which one should be used. It's common for users pick a wrong branch when opening a new PR.

It would be great the README file had some better explanation to avoid confusion.

PR #3319, opened 2023-08-07
Add branches description

uilianries:
Add description section in README about useful branches when contributing to docs.

It should be replicated to other docs branches after merging this one.

Related to #3318


memsharded (approved):
Excellent, many thanks!

PR #3337, opened 2023-08-28
Add docs for no user/channel pattern matches

czoido:
Had no idea where to put this... I finally mentioned this in the conan list docs that I think it could have more visibility there....

Docs for https://github.com/conan-io/conan/pull/14338

### DIFF
--- a/reference/commands/list.rst
+++ b/reference/commands/list.rst
@@ -135,6 +135,18 @@ Use the pattern for searching only references matching a specific channel:
         hello/2.3.2@mycompany/stable
         hello/1.0.4@mycompany/stable
 
+Use the pattern for searching only references that don't have user and channel using the
+``...@`` pattern:
+
+.. code-block:: text
+    :caption: *list references without user and channel*
+
+    $ conan list "*/*@"
+    Local Cache
+      string-view-lite
+        string-view-lite/1.6.0
+      zlib
+        zlib/1.2.11
 
 Listing recipe revisions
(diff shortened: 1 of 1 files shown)


memsharded (approved):

reference/commands/list.rst
```diff
@@ -135,6 +135,18 @@ Use the pattern for searching only references matching a specific channel:
         hello/2.3.2@mycompany/stable
         hello/1.0.4@mycompany/stable
 
+Use the pattern for searching only references that don't have user and channel using the
+``...@`` pattern:
```
czoido, line 139:
```suggestion
Use the ``...@`` pattern for searching only references that don't have _user_ and _channel_:
```

reference/commands/list.rst
```diff
@@ -135,6 +135,17 @@ Use the pattern for searching only references matching a specific channel:
         hello/2.3.2@mycompany/stable
         hello/1.0.4@mycompany/stable
 
+Use the ``...@`` pattern for searching only references that don't have _user_ and _channel_:
```
czoido, line 138:
```suggestion
Use the ``...@`` pattern for searching only references that don't have *user* and *channel*:
```

prince-chrismc:
> Had no idea where to put this

I would love a page for the pattern matching. Most of it's the same but there are a cases like `conan remove` where it's backwards and my copy-paste hurts me

Supported symbols and default values would be great

PR #3350, opened 2023-08-29
docs for credentials.json

memsharded:
https://github.com/conan-io/conan/pull/14392

### DIFF
--- a/reference/config_files/credentials.rst
+++ b/reference/config_files/credentials.rst
@@ -40,8 +40,8 @@ Conan will be able to extract the credentials from this file automatically when 
     using the 
 
 The priority of credentials origins is as follows:
-- If the ``credentials.json`` file exist, it has higher priority, if an entry for the remote exists, it will be used. If it doesn'take
-  work, it will be an error.
+- If the ``credentials.json`` file exist, it has higher priority, if an entry for the remote exists, it will be used. If it doesn't
+work, it will be an error.
 - If an entry in the ``credentials.json`` for that remote does not exist, it will look for defined environment variables
 - If environment variables don't exist, it will request interactively the credentials. If ``core:non_interactive=True``, it will error.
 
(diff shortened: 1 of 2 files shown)


reference/config_files/credentials.rst
```diff
@@ -0,0 +1,64 @@
...
+.. note::
+
+    Conan does not pre-emptively use the credentials to force a login automatically in every remote defined at every Conan command.
+    By default Conan uses the previously stored tokens or anonymous usage, until a explicit ``conan remote login`` command is done, 
+    or until a remote server launches an authentication error. When that happens, authentication against that server will be done,
+    using the 
+
+The priority of credentials origins is as follows:
+- If the ``credentials.json`` file exist, it has higher priority, if an entry for the remote exists, it will be used. If it doesn'take
+  work, it will be an error.
```
czoido, line 44:
```suggestion
- If the ``credentials.json`` file exist, it has higher priority, if an entry for the remote exists, it will be used. If it doesn't
work, it will be an error.
```

reference/config_files/credentials.rst
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+Conan will be able to extract the credentials from this file automatically when necessary and requested by the server.
+
+.. note::
+
+    Conan does not pre-emptively use the credentials to force a login automatically in every remote defined at every Conan command.
+    By default Conan uses the previously stored tokens or anonymous usage, until a explicit ``conan remote login`` command is done, 
+    or until a remote server launches an authentication error. When that happens, authentication against that server will be done,
+    using the
```
czoido, line 40: something missing here?

reference/config_files/credentials.rst
```diff
@@ -0,0 +1,64 @@
...
+- With a ``credentials.json`` file put in the Conan cache.
+
+This section describes the usage of ``credentials.json`` file.
+
+
+This file has the following format, in which every ``credentials`` entry should have a ``remote`` name, matching the name defined in ``conan remote list``.
+Then, the ``user`` and ``password`` fields.
+
+.. code-block:: json
+
```
czoido, line 22: I would add a comment somewhere, like a warning saying that is not recommended to store plain text credentials and to better use environment with the jinja syntax

reference/config_files/credentials.rst
memsharded, line 40:
```suggestion
    using the ``credentials.json`` file, the environment variables or the user interactive inputs.
```

reference/config_files/credentials.rst
memsharded: Users requesting this file requested it because they said it was more secure to have credentials in a file than having them in environment variables...

RubenRBS (approved):

czoido (approved):

Issue #3205, opened 2023-05-16
Add a note about Ubuntu 23.04 when installing Conan with pip


uilianries:
For Ubuntu 23.04, first is needed to create a virtual environment before running `pip` command. 

https://docs.conan.io/2/installation.html#known-installation-issues-with-pip

https://www.reddit.com/r/Python/comments/1338oge/you_cant_use_pip_on_ubuntu_2304_anymore/

samuel-emrys:
It's already on the installation page, but you should really push installation using `pipx` here - it installs python applications into a virtualenv and makes them available in the global context.. pretty much exactly what you would have to do manually here.

memsharded:
The PR #3360 is enough to close this ticket?

uilianries:
@memsharded Yes, it's clear and useful. Thank you @czoido !!!

PR #3360, opened 2023-09-05
Add a note about installing with pipx

czoido:
Closes: https://github.com/conan-io/docs/issues/3205

### DIFF
--- a/installation.rst
+++ b/installation.rst
@@ -70,7 +70,7 @@ The upgrade shouldn't affect the installed packages or cache information. If the
 Install with pipx
 -----------------
 
-In certain scenarios, attempting to install with `pip` may yield the following error:
+In certain scenarios, attempting to install with ``pip`` may yield the following error:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):

memsharded (approved):

installation.rst
```diff
@@ -67,6 +67,48 @@ If installed via ``pip``, Conan version can be updated with:
...
+
+For those cases it's recommended to use `pipx` to install Conan. `pipx` creates a virtual
+environment for each Python application, ensuring that dependencies do not conflict. The
+advantage is that it isolates Conan and its dependencies from the system Python and avoids
+potential conflicts with system packages while providing a clean environment for Conan to
+run.
+
+To install Conan with `pipx`:
+
+1. Ensure `pipx` is installed on your system. If it isn't, check the installation guidelines
```
jcar87, line 100: It may be worth adding instructions as an example for an Ubuntu/Debian distro?

the link from pipx doesn't really include instructions specific to Linux, other than `pip install --user pipx` - which really will have similar problems as trying to install Conan itself (in the sense that the environment is externally managed, although maybe it is more relax for `--user`).

on Debian based distros, one can do:
```
apt-get install pipx
pipx ensurepath
```
then restart the shell and install conan via `pipx install conan`

This is probably the safest, and equivalent to doing `brew install pipx`. 

So I would perhaps go beyond that the pipx documentation says, and recommend installing `pipx` via the system package manager, and then use _that_ pipx. Unfortunately it may be the case that the `pipx` package has different names depending on the distro (e.g. `pipx` vs `python3-pipx` or `python-pipx`)

PR #3361, opened 2023-09-05
Add documentation for `self.output` and `self.run()`

RubenRBS:
Thanks to Pete on Slack for noticing that these were missing and bringing it to my attention :)


reference/conanfile/methods/run.rst
```diff
@@ -0,0 +1,19 @@
...
+
+Running commands
+----------------
+
+.. code-block:: python
+
+    run(self, command, stdout=None, cwd=None, ignore_errors=False, env="", quiet=False, shell=True, scope="build")
+
+
+``self.run()`` is a helper to run system commands and throw exceptions when errors occur,
```
memsharded, line 14: It has a very important responsibility: it is the one injecting the calls to activate environment scripts.

reference/conanfile/methods.rst
```diff
@@ -23,6 +23,7 @@ Methods
    methods/package_id
    methods/package_info
    methods/requirements
+   methods/run
```
memsharded, line 26: I think this is different, as this is not a recipe method. This section is to document the methods that a recipe should write, not methods that it can call like this one. Not sure either where/how to document it, but it is a bit confusing imo.

reference/conanfile/attributes.rst
```diff
@@ -73,6 +73,27 @@ The profile entries have priority.
...
+
+Use the `self.output` to print contents to the output.
+
+..  code-block:: python
+
+   self.output.success("This is good, should be green")
+   self.output.info("This is neutral, should be white")
+   self.output.warning("This is a warning, should be yellow")
+   self.output.error("Error, should be red")
+   self.output.rewrite_line("for progress bars, issues a cr")
```
memsharded, line 92: Lets drop the ``rewrite_line`` atm, I think we want to internally get rid of it.

reference/conanfile/methods.rst
RubenRBS: I agree with this. I couldn't find a better place and after confirming with @czoido, I've moved it to the same place they have in Conan 1's docs: Another subsection in the `conanfile.py` one. Let me know what you think! :)

reference/conanfile/running_and_output.rst
```diff
@@ -0,0 +1,55 @@
...
+   info(msg)
+   highlight(msg)
+   success(msg)
+   warning(msg)
+   error(msg)
+
+   title(msg)
+   subtitle(msg)
+
+   flush()
```
memsharded, line 28: Probably hide the ``title, subtitle`` ones, those are reserved for Conan main flow. Probably remove ``flush()`` too

reference/conanfile/attributes.rst
```diff
@@ -73,6 +73,28 @@ The profile entries have priority.
...
+
+Use the `self.output` to print contents to the output.
+
+..  code-block:: python
+
+   self.output.success("This is good, should be green")
+   self.output.info("This is neutral, should be white")
+   self.output.warning("This is a warning, should be yellow")
+   self.output.error("Error, should be red")
+   self.output.title("This is a title, will take a few lines for visibility")
```
memsharded, line 92: Remove the ``title``, that is for Conan only.

memsharded (approved):
Looks good, I'd just remove the title, subtitle and flush

memsharded (approved):

czoido (approved):

PR #3363, opened 2023-09-06
Add docs for CLion integration

czoido:


memsharded (commented):
I think this is very good.
My only question is if the location under ``integrations/clion`` is simply too much. It is like a full tutorial inside the section. While other ``integrations`` will be mostly redirects to other places in the docs. 
When we add the future ``Visual Studio`` plugin documentation, will it be under ``integrations`` too? Because at the moment it is mostly a redirect to ``MSBuild`` integration: https://docs.conan.io/2/integrations/visual_studio.html

Maybe we should have a ``reference`` section for these things, and leave ``integrations`` as one entry point?

czoido:
> I think this is very good. My only question is if the location under `integrations/clion` is simply too much. It is like a full tutorial inside the section. While other `integrations` will be mostly redirects to other places in the docs. When we add the future `Visual Studio` plugin documentation, will it be under `integrations` too? Because at the moment it is mostly a redirect to `MSBuild` integration: https://docs.conan.io/2/integrations/visual_studio.html
> 
> Maybe we should have a `reference` section for these things, and leave `integrations` as one entry point?

For this specific case, I think we could leave it here. The rest of the integrations redirect to parts of the documentation that are scattered anyway, like CMake redirecting to CMakeToolchain, CMakeDeps, CMake, etc. but here we would redirect only to one place in the docs that would be the reference of the CLion plugin and I think we can save users one click. WDYT?

memsharded:
> For this specific case, I think we could leave it here. The rest of the integrations redirect to parts of the documentation that are scattered anyway, like CMake redirecting to CMakeToolchain, CMakeDeps, CMake, etc. but here we would redirect only to one place in the docs that would be the reference of the CLion plugin and I think we can save users one click. WDYT?

Sounds good, lets merge it

memsharded (approved):

PR #3377, opened 2023-09-13
Update attributes.rst

toon-leader-bacon:
Update the English explanation to match the python example.

The python code block example (on line 267) shows the following:
`"compiler": {"Visual Studio": {"version": [11, 12]}},`

However the explanation of this code says:
> In this example we have just defined that this package only works in Windows, with VS 10 and 11.

Notice the version numbers do not match from the code example to the explanation. 
This proposed change will update the explanation to match the code example more closely. 
Alternatively we could update the code example to update the explanation.

Please let me know if there are any other changes I should make here to improve the documentation quality.

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -267,7 +267,7 @@ attribute:
             "compiler": {"Visual Studio": {"version": [11, 12]}},
             "arch": None}
 
-In this example we have just defined that this package only works in Windows, with VS 10 and 11.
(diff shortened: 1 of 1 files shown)


toon-leader-bacon:
Good to know thanks for the tip! 
As you can probably tell, I'm learning Conan and looking over some of the documentation so I appreciate the advice.

memsharded:
Sure! :)

If you are learning Conan, then the first recommendation: you should definitely start with Conan 2.0, and 2.0 docs as well. Any specific reason you started with 1.X?

PR #3393, opened 2023-09-15
Add docu describing the basic usage of the prerelease resolve conf

RubenRBS:
I'm not 100% sold on th text I just wrote, but as discussed before, it's better than nothing :)

The graph looks like this:

image

### DIFF
--- a/devops.rst
+++ b/devops.rst
@@ -13,4 +13,4 @@ If you plan to use Conan in production in your project, team, or organization, t
    :maxdepth: 1
 
    devops/using_conancenter
-   devops/resolve_prereleases
+   devops/versioning
--- a/devops.rst
+++ b/devops.rst
@@ -13,3 +13,4 @@ If you plan to use Conan in production in your project, team, or organization, t
    :maxdepth: 1
 
    devops/using_conancenter
+   devops/resolve_prereleases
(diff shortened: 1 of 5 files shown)


devops.rst
```diff
@@ -13,3 +13,4 @@ If you plan to use Conan in production in your project, team, or organization, t
    :maxdepth: 1
 
    devops/using_conancenter
+   devops/resolve_prereleases
```
memsharded, line 16: Probably should go under a section called "Versioning" or the like

devops/resolve_prereleases.rst
```diff
@@ -0,0 +1,62 @@
...
+
+This conf has the added benefit of affecting the whole dependency graph, so that if any of our dependencies also define
+a requirement to our library of interest, the new version will also be picked up by it.
+
+Let's see this in action. Imagine we have the following (summarized) dependency graph,
+in which we depend on ``libpng`` and ``libmysqlclient``, both of which depend on ``zlib`` via the ``[>1.2 <2]`` version range:
+
+.. graphviz::
+
+    digraph G {
```
memsharded, line 25: It is possible to do a bit of styling, I think we have some in other parts of the docs if you want to copy it

memsharded (approved):
I think it is good!

czoido (approved):

PR #3400, opened 2023-09-25
Add products pipeline docu, add build-order example

RubenRBS:
The images come directly from slides, didn't want to waste time creating dot code while sketching, let me know if it looks bad and I'll take the time to create it


memsharded (commented):
Thanks for pushing this!

I think we need to define first an outline of the different sections, contents, etc, to provide some scope for this. I am totally ok with adding just parts of it, like the description of the concept of "products" pipeline, but lets define first its overall location. For a start, this is not mostly about "versioning", but this belongs to the "continuous integration" problem, almost irrespective of the versioning approach.

PR #3427, opened 2023-10-20
Minor fixes and new note for zlib unexpected case in tutorial

juansblanco:
- Made some minor wording changes on the "Consuming packages" section of the tutorial.

- Added new note to add context to the zlib missing error in tutorial.


tutorial/consuming_packages/different_configurations.rst
```diff
@@ -202,6 +204,23 @@ executable can't find the shared libraries for *Zlib* that we just installed.
     $ Release\compressor.exe
     (on a pop-up window) The code execution cannot proceed because zlib1.dll was not found. Reinstalling the program may fix this problem.
 
+.. note:: **Possible issue with Windows consoles**
```
memsharded, line 207: This reads a bit too heavy for the likelihood of impact and scope, only Windows and only for shared in some terminals:
I'll try to lighten it a bit, make it much shorter. The thing is that if you activate the Conan virtualenv it should take the Conan one, so this applies only to:
- Windows user
- With a terminal like cmder that contains some zlib.dll
- That didn't activate the ``conanrun.bat`` virtualenv
- And even if they didn't, it still worked, because it found zlib.dll in the path thanks to cmder

Is this correct?

juansblanco: Yes, it's true it's not as common as I thought. I think we could have a shorter notice saying that it can still run if the library is obtained from a different path by the system. It's a more generic case that can cover a wider range.

memsharded: Sounds good!

tutorial/consuming_packages/different_configurations.rst
```diff
@@ -46,7 +46,9 @@ folder:
 
 As you can see, the profile has different sections. The ``[settings]`` section is the one
 that has information about things like the operating system, architecture, compiler, and
-build configuration. When you call a Conan command setting the ``--profile`` argument,
+build configuration. The default profile can also be checked with the command :command:`conan profile show`.
+
```
czoido, line 50: Maybe we could move this to the comments in the code snippet, so that users can try also the command at the same moment that they output the profile to the console
```suggestion
build configuration.
```

juansblanco: done!

czoido (approved):

memsharded (approved):

PR #3443, opened 2023-11-02
Updated max C++ compiler version

juansblanco:
Changed the max compiler version from 14 to 23 (latest) to avoid failures with some configurations using versions newer than 14.

Change also made in the examples2 repository: https://github.com/conan-io/examples2/pull/120

### DIFF
--- a/tutorial/creating_packages/add_dependencies_to_packages.rst
+++ b/tutorial/creating_packages/add_dependencies_to_packages.rst
@@ -39,7 +39,7 @@ Let's check the relevant parts:
 
         def validate(self):
             check_min_cppstd(self, "11")
-            check_max_cppstd(self, "23")
+            check_max_cppstd(self, "20")
 
         def requirements(self):
             self.requires("fmt/8.1.1")
--- a/tutorial/creating_packages/add_dependencies_to_packages.rst
+++ b/tutorial/creating_packages/add_dependencies_to_packages.rst
@@ -72,7 +72,7 @@ Let's check the relevant parts:
   non-supported configurations. Here, we call the
   :ref:`check_min_cppstd()<conan_tools_build_check_min_cppstd>` and
   :ref:`check_max_cppstd()<conan_tools_build_check_max_cppstd>` to check that we are using at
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Not sure about this change, the idea of the limit is to make it fail if the cppstd is above a limit. Like the source code could have a fatal compilation error with some #ifdef if the cppstd is higher than 14, for example, forcing users to use c++14 and not higher

czoido:
> Not sure about this change, the idea of the limit is to make it fail if the cppstd is above a limit. Like the source code could have a fatal compilation error with some #ifdef if the cppstd is higher than 14, for example, forcing users to use c++14 and not higher

That's my fault, I suggested 23  in the PR for examples to just use a standard that would never fail in CI but stills show the tool functionality but obviously that's not the best choice, better put 20.

tutorial/creating_packages/add_dependencies_to_packages.rst
```diff
@@ -39,7 +39,7 @@ Let's check the relevant parts:
 
         def validate(self):
             check_min_cppstd(self, "11")
-            check_max_cppstd(self, "14")
+            check_max_cppstd(self, "23")
```
czoido, line 42:
```suggestion
            check_max_cppstd(self, "20")
```

tutorial/creating_packages/add_dependencies_to_packages.rst
```diff
@@ -72,7 +72,7 @@ Let's check the relevant parts:
   non-supported configurations. Here, we call the
   :ref:`check_min_cppstd()<conan_tools_build_check_min_cppstd>` and
   :ref:`check_max_cppstd()<conan_tools_build_check_max_cppstd>` to check that we are using at
-  least C++11 and at most C++14 standards in our settings.
+  least C++11 and at most C++23 standards in our settings.
```
czoido, line 75:
```suggestion
  least C++11 and at most C++20 standards in our settings.
```

czoido (approved):

PR #3449, opened 2023-11-06
conf is read-only

memsharded:
Docs for https://github.com/conan-io/conan/pull/14932

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -73,8 +73,10 @@ The profile entries have priority.
           self.output.info("Custom var1: %s" % self.conf.get("user.custom.var1"))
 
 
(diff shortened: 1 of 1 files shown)


reference/conanfile/attributes.rst
```diff
@@ -73,6 +73,10 @@ The profile entries have priority.
           self.output.info("Custom var1: %s" % self.conf.get("user.custom.var1"))
 
 
+NOTE: The ``conf`` attribute is a **read-only** attribute. It can only be defined in profiles and command lines, but it should never be set by recipes.
+Recipes can only read its value via ``self.conf.get()`` method.
```
czoido, line 77:
```suggestion
.. note::

    The ``conf`` attribute is a **read-only** attribute. It can only be defined in profiles and command lines, but it should never be set by recipes.
    Recipes can only read its value via ``self.conf.get()`` method.
```

czoido (approved):

PR #3479, opened 2023-12-18
Update Python version policy

SSE4:
version policy states: `Conan will deprecate support for Python versions 1 year after those versions have been declared End Of Life (EOL).`
Python 3.6 EOL data is 2021-12-23 (see [PEP 494](https://peps.python.org/pep-0494/)).
so this is already more than one year since Python 3.6 EOL, policy should now mentioned Python 3.7.

### DIFF
--- a/introduction.rst
+++ b/introduction.rst
@@ -93,7 +93,7 @@ There are some things that are not included in this commitment:
 - Other tools and repositories outside of the Conan client
 
 
-Conan needs Python>=3.6 to run. Conan will deprecate support for Python versions 1 year after those versions have been declared End Of Life (EOL).
+Conan needs Python>=3.7 to run. Conan will deprecate support for Python versions 1 year after those versions have been declared End Of Life (EOL).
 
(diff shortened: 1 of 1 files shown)


memsharded (commented):
We will not merge this yet. 
Conan 2.0 still supports and is tested with Python 3.6, and we plan to keep it a bit more.

The policy is still good to have, we might need to deprecate it, but at the moment we are not mandating this.

This PR can be closed, thanks very much for contributing

SSE4:
> we plan to keep it a bit more.

can we define "more" in docs then to make it crystal clear for readers? otherwise now it looks like conan violates its own policy for unclear reason.

memsharded:
> can we define "more" in docs then to make it crystal clear for readers? otherwise now it looks like conan violates its own policy for unclear reason.

We are doing a best effort to keep it to support some users that are still tied to Python 3.6, but can't guarantee how long we will be able to do it. At that point in time, we might decide to move forward and deprecate 3.6. So it is not possible to define "more" in months or whatever, we will do it for some time, and depending on the circumstances, users, python ecosystem, team, etc, we will make the decision when we think it is time to move forward.

PR #3484, opened 2023-12-18
Add `core:warnings_as_errors` conf docs

RubenRBS:
For https://github.com/conan-io/conan/pull/15149


reference/conanfile/running_and_output.rst
```diff
@@ -30,6 +30,10 @@ so running with ``-vwarning` will output calls to ``warning()`` and ``error()``,
 
 Note that these methods return the output object again, so that you can chain output calls if needed.
 
+Using the ``core:warnings_as_errors`` conf, you can make Conan raise an exception when either a warning or error is printed.
```
memsharded, line 33: I am missing some example of syntax like ``core:warning_as_errors=["network", "..."]`` (is it a list?)

reference/commands/config.rst
```diff
@@ -216,6 +216,7 @@ Displays all the Conan built-in configurations. There are 2 groups:
     core:non_interactive: Disable interactive user input, raises error if input necessary
     core:required_conan_version: Raise if current version does not match the defined range.
     core:skip_warnings: Do not show warnings in this list
+    core:warnings_as_errors: Treat warnings as errors
```
memsharded, line 219: maybe even here, add what possible values and format it can take

reference/conanfile/running_and_output.rst
memsharded: Yeah, it is not very clear, the ``skip_warnings`` is another config, it is not evident what format ``warnings_as_errors`` gets, is it a boolean?. It is a new ``conf`` with its own values.

memsharded (approved):

reference/conanfile/running_and_output.rst
```diff
@@ -30,6 +30,10 @@ so running with ``-vwarning` will output calls to ``warning()`` and ``error()``,
...
+This is useful to make sure that recipes are not printing unexpected warnings or errors.
+Additionally, you can skip which warnings trigger an exception :ref:`with the *core:skip_warnings* conf<reference_config_files_global_conf_skip_warnings>`.
```
memsharded, line 35: this needs fixing with the new approach

reference/conanfile/running_and_output.rst
memsharded: I think that now that it is a list, it really deserves a bit of explanation and example

czoido (approved):

PR #3488, opened 2023-12-18
Add environment information to CMakePresets docs

czoido:
Docs for https://github.com/conan-io/conan/pull/15192


reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -504,6 +523,7 @@ CMakeToolchain is affected by these ``[conf]`` variables:
 - **tools.cmake.cmaketoolchain:system_processor** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_PROCESSOR``.
 - **tools.cmake.cmaketoolchain:toolset_arch**: Will add the ``,host=xxx`` specifier in the ``CMAKE_GENERATOR_TOOLSET`` variable of ``conan_toolchain.cmake`` file.
 - **tools.cmake.cmake_layout:build_folder_vars**: Settings and Options that will produce a different build folder and different CMake presets names.
+- **tools.cmake.cmaketoolchain:presets_environment**: Set to ``'disable'`` to prevent the addition of the environment section to the generated CMake presets.
```
memsharded, line 526: disable or disabled?

memsharded (approved):

reference/tools/cmake/cmaketoolchain.rst
czoido, line 526:
```suggestion
- **tools.cmake.cmaketoolchain:presets_environment**: Set to ``'disabled'`` to prevent the addition of the environment section to the generated CMake presets.
```

memsharded (approved):
I think it is good for now, to see how it goes, if it works well, etc.
But later, I think we want:

- to integrate a bit more in the tutorial, maybe remove some examples call to ``conanbuild.bat|sh`` that could be resolved by this feature
- Define a full example that shows how this is used, and doesn't require calling conanbuild/conanrun scripts

db4:
Can this feature be backported to Conan v1?

PR #3492, opened 2023-12-18
cmakedeps_legacy

memsharded:
Docs for https://github.com/conan-io/conan/pull/15207

### DIFF
--- a/reference/tools/cmake/cmakedeps.rst
+++ b/reference/tools/cmake/cmakedeps.rst
@@ -70,7 +70,7 @@ Generated files
 - **Other necessary *.cmake**:  files like version, flags and directory data or configuration.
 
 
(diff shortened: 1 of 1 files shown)


reference/tools/cmake/cmakedeps.rst
```diff
@@ -70,6 +70,9 @@ Generated files
 - **Other necessary *.cmake**:  files like version, flags and directory data or configuration.
 
 
+Note that it will also generate a **conandeps_legacy.cmake** file. This is a file that provides a behavior similar to the Conan 1 ``cmake`` generator, allowing to include this file with ``include(${CMAKE_BINARY_DIR}/generators/conandeps_legacy.cmake)``, and providing a single CMake ``CONANDEPS_LEGACY`` variable that allows to link with all the direct and transitive dependencies without explicitly enumerating them like: ``target_link_libraries(app ${CONANDEPS_LEGACY})``. This is a convenience provided for Conan 1.X users to upgrade to Conan 2.0 without changing their overall developer flow, but it is not recommended otherwise, and using the CMake canonical flow of explicitly use ``find_package()`` and ``target_link_libraries(... pkg1::pkg1 pkg2::pkg2)`` using targets is the correct approach.
```
RubenRBS, line 73:
```suggestion
Note that it will also generate a **conandeps_legacy.cmake** file. This is a file that provides a behavior similar to the Conan 1 ``cmake`` generator, allowing to include this file with ``include(${CMAKE_BINARY_DIR}/generators/conandeps_legacy.cmake)``, and providing a single CMake ``CONANDEPS_LEGACY`` variable that allows to link with all the direct and transitive dependencies without explicitly enumerating them like: ``target_link_libraries(app ${CONANDEPS_LEGACY})``. This is a convenience provided for Conan 1.X users to upgrade to Conan 2.0 without changing their overall developer flow, but it is not recommended otherwise, and using the CMake canonical flow of explicitly using ``find_package()`` and ``target_link_libraries(... pkg1::pkg1 pkg2::pkg2)`` with targets is the correct approach. 
```

RubenRBS (approved):
Looks good

czoido (approved):

PR #3493, opened 2023-12-18
Docs for source_buildenv

danimtb:
Just documenting the opt-out attribute here.

Maybe we can add an example to show the usage of tool requires in the source method?

Docs for https://github.com/conan-io/conan/pull/15153

### DIFF
--- a/reference/conanfile/attributes/sources.inc
+++ b/reference/conanfile/attributes/sources.inc
@@ -112,3 +112,24 @@ For example, a *conandata.yml* with information about sources that looks like th
 
     def source(self):
         tools.get(**self.conan_data["sources"][self.version])
+
+source_buildenv
+---------------
+
+Boolean attribute to opt-out from using tool requires in the `source()` method.
+
+Setting this attribute to `False` (default value `True`) will prevent Conan from inyecting the VirtualBuildEnv generated environment from tool requires
+when executing the `source()` method.
+
+.. code-block:: python
+   :emphasize-lines: 4,5,8
+
+    class MyConan:
+       name = "mylib"
+       version = "1.0.0"
+       source_buildenv = False
+       tool_requires = "7zip/1.2.0"
+
+    def source(self):
+        tools.get(**self.conan_data["sources"][self.version])
+        self.run("7z x *.zip -o*")  ## Error! tool require not available
--- a/reference/conanfile/attributes/sources.inc
+++ b/reference/conanfile/attributes/sources.inc
@@ -130,6 +130,6 @@ when executing the `source()` method.
        source_buildenv = False
        tool_requires = "7zip/1.2.0"
 
-    def source(self):
-        get(self, **self.conan_data["sources"][self.version])
-        self.run("7z x *.zip -o*")  ## Error! tool require not available
+        def source(self):
+            get(self, **self.conan_data["sources"][self.version])
+            self.run("7z x *.zip -o*")  ## Error! tool require not available
--- a/reference/conanfile/attributes/sources.inc
+++ b/reference/conanfile/attributes/sources.inc
@@ -130,6 +130,6 @@ when executing the `source()` method.
        source_buildenv = False
        tool_requires = "7zip/1.2.0"
 
-        def source(self):
-            get(self, **self.conan_data["sources"][self.version])
-            self.run("7z x *.zip -o*")  ## Error! tool require not available
+       def source(self):
+           get(self, **self.conan_data["sources"][self.version])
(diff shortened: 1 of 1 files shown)


reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
 .. code-block:: python
 
     def source(self):
-        tools.get(**self.conan_data["sources"][self.version])
\ No newline at end of file
+        tools.get(**self.conan_data["sources"][self.version])
```
memsharded, line 114: ``tools.get`` no longer exist in Conan 2? This should have been fixed some time ago
```suggestion
        get(self, **self.conan_data["sources"][self.version])
```

reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
...
+
+source_buildenv
+---------------
+
+Boolean attribute to opt-out from using tool requires in the `source()` method.
+
+Setting this attribute to `False` (default value `True`) will prevent Conan from inyecting the VirtualBuildEnv generated environment from tool requires
```
memsharded, line 121:
```suggestion
Setting this attribute to `False` (default value `True`) will prevent Conan from injecting the ``VirtualBuildEnv`` generated environment from tool requires
```

reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
...
+   :emphasize-lines: 4,5,8
+
+    class MyConan:
+       name = "mylib"
+       version = "1.0.0"
+       source_buildenv = False
+       tool_requires = "7zip/1.2.0"
+
+    def source(self):
+        tools.get(**self.conan_data["sources"][self.version])
```
memsharded, line 134: Conan 1.X legacy code, please remember this is Conan 2.0, we don't even use ``tools.get()`` in ConanCenter in 1.X long time ago.

memsharded (changes requested):

reference/conanfile/attributes/sources.inc
danimtb:
```suggestion
    def source(self):
        get(self, **self.conan_data["sources"][self.version])
```

reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
...
 
     def source(self):
-        tools.get(**self.conan_data["sources"][self.version])
\ No newline at end of file
+        get(self, **self.conan_data["sources"][self.version])
+
+source_buildenv
+---------------
+
+Boolean attribute to opt-out from using tool requires in the `source()` method.
```
czoido, line 119: Maybe something like this would be more accurate?
```suggestion
Boolean attribute to opt-out from from injecting the :ref:`VirtualBuildEnv<conan_tools_env_virtualbuildenv>` generated environment in the `source()` method.
```

reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
...
+
+    class MyConan:
+       name = "mylib"
+       version = "1.0.0"
+       source_buildenv = False
+       tool_requires = "7zip/1.2.0"
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version])
+        self.run("7z x *.zip -o*")  ## Error! tool require not available
```
czoido, line 135:
```suggestion
        def source(self):
            get(self, **self.conan_data["sources"][self.version])
            self.run("7z x *.zip -o*")  ## Error! tool require not available
```

reference/conanfile/attributes/sources.inc
```diff
@@ -111,4 +111,25 @@ For example, a *conandata.yml* with information about sources that looks like th
...
+
+    class MyConan:
+       name = "mylib"
+       version = "1.0.0"
+       source_buildenv = False
+       tool_requires = "7zip/1.2.0"
+
+        def source(self):
+            get(self, **self.conan_data["sources"][self.version])
+            self.run("7z x *.zip -o*")  ## Error! tool require not available
```
czoido, line 135:
```suggestion
       def source(self):
           get(self, **self.conan_data["sources"][self.version])
           self.run("7z x *.zip -o*")  ## Error! tool require not available
```

czoido (approved):

memsharded (approved):

PR #3496, opened 2023-12-19
lock remove

memsharded:
Docs for https://github.com/conan-io/conan/pull/15284

### DIFF
--- a/reference/commands/lock/remove.rst
+++ b/reference/commands/lock/remove.rst
@@ -85,8 +85,8 @@ It is possible to specify different patterns:
 
 The ``conan lock remove`` can be useful for:
 
-- In combination with ``conan lock add`` can be used to force the downgrade of a locked version to an older one. As ``conan lock add``
-  always add and sorts the order, resulting in newer versions with high priority, it is not possible to force going back to an older
-  version with just ``add``. But using first ``conan lock remove``, then ``conan lock add``, it is possible to do so.
+- In combination with ``conan lock add``, it can be used to force the downgrade of a locked version to an older one. As ``conan lock add``
+  always adds and sorts the order, resulting in newer versions with high priority, it is not possible to force going back to an older
+  version with just ``add``. But first using ``conan lock remove``, then ``conan lock add``, it is possible to do so.
 - ``conan lock remove`` can unlock certain dependencies, resulting in an incomplete lockfile, that can be used with ``--lockfile-partial``
   to resolve to the latest available versions for the unlocked dependencies, while keeping locked the rest.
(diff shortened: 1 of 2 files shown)


czoido (approved):

reference/commands/lock/remove.rst
```diff
@@ -0,0 +1,92 @@
...
+- Remove a specific revision: ``--requires=math/1.0#revision``
+- Remove locked dependencies for a given "team" user ``--requires=*/*@team*``
+
+The ``conan lock remove`` can be useful for:
+
+- In combination with ``conan lock add`` can be used to force the downgrade of a locked version to an older one. As ``conan lock add``
+  always add and sorts the order, resulting in newer versions with high priority, it is not possible to force going back to an older
+  version with just ``add``. But using first ``conan lock remove``, then ``conan lock add``, it is possible to do so.
+- ``conan lock remove`` can unlock certain dependencies, resulting in an incomplete lockfile, that can be used with ``--lockfile-partial``
+  to resolve to the latest available versions for the unlocked dependencies, while keeping locked the rest.
```
RubenRBS, line 92:
```suggestion
- In combination with ``conan lock add``, it can be used to force the downgrade of a locked version to an older one. As ``conan lock add``
  always adds and sorts the order, resulting in newer versions with high priority, it is not possible to force going back to an older
  version with just ``add``. But first using ``conan lock remove``, then ``conan lock add``, it is possible to do so.
- ``conan lock remove`` can unlock certain dependencies, resulting in an incomplete lockfile, that can be used with ``--lockfile-partial``
  to resolve to the latest available versions for the unlocked dependencies, while keeping locked the rest.
```

RubenRBS (approved):

reference/commands/lock/remove.rst
```diff
@@ -0,0 +1,92 @@
...
+      --lockfile LOCKFILE   Filename of the input lockfile
+
+
+The ``conan lock remove`` command is able to remove ``requires``, ``build_requires`` or ``python_requires`` items from an existing lockfile.
+
+For example, if we have the following ``conan.lock``:
+
+.. code-block:: bash
+
+  $cat conan.lock
```
RubenRBS, line 38:
```suggestion
  $ cat conan.lock
```

reference/commands/lock/remove.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. code-block:: bash
+
+  $ conan lock remove --requires="math/*" --build-requires=cmake/1.0 --python-requires="*tool/*"
+
+Will result in the following ``conan.lock``:
+
+.. code-block:: bash
+
+  $cat conan.lock
```
RubenRBS, line 67:
```suggestion
  $ cat conan.lock
```

PR #3512, opened 2024-01-11
Add note about settings.os requirement for VirtualRunEnv

shoeffner:
This commit adds a short note that settings.os must be set properly for LD_LIBRARY_PATH etc. to appear.

### DIFF
--- a/reference/tools/env/virtualrunenv.rst
+++ b/reference/tools/env/virtualrunenv.rst
@@ -10,13 +10,9 @@ The launcher contains the runtime environment information, anything that is nece
 the compiled executables and applications. The information is obtained from:
 
     - The ``self.runenv_info`` of the dependencies corresponding to the ``host`` context.
-    - Also automatically deduced from the ``self.cpp_info`` definition of the package, to define ``PATH``,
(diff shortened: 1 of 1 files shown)


reference/tools/env/virtualrunenv.rst
```diff
@@ -13,6 +13,11 @@ the compiled executables and applications. The information is obtained from:
     - Also automatically deduced from the ``self.cpp_info`` definition of the package, to define ``PATH``,
       ``LD_LIBRARY_PATH``, ``DYLD_LIBRARY_PATH`` and ``DYLD_FRAMEWORK_PATH`` environment variables.
 
+.. note::
```
memsharded, line 16: Thanks for the contribution.
I think the clarification is good, but probably doesn't need to be highlighted as a note, as the vast majority of users won't have this problem. Maybe as an extra bullet in the list above?

shoeffner: Thanks for the suggestion, I split the above bullet point to make it clearer but stay more in line with the "The information is obtained from:".

memsharded (approved):

memsharded:
Thanks for your contribution!

PR #3516, opened 2024-01-19
Upgrade to Sphinx 7

RubenRBS:
This finally updates the docu to the latest sphinx release, which fixes a number of issues (For me the most important one is fixing environment pickling so that recompilations are now fast again!)

The look of the theme is a bit different, and I still need to fix up some paths in the layout file, but so far so good!

The sitemap plugin is not working either in the current old version, nor in the PR. I've left it commented out for a later time .

This work was initially triggered because I was done with updating the help messages every time we change the help, so I was trying to add a custom extension (As https://sphinx-argparse.readthedocs.io/en/stable/ does not work with the current Conan setup)


RubenRBS (commented):
Some comments

memsharded (approved):
amazing

czoido:
Closing as already implemented in another PR

Issue #3525, opened 2024-01-25
[question] Are there any examples of custom/user defined templates?


samuel-emrys:
### What is your question?

I have a couple of questions about custom templates:

1. Are there any examples of these and how to define them? I found https://docs.conan.io/2/reference/commands/new.html#custom-templates but it seems to be pretty light on on getting started
2. Can custom templates be distributed through a `conan config`-installable repository? i.e., can I add some structure to `extensions` to get my templates to be identified by the `conan new` command?

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

memsharded:
Documented in https://github.com/conan-io/docs/pull/3526, thanks!

PR #3526, opened 2024-01-25
New templates docs

memsharded:
Close https://github.com/conan-io/docs/issues/3525

### DIFF
--- a/reference/commands/new.rst
+++ b/reference/commands/new.rst
@@ -162,7 +162,56 @@ inside the generated ``game()`` function.
 Custom templates
 ----------------
 
-There's also the possibility to create your own templates by passing a path to your template directory,
-both as an absolute path, or relative to your Conan home folder. This directory should contain Jinja2 templates,
-which will produce your desired template structure. You can use custom variables that will be needed to be passed
-as ``name`` and ```version`` does, or use your custom variables.
+There's also the possibility to create your own templates. Templates in the Conan home should be 
+located in the ``templates/command/new`` folder, and each template should create a new folder, being
+the name of the folder the name of the template. If we create the ``templates/command/new/mytemplate``
+folder, the command will be called with:
+
+
+.. code-block:: text
+
+    $ conan new mytemplate
+
+
+As other files in the Conan home, you can manage these templates with ``conan config install``, putting them
+in a git repo or http server and sharing them with your team. It is also possible to use templates from 
+any folder, just passing the full path to the template in the ``conan new <full_path>``, but in general it
+is more convenient to manage them in the Conan home.
+
+The folder can contain as many files as desired. Both the filenames and the contents of the files can be
+templatized using Jinja2 syntax. The command ``-d/--define`` arguments will define the ``key=value`` inputs
+to the templates. 
+
+There are some special ``-d/--defines`` names. The ``name`` one is always mandatory. The ``conan_version``
+definition will always be automatically defined. The ``requires`` and ``tool_requires`` definitions, if existing, 
+will be automatically converted to lists. The ``package_name`` will always be defined, by default equals to ``name``.
+
+The file contents will be like (Jinja2 syntax):
+
+.. code-block:: python
+
+   class Conan(ConanFile):
+      name = "{{name}}"
+      version = "{{version}}"
+      license = "{{license}}"
+
+
+And it will require passing these values:
+
+.. code-block:: text
+
+    $ conan new mytemplate -d name=pkg -d version=0.1 -d license=MIT
+
+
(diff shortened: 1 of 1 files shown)


reference/commands/new.rst
```diff
@@ -162,7 +162,56 @@ inside the generated ``game()`` function.
...
+any folder, just passing the full path to the template in the ``conan new <full_path>``, but in general it
+is more convenient to manage them in the Conan home.
+
+The folder can contain as many files as desired. Both the filenames and the contents of the files can be
+templatized using Jinja2 syntax. The command ``-d/--define`` arguments will define the ``key=value`` inputs
+to the templates. 
+
+There are some special ``-d/--defines`` names. The ``name`` one is always mandatory. The ``conan_version``
+definition will always be automatically defined. The ``requires`` and ``tool_requires`` definitions, if existing, 
+will be automatically converted to lists. The ``package_name`` will always be defined, by default equals to ``name``.
```
franramirez688, line 187: IMO, this is a bit confusing for me. I'd not mention anything about these variables (at least at this point) because they don't appear either before explanations or examples below.

reference/commands/new.rst
```diff
@@ -162,7 +162,56 @@ inside the generated ``game()`` function.
...
+
+    $ conan new mytemplate -d name=pkg -d version=0.1 -d license=MIT
+
+
+For variable filenames, the filenames themselves can have Jinja2 syntax. For example if we store a file with
+named literally ``templates/command/new/mytemplate/{{name}}``, with the brackets in the filename, when running
+
+.. code-block:: text
+
+    $ conan new mytemplate -d name=file.txt
```
franramirez688, line 211: I have a doubt here. How does it fit with the example above?
```python
   class Conan(ConanFile):
      name = "{{name}}"
      version = "{{version}}"
      license = "{{license}}"
```
Will the name be `file.txt` too?

reference/commands/new.rst
```diff
@@ -162,7 +162,56 @@ inside the generated ``game()`` function.
...
+
+
+a filename called ``file.txt`` will be created.
+
+If there are files in the template to not be rendered with Jinja2, like image files, then their names should be
+added to a file called ``not_templates`` inside the template directory, one filename per line.
```
franramirez688, line 217: Missing how the file tree would look like.

franramirez688 (changes requested):
@memsharded IMHO, I think that we should improve a little bit this section. I'm missing something like going from scratch to something bigger and showing the file structure all the time. Let's discuss it if any doubts.

reference/commands/new.rst
```diff
@@ -162,7 +162,56 @@ inside the generated ``game()`` function.
...
+folder, the command will be called with:
+
+
+.. code-block:: text
+
+    $ conan new mytemplate
+
+
+As other files in the Conan home, you can manage these templates with ``conan config install``, putting them
+in a git repo or http server and sharing them with your team. It is also possible to use templates from
```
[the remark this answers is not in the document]
memsharded: This would be ``a http server``, not ``an``

reference/commands/new.rst
```diff
@@ -162,7 +162,94 @@ inside the generated ``game()`` function.
...
+located in the ``templates/command/new`` folder, and each template should have a folder named like the template one. If we create the ``templates/command/new/mytemplate``
+folder, the command will be called with the following:
+
+
+.. code-block:: bash
+
+    $ conan new mytemplate
+
+
+As other files in the Conan home, you can manage these templates with ``conan config install <url>``, putting them
```
RubenRBS, line 175:
```suggestion
As with other files in the Conan home, you can manage these templates with ``conan config install <url>``, putting them
```

reference/commands/new.rst
```diff
@@ -162,7 +162,94 @@ inside the generated ``game()`` function.
...
+in a git repo or a http server and sharing them with your team. It is also possible to use templates from
```
RubenRBS, line 176: Pedantic review warning ⚠️ 


```suggestion
in a git repo or an http server and sharing them with your team. It is also possible to use templates from 
```

@franramirez688 was correct here! (But don't tell them, they might get used to it!) As an acronym, it's spelled out letter by letter, and `h` has a vowel as its first phoneme, thus needs an `an` :)

RubenRBS (approved):
Nice! Minor nitpicks, otherwise looks good :)

reference/commands/new.rst
memsharded: Thats right, thanks for the correction 😄

franramirez688 (approved):
Nice!!!

PR #3563, opened 2024-02-03
Fix wrongly assigned code-block and caption of CMakeLists.txt

yudi-matsuzake:


memsharded (approved):
Thanks very much for the fix @yudi-matsuzake !

Merging it, it will be in next release.

PR #3579, opened 2024-02-13
Fixup cmake presets in example, use conan-release, not conan-conan-release

paulharris:


RubenRBS (approved):
Thanks a lot for taking the time to fix the issue :)

PR #3583, opened 2024-02-14
Add section about CONAN_NO_DEFAULT_REMOTES

DailyMats:


reference/env_vars.rst
```diff
@@ -314,6 +314,13 @@ This variable, however, is useful for shared CI servers where the stored value c
 assign a random generated string to this value for each of the builds and configure your server to expire
 tokens, this will make the value stored in the database harder to crack.
 
+CONAN_NO_DEFAULT_REMOTES
```
memsharded, line 317: I am afraid this is very legacy, this is not part anymore of Conan 1.X, are you sure you are in latest Conan 1.62 version?

DailyMats: You mean the adding of a default remote? It's still happening with 1.62, yes. For example with `conan export`:
```
$ conan --version
Conan version 1.62.0
$ conan export conanfiles/re2-conanfile.py
WARN: Remotes registry file missing, creating default one in /Users/xxx/.conan/remotes.json
ERROR: conanfile didn't specify name
```

memsharded: I didn't see there is an associated PR. Let's follow the conversation in the PR

PR #3586, opened 2024-02-14
Clean up some of the docs

RubenRBS:
These are some minor issues that I found while fixing a few of the ones that @ErniGH reported when reading the tutorial

### DIFF
--- a/tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
+++ b/tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
@@ -190,9 +190,12 @@ platform without adding more changes.
 
         def requirements(self):
             self.requires("zlib/1.2.11")
+            if self.settings.os == "Windows":
+                self.requires("base64/0.4.0")
 
         def build_requirements(self):
-            self.tool_requires("cmake/3.22.6")
+            if self.settings.os != "Windows":
+                self.tool_requires("cmake/3.22.6")
 
         def layout(self):
             # We make the assumption that if the compiler is msvc the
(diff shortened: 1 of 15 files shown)


tutorial/versioning/lockfiles.rst
```diff
@@ -315,9 +315,12 @@ configuration, and then merge them:
 This multiple-clean + merge operation is not something that developers should do, only CI
 scripts, and for some advanced CI flows that will be explained later.
 
+.. note:
+
+   It is possible to lock down to package revisions, but this would be not recommended for
+   most use cases, and should only be used in extreme and problematic cases.
+
```
memsharded, line 322:
```suggestion
```

We have even removed this from the command help and from the docs, it is not mentioned anywhere.  So it is not really possible at the moment.

tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
```diff
@@ -190,9 +190,12 @@ platform without adding more changes.
 
         def requirements(self):
             self.requires("zlib/1.2.11")
+            if self.settings.os == "Windows":
```
memsharded, line 193: This needs to be reviewed by @czoido, cause this tutorial has always been a bit challenging to follow in sync the code and the docs

memsharded (commented):
Great job

tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
RubenRBS: This is in fact an update to the docs to sync them back :)

czoido (approved):

PR #3589, opened 2024-02-15
Add custom deployer for metadata example

RubenRBS:


devops/metadata.rst
```diff
@@ -340,7 +340,8 @@ downloading it, and copying it to our current folder:
 
 .. seealso::
 
-   - TODO: Examples how to collect the metadata of a complete dependency graph with some custom deployer or command
+   - Check an :ref:`example on how to collect the metadata of a complete dependency graph with some custom deployer<examples_extensions_deployers_metadata>`
```
memsharded, line 343: The challenge with this example is that there will be issues with users expectations:
- ``conan install`` will install dependencies, but it will not install metadata from the servers
- Then the metadata is NOT in the Conan cache
- Gathering the metadata will not collect anything, only metadata that has been created locally.

The solution is that the metadata has to be downloaded from the server with ``conan download --metadata`` explicitly first, then collected.

memsharded (changes requested):
This needs some further elaboration, because the very common case in which users will want to gather metadata from servers, is not covered by this.

Probably we need to make sure how it looks like in a test first, the UX etc.

PR #3595, opened 2024-02-20
Use Conan 2 instead of Conan 2.X

czoido:
Not to merge


memsharded (approved):
All ``examples2.0`` -> ``examples2`` would be good to change too

PR #3598, opened 2024-02-20
Fix channel typo

stevn:


memsharded (approved):
Thanks for fixing this!

Issue #3612, opened 2024-03-04
Missing read more sections in the docs


czoido:
Collecting missing references from the docs to start completing missing parts:

[In cross_building tutorial](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/consuming_packages/cross_building_with_conan.rst?plain=1#L217-L219):

- Cross-build using a tool_requires
- How to require test frameworks like gtest: using ``test_requires``
- Using Conan to build for iOS

In [building for multiple configurations tutorial](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/consuming_packages/different_configurations.rst?plain=1#L324-L327)

- Installing configurations with conan config install
- VS Multi-config
- Example about how settings and options influence the package id
- Using patterns for settings and options

[In understanding the flexibility of using conanfile.py vs conanfile.txt
](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst?plain=1#L354-L355)
- Importing resource files in the generate() method
- Conditional generators in configure()

[Using build tools as Conan packages](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/consuming_packages/use_tools_as_conan_packages.rst?plain=1#L243-L245)

- Using MinGW as tool_requires
- Using tool_requires in profiles
- Using conf to set a toolchain from a tool requires

[Create your first Conan package](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/creating_packages/create_your_first_package.rst?plain=1#L322-L324)

- Create your first Conan package with Autotools.
- Create your first Conan package with Meson.
- Create your first Conan package with Visual Studio.

[Preparing the build](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/creating_packages/preparing_the_build.rst?plain=1#L196-L197)

- Use the ``generate()`` method to import files from dependencies.
- More based on the examples mentioned above …

[Test_conan_packages](https://github.com/conan-io/docs/blob/baa641deeb9e7107fead0488c04f17d85f3fbcd4/tutorial/creating_packages/test_conan_packages.rst?plain=1#L164)

[...]

PR #3620, opened 2024-03-06
Add a note about copy resources in the generate method

czoido:
Part of completing this: https://github.com/conan-io/docs/issues/3612

### DIFF
--- a/tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
+++ b/tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
@@ -378,7 +378,7 @@ Here's an example that shows how to copy all resources from a dependency's
 
 
 Then, after the ``conan install`` step, all those resource files will be copied locally,
-allowing you to use them in your project's build process. For a comprehensive example of
+allowing you to use them in your project's build process. For a complete example of
 how to import files from a package in the ``generate()`` method, you can refer to the
 `blog post about using the Dear ImGui library
 <https://blog.conan.io/2019/06/26/An-introduction-to-the-Dear-ImGui-library.html>`, which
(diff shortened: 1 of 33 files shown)


tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
```diff
@@ -346,10 +346,48 @@ Windows or that you want to use the system's CMake installation instead of using
                 self.tool_requires("cmake/3.22.6")
 
 
+Use the generate() method to copy resources from packages
```
memsharded, line 349: The section is great, my only question is if it deserves to be part of the main thread of the tutorial. Is it an important use case, something that >50% of users need to learn the Conan basics? Or maybe it should be moved to elsewhere?

memsharded (approved):

tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
czoido: That was exactly my question too... I opted for this because I thought that maybe it could be useful for someone checking what are the benefits of the conanfile.py vs the txt because this can't be done in the txt... but this could be moved to a complete example elsewhere...

memsharded: Yeah, I see... I think it can stay here, but maybe add a note clarifying that this isn't necessary to copy libraries, static or shared, as Conan generators and virtualenvs can use the libraries from their locations without copying it over and over.

tutorial/consuming_packages/the_flexibility_of_conanfile_py.rst
```diff
@@ -346,10 +346,48 @@ Windows or that you want to use the system's CMake installation instead of using
...
+---------------------------------------------------------
+
+In some scenarios, Conan packages include files that are useful or even necessary for the
+consumption of the libraries they package. These files can range from configuration files,
+assets, to specific files required for the project to build or run correctly. Using the
+:ref:`generate() method<reference_conanfile_methods_generate>` method you can copy these
```
RubenRBS, line 355:
```suggestion
:ref:`generate() method<reference_conanfile_methods_generate>` you can copy these
```

RubenRBS (approved):
Nice!

Issue #3066, opened 2023-03-06
Create your first Conan package with Meson example


czoido:
An example [for the examples section](https://docs.conan.io/2/examples.html), similar to this part of the tutorial https://docs.conan.io/2/tutorial/creating_packages/create_your_first_package.html but for Meson.

memsharded:
I have been doing some support in https://github.com/conan-io/conan/issues/13344, and it needs some fix in the associated PR to work with ``tool_requires`` (it works if meson+pkgconf+ninja are in the environment, but not in Conan packages):

memsharded:
Closed by https://github.com/conan-io/docs/pull/3621

PR #3621, opened 2024-03-06
meson create package

memsharded:
Close https://github.com/conan-io/docs/issues/3066


examples/tools/meson/create_your_first_package.rst
```diff
@@ -0,0 +1,162 @@
...
+
+  .. code-block::
+
+    include(default)
+    
+    [tool_requires]
+    meson/[*]
+    pkgconf/[*]
+
+  and use it in ``conan create . -pr=myprofile``
```
RubenRBS, line 111: Maybe a ref to the relevant profile section?

RubenRBS (approved):
Minor suggestion but not a blocker :)

examples/tools/meson/create_your_first_package.rst
czoido, line 111: I agree, as this is for the beginner, maybe we can add a reference to the profiles section in case that addition to the profile sounds weird for them:

```suggestion
We added `Meson` and `pkgconf` as :ref:`tool requirements to the profile <reference_config_files_profiles_tool_requires>`. By executing ``conan create . -pr=myprofile``, those tools will be installed and made available during the package's build process.
```

PR #3623, opened 2024-03-07
How settings/options influence package id

danimtb:


examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. code-block:: python
+
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+        version = "1.0.0"
+        options = {"diagnostics_level": [0, 1, 2]}
+        default_options = {"diagnostics_level": 0}
```
memsharded, line 136: This is no longer recommended usage anymore, better use ``conf`` for this.

Removing options should be way less common nowadays.
I'd recommend documenting this possibility just as such "If you are using options and removing them from ``package_id``, consider using ``conf`` instead.

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,145 @@
+.. _examples_conanfile_package_id_settings_options:
+
+How settings and options of a recipe influence its pacakge ID
```
memsharded, line 3: The ``conan list`` command can be very useful to illustrate this section, using it after the ``conan create`` it is possible to see the different binaries with different settings, not only hte "package-id" which in isolation is less explicit

memsharded (commented):
To be linked from https://docs.conan.io/2/reference/binary_model/package_id.html too

examples/conanfile/package_id/settings_and_options.rst
danimtb: Sure. I have left the example but recommended the conf usage at the end. Thanks!

examples/conanfile/package_id/settings_and_options.rst
danimtb: Done

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+                  9a4eb3c8701508aa9458b1a73d0633783ecc2270
+                    info
+                      settings
+                        os: Linux
+
+If we do the same thing with Windows, now the package ID will be diffent:
+
+.. code-block:: bash
+
+    $ conan create . --settings os=Linux
```
czoido, line 52:
```suggestion
    $ conan create . --settings os=Windows
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+                  9a4eb3c8701508aa9458b1a73d0633783ecc2270
+                    info
+                      settings
+                        os: Linux
+                  ebec3dc6d7f6b907b3ada0c3d3cdc83613a2b715
+                    info
+                      settings
+                        os: Windows
+
+Whenever a value of the settings or subsettings changes, the pacakge ID will be different to reflect that.
```
czoido, line 72:
```suggestion
Whenever a value of the settings or subsettings changes, the package ID will be different to reflect that.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+.. code-block:: python
+
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+        version = "1.0.0"
+        settings = "os", "arch", "compiler", "build_type"
+
+Now, compiling a package with different compiler versions will result into different pacakge IDs:
```
czoido, line 86:
```suggestion
Now, compiling a package with different compiler versions will result into different package IDs:
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+        version = "1.0.0"
+        settings = "os"  # Only OS setting affects the package ID
+
+If we generate a package from this recipe for Linux we will get the following package ID:
+
+.. code-block:: bash
```
czoido, line 30: I think it would be superuseful to highlight the package id's lines in all the code blocks

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+        settings = "os", "arch", "compiler", "build_type"
+
+        def build(self):
+            # self.settings.compiler value will be used here to compile
+
+        def package_id(self):
+            # Remove compiler setting from package ID
+            del self.info.settings.compiler
+
+Why not removing the setting from the `settings` attribute? Because the compiler value is still needed in the `build()` method to perform the copilation of the executable.
```
czoido, line 148:
```suggestion
Why not removing the setting from the `settings` attribute? Because the compiler value is still needed in the `build()` method to perform the compilation of the executable.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "pkg"
+        version = "1.0.0"
+        options = {"shared": [True, False]}
+        default_options = {"shared": True}
+
+As in the previous case with settings, the different values of an option will incluence the package ID and therefore, generate different packages depending on it.
```
czoido, line 169:
```suggestion
As in the previous case with settings, the different values of an option will influence the package ID and therefore, generate different packages depending on it.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,202 @@
...
+    $ conan create . --options shared=True
+    ...
+    pkg/1.0.0: Package '1744785cb24e3bdca70e27041dc5abd20476f947' created
+
+    $ conan create . --options shared=False
+    ...
+    pkg/1.0.0: Package '55c609fe8808aa5308134cb5989d23d3caffccf2' created
+
+In the same way, there might be options that are needed as input in a recipe to generate a package but that they shouldn't be taken into account in the package ID.
+To do so, the recipe can remove the option in the `pacakge_id()` method:
```
czoido, line 182:
```suggestion
To do so, the recipe can remove the option in the `package_id()` method:
```

RubenRBS (commented):
Nice!

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,206 @@
+.. _examples_conanfile_package_id_settings_options:
+
+How settings and options of a recipe influence its package ID
```
memsharded, line 3: Should we define some convention to name package-id? package ID, package-id, package_id?

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,206 @@
...
+=============================================================
+
+In Conan, a package ID is a unique identifier for a package that takes into account all the factors that affect its contents.
```
memsharded, line 6:
```suggestion
In Conan, a package ID is a unique identifier for a package binary that takes into account all the factors that affect its contents.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,206 @@
...
+        settings = "os", "arch", "compiler", "build_type"
+
+        def build(self):
+            # self.settings.compiler value will be used here to compile
+
+        def package_id(self):
+            # Remove compiler setting from package ID
+            del self.info.settings.compiler
+
+Why not removing the setting from the `settings` attribute? Because the compiler value is still needed in the `build()` method to perform the compilation of the executable.
```
memsharded, line 151: I think this is at least controversial, and should be at least clarified, but probably changed. 

There is a much different use case if we are concerned about a ``tool_require``, like ``cmake/3.25``, or if we are developing our own final applications that is our product. For our own product I'd never recommend doing this, but this section is written in a way that it seems this must be done for all executables.

I'd suggest to elaborate it as two approaches:
- approach 1: removing the ``compiler``: why and when (tool-requires)
- approach 2: not removing the ``compiler``: why and when (my own apps)

And present the 2 approaches as equal alternatives

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,206 @@
...
+        name = "pkg"
+        version = "1.0.0"
+        options = {"diagnostics_level": [0, 1, 2]}
+        default_options = {"diagnostics_level": 0}
+
+        def build(self):
+            # self.options.diagnostics_level value will be used here
+
+        def package_id(self):
+            # Remove diagnostics_level option from package ID
```
memsharded, line 202: I think this direct removal of the option is still misleading. If the options are being used in the ``build()`` how is it that its not producing a different binary? And if it is not used at ``build()``, then why not removing it earlier in ``config_options()`` or ``configure()``?

I think I'd still prefer the opposite, the lines below are not that evident or explicit, and it would be better to say: "if you have options and you want to remove them from package_id(), then probably they are not options, so better use ``conf``", instead of just putting the code here, which looks like a legitimate, recommended and blessed usage.

examples/conanfile/package_id/settings_and_options.rst
jcar87: Agree here - the setting should be in the `settings` attribute "if it is actually needed at any point". If it's not needed, I think the cleanest, easiest, way of handling it is just by not adding is an an attribute.

I would expand on this with actual examples - maybe add some CMake to the `build()` method or something (where the compiler setting is needed), and have another example of "building" a python only app (like meson) where the `compiler` setting is not needed in the slightest.

examples/conanfile/package_id/settings_and_options.rst
jcar87: I can think of 3 patterns for options (there may be more):
1. The ones that affect what is being built (so they should affect package ID) - this is the most common case
2. The ones that don't affect the actual product, but control something that only makes sense during the `build()` phase - think the verbosity of the boost build configurator, or "verbose makefile" or something of the likes - I think that's the example being shown here with `diagnostics_level` - perhaps it could
3. options that don't affect either (1) or (2), but affect consumers (think `INTERFACE` type options in CMake targets). There's a few of these cases. These would typically be evaluated only in the `package_id()` method.

> If the options are being used in the build() how is it that its not producing a different binary?

@memsharded  I think the example being shown here would be what I'm listing as (2) - perhaps it could be clearer.

examples/conanfile/package_id/settings_and_options.rst
memsharded: I wasn't even covering the case where the ``compiler`` is not needed for building the package (python things), only the case where the ``compiler`` is really needed for the build, then the 2 approaches above. But it might be worth clarifiying the 3 different use cases, presenting them at the same level, not prioritizing the removal of settings in ``package_id()``.

examples/conanfile/package_id/settings_and_options.rst
memsharded: Yes, the approach presented here is 2). My main concern is that this shouldn't be presented this way, looking like an almost standard mechanism of passing information to the recipes without affecting their binaries. 

By defaults ``options`` should be binary-affecting options, and for things that are not binary-affecting, then the general recommended way is ``conf``, not options. This is not clear and evident from this writing.

examples/conanfile/package_id/settings_and_options.rst
danimtb: done

examples/conanfile/package_id/settings_and_options.rst
danimtb: I have always thought about it as "the ID of the package", so I noramlly use just "package ID" (although maybe "package's ID" is the most correct one? I don't like it :smile:).  Anyway, "package-id" or "package ID" will be fine for me 👍

examples/conanfile/package_id/settings_and_options.rst
danimtb: Thanks for the comments. Indeed, I agree with you and I don't want to create any controversy here and just expose a clear usage to the user.

If we think this isn't a very common thing for managing options, we can just drop the example. I just wanted to show how options/settings influence the package ID and removing them in the package_id() method was a good example for me to show the user the influence on the result (not only tied to the input values of the options/settings).

Anyway, I'd rather remove it than add complexity with the other examples or cases 👍

examples/conanfile/package_id/settings_and_options.rst
RubenRBS: 👍 for plain package id

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,206 @@
...
+                      compiler: msvc
+                      compiler.cppstd: 14
+                      compiler.runtime: dynamic
+                      compiler.runtime_type: Release
+                      compiler.version: 193
+                      os: Windows
+
+However, there are cases when a setting should not influence the resulting package ID.
+An example of this could be when a project generates an executable binary as a compilation result. In that case, the value of the compiler
+used is not relevant for the package and we can remove its influence in the package ID like so:
```
memsharded, line 133: Please change this to take into account the feedback of the different use cases below. Thanks!

examples/conanfile/package_id/settings_and_options.rst
memsharded: Yes, lets elaborate a bit more instead of presenting this case, as is. Please describe the 2 strategies:
- Using an ``option`` and remove it (in text, no code). In principle ``options`` are there to affect the ``package_id``, so in general ``conf`` might be preferred. 
- Using a ``conf``. Recommended for general configuration, that end user can select, possibly common for more than one package

memsharded (changes requested):

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,242 @@
...
+        def package_id(self):
+            # Remove compiler setting from package ID
+            del self.info.settings.compiler
+
+Why not removing the setting from the `settings` attribute? Because the compiler value is still needed in the `build()` method to perform the compilation of the executable.
+
+.. note::
+
+    In the case we are generating our own executables (our own apps, not a tool_require), **removing the compiler setting from package ID is not recommended**, as we would always want to know
+    that the pacakge was generated with a specific compiler.
```
memsharded, line 161:
```suggestion
    that the package was generated with a specific compiler.
```

Please check spelling, this is a common misspell.

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,242 @@
...
+
+        def package_id(self):
+            # Remove compiler setting from package ID
+            del self.info.settings.compiler
+
+Why not removing the setting from the `settings` attribute? Because the compiler value is still needed in the `build()` method to perform the compilation of the executable.
+
+.. note::
+
+    In the case we are generating our own executables (our own apps, not a tool_require), **removing the compiler setting from package ID is not recommended**, as we would always want to know
```
memsharded, line 160:
```suggestion
    In the case we are generating our own executables (our own apps, not a ``tool_require``), **removing the compiler setting from package ID is not recommended**, as we would always want to know
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,242 @@
+.. _examples_conanfile_package_id_settings_options:
+
+How settings and options of a recipe influence its package ID
+=============================================================
+
+In Conan, a package ID is a unique identifier for a package binary that takes into account all the factors that affect its contents.
```
czoido, line 6: Maybe something like this would be more accurate?
```suggestion
In Conan, a package ID is a unique identifier for a package binary that takes into account all the factors that affect its binary compatibility.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,242 @@
...
+These factors include recipe options and settings as well as requirements or tool requirements.
+
+Let's see how settings and options affect the package ID and some examples where they should not.
+
+How settings influence the package ID
+-------------------------------------
+
+Settings are development project-wide variables, like the compiler, its version, or the OS itself.
+These variable values have to be defined, they should match the values of our development environment, and they cannot have a default value.
```
czoido, line 15: I did not understand the default value thing in the beginning, maybe this change makes it clearer.
```suggestion
These variable values have to be defined, they should match the values of our development environment, and they cannot have a default value like options do.
```

examples/conanfile/package_id/settings_and_options.rst
```diff
@@ -0,0 +1,242 @@
...
+            self.output.info(f"Using verbosity level: {verbosity})
+            ...
+
+.. code-block:: text
+    :caption: *myprofile*
+
+    [conf]
+    user.my-pkg:verbosity=silent
+
+That way the package ID will be not affected, the recipe will be cleaner (without irrelevant options for package ID) and the input is easily managed via the profile's conf section.
```
czoido, line 242: I think it would be good to add a read more section also here, linking back to other places of interest, like this one: https://docs.conan.io/2/tutorial/creating_packages/configure_options_settings.html

czoido: Also this one: https://docs.conan.io/2/reference/binary_model/package_id.html and linking back would be good too

danimtb: done

PR #3636, opened 2024-03-18
Add docs for name/version being allowed in `build_folder_vars` conf

RubenRBS:
For https://github.com/conan-io/conan/pull/15705


memsharded (approved):
Looks good.

The feature is actually more powerful, it can use any recipe attribute, like ``self.context`` or anything that resembles a string. But I think it is better to document it only for name/version at the moment.

memsharded (approved):

PR #3640, opened 2024-03-19
[PkgConfigDeps][MesonToolchain] Added build context folder mechanism

franramirez688:
Docs for https://github.com/conan-io/conan/pull/15813. (MesonToolchain explanation is covered by the docstring)

### DIFF
--- a/reference/tools/gnu/pkgconfigdeps.rst
+++ b/reference/tools/gnu/pkgconfigdeps.rst
@@ -111,6 +111,8 @@ you can activate it using the **build_context_activated** attribute:
 build_context_suffix
 ^^^^^^^^^^^^^^^^^^^^
 
+*DEPRECATED since Conan 2.2.0: use build_context_folder attribute instead*
+
 When you have the same package as a **build-require** and as a **regular require** it will
 cause a conflict in the generator because the file names of the ``*.pc`` files will
 collide as well as the names, requires names, etc.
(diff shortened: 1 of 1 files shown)


reference/tools/gnu/pkgconfigdeps.rst
```diff
@@ -111,6 +111,8 @@ you can activate it using the **build_context_activated** attribute:
 build_context_suffix
 ^^^^^^^^^^^^^^^^^^^^
 
+*DEPRECATED: use build_context_folder attribute instead*
```
memsharded, line 114: Is this true for ``autotools`` projects too? Or only for ``Meson``? I think only Meson has the capability of specifying a folder for the build-requires?

reference/tools/gnu/pkgconfigdeps.rst
```diff
@@ -139,6 +141,32 @@ renamed:
         pc.generate()
 
 
+build_context_folder
+^^^^^^^^^^^^^^^^^^^^
+
+*New since Conan 2.2.0*
+
+Same concept as the quoted ``build_context_suffix`` attribute above, but this is meant to specify a folder to save
+the `*.pc` files created by all those build requirements listed in the ``build_context_activated`` one:
```
czoido, line 150: I think it would be better to not reference the deprecated attribute to document this one, just explaining this as if the other one did not exist as one day we will remove that one.

franramirez688: Reordered

reference/tools/gnu/pkgconfigdeps.rst
```diff
@@ -122,9 +124,30 @@ also have a **regular require**. Solving this conflict is specially important wh
...
+        # save all the *.pc files coming from the "my_tool" build context and its requirements
+        pc.build_context_folder = "build"  # [generators_folder]/build/
+        pc.generate()
+
+
+
+build_context_suffix
+^^^^^^^^^^^^^^^^^^^^
+
+*DEPRECATED: use build_context_folder attribute instead*
```
memsharded, line 147: It is still not very clear that the deprecation of this and the recommendation to use ``build_context_suffix`` is the way to go also for autotools.  If it is not, maybe we should recommend this for Meson only?

Maybe we need to understand the potential autotools usages of the ``build_context_suffix`` and ``build_context_folder``

reference/tools/gnu/pkgconfigdeps.rst
```diff
@@ -139,6 +162,11 @@ renamed:
         pc.generate()
 
 
+.. important::
+
+    You cannot use this attribute along together with the ``build_context_folder`` one.
```
czoido, line 167:
```suggestion
    This attribute should not be used simultaneously with the ``build_context_folder`` attribute.
```

czoido (approved):

memsharded (approved):

PR #3666, opened 2024-04-05
msvc package

memsharded:
Close https://github.com/conan-io/conan/issues/13043

### DIFF
--- a/examples/dev_flow/msvc_package.rst
+++ b/examples/dev_flow/msvc_package.rst
@@ -1,0 +1,94 @@
+Packaging Visual Studio as a Conan package
+==========================================
+
+It is possible to package the Visual Studio tools into a Conan package and use it later as a ``tool_requires``.
+This can be convenient in some scenarios, for example in CI, or in some developer setups.
+
+In this example we are going to start with a standard Visual Studio 16 2019 installation on disk. The first
+thing we'll do is to rename the folder ``C:/Program Files (x86)/Microsoft Visual Studio/2019`` to
+``C:/Program Files (x86)/Microsoft Visual Studio/2019_2``. This is not strictly necessary, but in this way
+we make sure that CMake or vswhere are not able to locate it, and we can validate better that the consumer
+project is using VS from the Conan package and not from the system.
+
+This is the recipe to package VS 2019:
+
+.. code-block:: python
+    :caption: conanfile.py
+
+    from conan import ConanFile
+    from conan.tools.files import copy, save
+    import os
+
+    class x86_64_windows_msvc2017Conan(ConanFile):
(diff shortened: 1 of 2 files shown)


examples/dev_flow/msvc_package.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+Note a few important things:
+
+- The ``package()`` method is creating a ``msvc_toolchain.cmake`` that defines the ``CMAKE_GENERATOR_INSTANCE``.
+- The ``CMAKE_GENERATOR_INSTANCE`` should point to the location of VS, which is inside the Conan package. But this location
+  will change from machine to machine, so it needs to be generalized.
+- The ``version_vs = "16.11.34601.136"`` needs to match your installation. It can be obtained with ``vswhere``. It is possible
+  to automate that in the recipe, but it is hardcoded for simplicity, like the ``vs_path``.
+- A ``CONAN_MSVC_16_FOLDER`` environment variable, which value is defined in the ``package_info()`` method is used to
```
RubenRBS, line 50:
```suggestion
- A ``CONAN_MSVC_16_FOLDER`` environment variable, whose value is defined in the ``package_info()`` method is used to 
```

RubenRBS (approved):
This is quite useful! Reads nicely, only 1 wording change

danimtb (approved):
good addition!

jcar87:
Just a few notes from me:
* I wouldn't advise renaming ``C:/Program Files (x86)/Microsoft Visual Studio/2019`` to
``C:/Program Files (x86)/Microsoft Visual Studio/2019_2``. This breaks a valid installation, and it's a system location. I would advise packaging the installation with the [Windows Sandbox](https://learn.microsoft.com/en-us/windows/security/application-security/application-isolation/windows-sandbox/windows-sandbox-overview) and retrieving the Conan package outside of it (`conan upload`, `conan cache save`, etc), or alternatively use the [command line interface ](https://learn.microsoft.com/en-us/visualstudio/install/use-command-line-parameters-to-install-visual-studio?view=vs-2022) to the visual studio installer and specify an `--installPath`

* A working installation of Visual C++ would *need* access to the Windows SDK (for windows headers and so on) - if I'm not mistaken, those  paths would not be copied into the package in the given example, so it is not self-contained

* CMake seems to use vswhere.exe and query the Windows registry to locate some things (at least the Windows SDK which is not being packaged here) - those may not work.


I'd say this needs more testing - it looks like it only works on a machine that already has visual studio installed

memsharded:
https://gist.github.com/mmozeiko/7f3162ec2988e81e56d5c4e22cde9977 is a good information, thanks @jcar87

PR #3684, opened 2024-04-17
Add warning for outdated 2.X versions

RubenRBS:
Only if that same file is present in the newest docs version


czoido (approved):
great! 👏

PR #3694, opened 2024-04-23
Add debugging with Visual and PDBs page

juansblanco:
Add example off debugging shared dependencies with Visual Studio, and explain usage of new hook that copies PDBs to package.


examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
+
+.. _examples_dev_flow_debug_visual:
+
+
+Debugging shared libraries with Visual Studio
+=============================================
+
+Conan packages don't contain the necessary information for debugging libraries with Visual Studio by default.
```
czoido, line 8: I think it would be start with a reference to the previous example on how to debug dependencies here: https://docs.conan.io/2/examples/dev_flow/debug/step_into_dependencies.html

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+This information is stored in PDBs that are generated during the compilation of the libraries. When using Conan these
+PDBs are generated in the build folder, which isn't always available, making impossible to debug the dependencies.
+
```
czoido, line 11: I would say something like "...which may not be available..." and explain that the reason why is because you have cleaned you cache of the builds using `conan cache clean` for example...

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+The goal is to have the necessary PDB files in the package folder, so that Visual is able to debug through the
+dependencies of a project. To solve this we created a hook that copies the PDBs created in the build folder to the
+package folder. This behavior can't be forced by default because PDB files are usually larger than the whole package,
+and it would greatly increase the package sizes.
```
czoido, line 15: You can continue here saying that to prevent that situation, you can store the pdb's in the package folder, so even after a `conan cache clean` you will be able to debug the dependencies.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+When using the Visual Studio debugger, it will look for PDBs to load in the following paths:
+
+- The project folder.
+- The original path where the associated file was compiled.
+- The path where Visual is currently finding the compiled file, in our case the DLL in the package folder.
+
+The PDB has by default the same name as its associated file, so Visual will look for it based on the name of the DLL.
+
+
+Making the PDBs available for Visual Studio
```
czoido, line 45: I would add to the title something like "... using a post-package Hook"

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+-------------------------------------------
+
+With the information above we developed a hook that will look through all the DLLs in the package folder and get the
```
czoido, line 48: I would start with something like: "We are going to explain how to take advantage of hooks to prevent this situation to happen... " something in that direction

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+location of where its PDB was generated. To do this we used the ``dumpbin`` tool which is provided with Visual Studio
```
czoido, line 49: Specify that the pdb's are generated in the build folder of the conan cache

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+and can be located for each user through the ``vswhere`` tool. Using ``dumpbin \PDBPATH`` with a DLL as parameter
+we can get the name and path of the PDB which contains the debug information for that DLL. The hook will do this for
+every DLL in the package and then copy the PDB next to its PDB so Visual can find and load it.
+
+The hook is implemented as a post-package hook, which will execute after the package is created through the
+``package()`` method of a recipe.
+
```
czoido, line 56: Maybe we can make the summary at some point of the steps that happen:
- building 
- packaging
- the hook searches for the dll's that are packaged, then gets the pdb's location in the build folder and copies them to the package folder to be next to the dll's

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+The hook is available in the `conan-extensions repository <https://github.com/conan-io/conan-extensions>`_.
+It can be installed with all the other conan extensions with the command:
+
+  .. code-block:: text
+
+      $ conan config install https://github.com/conan-io/conan-extensions.git
+
```
czoido, line 63: I don't think this is relevant, I just would say that to install the hook you have to do the one with the `-sf` and `-tf`

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+.. _examples_dev_flow_debug_visual:
+
+
+Debugging shared libraries with Visual Studio
```
czoido, line 5: Please, check other examples in the examples section of the docs, we usually add code for the example, in the examples2 repo: https://github.com/conan-io/examples2/
Maybe we can consider if it would make sense to add one, with some screenshots and also it would be nice to invite the user to open a project in visual studio, then using conan cache clean to clean all build folders and showing them that with the hook even after doing that you can still debug the libraries (also remark the difference with the previous section of the docs on that matter: https://docs.conan.io/2/examples/dev_flow/debug/step_into_dependencies.html )

examples/dev_flow/debug/debugging_visual.rst
juansblanco: How about "Post-package hook for copying PDBs to package"? Since we already explained the problematic we can focus more on the solution maybe?

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,72 @@
...
+
+PDBs and how to locate them
+---------------------------
+
+A PDB has the information to link the source code of a debuggable object to the Visual Studio debugger. Each PDB is linked to a
+specific file (executable or library) and contains the source file name and line numbers to display in the IDE.
+When compiling the files in Debug mode the created binary will contain the information of where the PDB will be
+generated, which by default is the same path where the file is being compiled. The PDBs are created by the ``cl.exe``
+compiler with the ``/Zi`` flag, or by the ``link.exe`` when linking a DLL or executable.
+
```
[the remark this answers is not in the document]
juansblanco: Done, added a console example with both `vswhere` to locate the `dumpbin.exe` and then `dumpbin` command with it's output.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -1,16 +1,18 @@
...
 .. _examples_dev_flow_debug_visual:
 
 
 Debugging shared libraries with Visual Studio
 =============================================
 
-Conan packages don't contain the necessary information for debugging libraries with Visual Studio by default.
-This information is stored in PDBs that are generated during the compilation of the libraries. When using Conan these
-PDBs are generated in the build folder, which isn't always available, making impossible to debug the dependencies.
+We previously discussed how to debug dependencies in Conan with Visual Studio, but that requires the original build
```
czoido, line 7: I think you should rephrase this, there's no such thing as `debug dependencies in Conan`

RubenRBS (approved):

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+After closing and reopening our solution in Visual Studio, we can try to debug again. If you try to step into the
+dependency, with the breakpoint on line 22, you will notice it will directly skip over to the next line as Visual Studio
+doesn't have any information on the dependencies to debug.
+
+
+Installing the new hook to copy PDBs
+------------------------------------
+
+The hook is available in the `conan-extensions repository <https://github.com/conan-io/conan-extensions>`_.
+Installing the whole repository will function, but we recommend to only install the hooks folder from the
```
czoido, line 74:
```suggestion
Installing the whole repository will work, but we recommend to only install the hooks folder from the
```

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+
+- The ``build()`` method of the recipe is executed, generating the DLLs and PDBs
+- The ``package()`` method of the recipe is executed, copying the necessary files to the package folder (in this case the DLLs but not the PDBs)
+- The hook is executed copying the PDBs from the build folder next to its DLL for every DLL in the package
+
+
+Debugging without build files
+-----------------------------
+
+After installing the hook we will create again the project from sources so the hook can now copy the PDBs to the package.
```
czoido, line 99: I think we should clearly state that they are copied besides the DLL's and then the DLL can find the PDB because it's in the same folder
```suggestion
After installing the hook we will create again the project from sources so the hook can now copy the PDBs to the package folder alongside the package DLL's.
```

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+    ...
+    [HOOK - hook_copy_pdbs_to_package.py] post_package(): PDBs post package hook running
+    ...
+    Install finished successfully
+
+    # CMake presets require CMake>=3.23
+    $ cmake --preset=conan-default
+
+Notice that when running the conan install now you will see the outputs of the hook running after the call to package().
+We will clean the cache once again to test the debugging without the build files:
```
czoido, line 115: We should remark what's happening here: that with conan cache clean we will delete the build folder and inside it are the sources that were used to build the library.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+
+.. code-block:: bash
+
+    $ conan cache clean zlib/1.2.11 --build
+
+Open the solution in Visual Studio again and start the debugger. When you try to step into the dependency in line 22, an error
+message will pop up telling us the file was not found and it will ask where the file is located. We can close this window
+and it will give the option to view the disassembly which can be debugged thanks to the PDB.
```
czoido, line 123: Add a sentence to explain that we can't step into the source files now, like we did before because we removed the source files from the cache.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+
+.. image:: ../../../images/examples/dev_flow/add_path_to_debug_source_files.png
+    :alt: Setting source path
+
+Starting the debugger again will allow to step into the code of the dependency as in the first example we did.
+
+.. note::
+
+    For Visual Studio to find the source files it is necessary that these are the same exact files as when the
+    libraries were compiled because Visual does a checksum check.
```
czoido, line 158: I think this should be the second paragraph and the next paragraph should be the first, because the fact that the patched sources are not the same than the ones from the source folder can make that visual studio does not recognize the sources as related with the PDB's because of the checksum check.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+    $ conan cache path --folder build zlib/1.2.11:17b26a16efb893750e4481f98a154db2934ead88
+    $ conan cache clean zlib/1.2.11 --build
+    $ conan cache path --folder build zlib/1.2.11:17b26a16efb893750e4481f98a154db2934ead88
+
+After closing and reopening our solution in Visual Studio, we can try to debug again. If you try to step into the
+dependency, with the breakpoint on line 22, you will notice it will directly skip over to the next line as Visual Studio
+doesn't have any information on the dependencies to debug.
+
+
+Installing the new hook to copy PDBs
```
czoido, line 70: I think we can improve the title for the section, like: "Installing a hook to copy the PDB's to the package folder" or something like that...

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,211 @@
...
+    libraries were compiled because Visual does a checksum check.
+
+    Also note that we are using the source files from the source folders, in the case that there exist patches
+    for those source files they won't be applied, as that's done in a later step in the build folder.
+
+
+.. _examples_dev_flow_debug_visual_pdb:
+
+PDBs and how to locate them
+---------------------------
```
czoido, line 167: this section is an explanation for the hook and feels a bit out of context here, I think we should move it after the hook is installed to explain how the hook works

juansblanco: Moved section and resumed it. The complete info will be in the readme of the hook.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,207 @@
...
+
+
+Debugging shared libraries with Visual Studio
+=============================================
+
+We previously discussed how to debug dependencies in Visual Studio, but when creating your project with Conan is
+possible that the original build folder and build files don't exist. Conan packages don't contain the necessary
+information for debugging libraries with Visual Studio by default, this information is stored in PDBs that are generated during the
+compilation of the libraries. When using Conan these PDBs are generated in the build folder, which is generated during
+the building of the libraries and it's no longer needed afterwards. For example, cleaning the Conan cache with
```
czoido, line 11:
```suggestion
compilation of the libraries. When using Conan these PDBs are generated in the build folder,  during the building of the libraries and are no longer needed afterwards. For that reason it's a common operation to clean the Conan cache with
```

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,207 @@
...
+``conan cache clean`` will remove the build folder.
```
czoido, line 12:
```suggestion
compilation of the libraries. When using Conan these PDBs are generated in the build folder,  during the building of the libraries and are no longer needed afterwards. For that reason it's a common operation to clean the Conan cache with
``conan cache clean`` to remove the build folder and save disk space.
```

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,207 @@
...
+
+The goal is to store the necessary PDB files in the package folder to make sure they are always present and don't depend on the
```
czoido, line 14: This introduction needs to be better linked with the previous paragraph.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,207 @@
...
+    $ conan list zlib/1.2.11:*
+    $ conan cache path --folder build zlib/1.2.11:17b26a16efb893750e4481f98a154db2934ead88
+    $ conan cache clean zlib/1.2.11 --build
+    $ conan cache path --folder build zlib/1.2.11:17b26a16efb893750e4481f98a154db2934ead88
+
+After closing and reopening our solution in Visual Studio, we can try to debug again. If you try to step into the
+dependency, with the breakpoint on line 22, you will notice it will directly skip over to the next line as Visual Studio
+doesn't have any information on the dependencies to debug.
+
+
```
czoido, line 66: Add a sentence to link better with the next section.

examples/dev_flow/debug/debugging_visual.rst
```diff
@@ -0,0 +1,182 @@
...
+    zlib/1.2.11: Calling package()
+    ...
+    [HOOK - hook_copy_pdbs_to_package.py] post_package(): PDBs post package hook running
+    ...
+    Install finished successfully
+
+    # CMake presets require CMake>=3.23
+    $ cmake --preset=conan-default
+
+Notice that when running the conan install now you will see the outputs of the hook running after the call to package().
```
czoido, line 129:
```suggestion
Notice that when running the conan install now you will see the outputs of the hook running after the call to ``package()``.
```

czoido (approved):

czoido (approved):
I changed the target branch to the release one so it's published immediately, good job!

PR #3695, opened 2024-04-26
Add local recipes index docs

czoido:
We can discuss if this is the best structure, I added one part in the devops guide and other next to the Conan remotes section in the tutorial but trying to clarify that is for another purpose...


memsharded (approved):
Looks great to me so far.

PR #3719, opened 2024-05-09
Documented cmaketoolchain:extra_variables (#16229)

perseoGI:
Added documentation for new `cmaketoolchain` option `extra_variables`
Refers to: https://github.com/conan-io/conan/pull/16242


reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -607,6 +607,16 @@ CMakeToolchain is affected by these ``[conf]`` variables:
 - **tools.cmake.cmaketoolchain:system_name** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_NAME``.
 - **tools.cmake.cmaketoolchain:system_version** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_VERSION``.
 - **tools.cmake.cmaketoolchain:system_processor** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_PROCESSOR``.
+- **tools.cmake.cmaketoolchain:extra_variables**: dict-like python object which specifies the CMake variable name and value. E.g.
+    
+.. code:: bash
+
+    conan install . -c "tools.cmake.cmaketoolchain:extra_variables*={'CMAKE_GENERATOR_INSTANCE': 'C:/Program Files (x86)/Microsoft Visual Studio/2017/BuildTools/'}"
+
+.. tip::
```
RubenRBS, line 616: TIL! Can I get a pic of how this looks when built?

RubenRBS (approved):

reference/tools/cmake/cmaketoolchain.rst
perseoGI: image
It looks good to me, but maybe it is not what we want.
As a noobie, I like command examples so that's why I added it. And also the *= looks very important because really nobody read the docs xD

perseoGI: Update: 
image

reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -625,6 +625,26 @@ CMakeToolchain is affected by these ``[conf]`` variables:
 - **tools.cmake.cmaketoolchain:system_name** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_NAME``.
 - **tools.cmake.cmaketoolchain:system_version** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_VERSION``.
 - **tools.cmake.cmaketoolchain:system_processor** is not necessary in most cases and is only used to force-define ``CMAKE_SYSTEM_PROCESSOR``.
+- **tools.cmake.cmaketoolchain:extra_variables**: dict-like python object which specifies the CMake variable name and value. Value can be a plain string, number or a dict-like python object which must specifies ``value`` (string/number) , ``cache`` (boolean), ``type`` (CMake cache type) and optionally, ``docstring`` (string: defaulted to variable name). Potential override of CMakeToolchain defined variables, users are at their own risk. E.g.
+    
+.. code-block:: text
+
+    [conf]
+    tools.cmake.cmaketoolchain:extra_variables*={'STR': 'string value', 'NUM': 1.3, 'CMAKE_GENERATOR_INSTANCE': '${ENV}/buildTools/', 'SOME_DICT': {'value': '42 sense', 'cache': True, 'type': 'BOOL', 'docstring': 'test cache var'}}
```
danimtb, line 633: I think it is easier to understand the usage with a simple example without the "update" configuration data operator
```suggestion
    tools.cmake.cmaketoolchain:extra_variables={'STR': 'string value', 'NUM': 1.3, 'CMAKE_GENERATOR_INSTANCE': '${ENV}/buildTools/', 'SOME_DICT': {'value': '42 sense', 'cache': True, 'type': 'BOOL', 'docstring': 'test cache var'}}
```

Then maybe, add a meaninful example with the update operator `*=` in the tip box below

reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -625,6 +625,26 @@ CMakeToolchain is affected by these ``[conf]`` variables:
...
+.. code-block:: cmake
+
+    set(STR "string value")
+    set(NUM 1.3)
+    set(CMAKE_GENERATOR_INSTANCE "${ENV}/buildTools/")
+    set(SOME_DICT "42 sense" CACHE BOOL "test cache var")
+
+.. tip::
+
+    Use the ``*=`` operator described in :ref:`configuration data operators <configuration_data_operators>` to update the dictionary merging it with previous definitions of ``extra_variables``.
```
danimtb, line 646: I am not sure if this tip is relevant here. Is there a good example of a use case of this usage for extra_variables?

Anyway, I think it reads better this way. WDYT?
```suggestion
    Use the :ref:`configuration data operator<configuration_data_operators>` ``*=`` to **update** (instead of redefine) conf variables already defined in profiles or global configuration.
```

memsharded (changes requested):
The dict is not very readable or didactical. I think an incremental approach, starting with a single ``{"MYCMAKEVAR": "Myvalue"}`` is better:
- Simple complete use case of the ``extra_variables``
- Use something better than ``STR`` because that seems a type, not a variable name.
- Single line that fits in the screen

Then explain the .cmake generated from it, and explain clearly that it is injected later so it can overwrite.

Then, finally, explain the advanced syntax, with the whole dict with the type, cache and docstring.

reference/tools/cmake/cmaketoolchain.rst
perseoGI: Okey thanks!

reference/tools/cmake/cmaketoolchain.rst
perseoGI: Final result
image

reference/tools/cmake/cmaketoolchain.rst
```diff
@@ -653,6 +653,42 @@ CMakeToolchain is affected by these ``[conf]`` variables:
...
+
+    set(MyIntegerVariable 42)
+    set(CMAKE_GENERATOR_INSTANCE "${ENV}/buildTools/")
+    set(CACHED_VAR "/var/run" CACHE BOOL "test cache var")
+
+This block injects ``$`` which will be expanded later. It also defines a cache variable of type ``PATH``.
+
+.. tip::
+
+    Use the :ref:`configuration data operator<configuration_data_operators>` ``*=`` to **update** (instead of redefine) conf variables already defined in profiles or global configuration.
```
RubenRBS, line 690:
```suggestion
    Use the :ref:`configuration data operator<configuration_data_operators>` ``*=`` to **update** (instead of redefining) conf variables already set in profiles or the global configuration.
```

czoido (approved):

memsharded (commented):
Much better, thanks!

PR #3734, opened 2024-05-17
Make search highlight less hideous

RubenRBS:
image


memsharded (approved):
Agree it is too strong at the moment

czoido (approved):

PR #3735, opened 2024-05-17
Fix global config zlib example pattern

stevn:
I am trying to use `Xcode` CMake generator as default, but `Unix Makefiles` CMake generator for zlib package.

The old example config (`zlib:tools.cmake...`) didn't work for me using Conan client 1.64.1 on MacOS. Conan kept using the Xcode generator for build.

I had to use the new example pattern using star sign for the version of the zlib package (`zlib/*:tools.cmake...`).

This the `global.conf` that works for me:

```
tools.cmake.cmaketoolchain:generator=Xcode
zlib/*:tools.cmake.cmaketoolchain:generator=Unix Makefiles
```

So I think the documentation should be updated for other users to not have to deal with this issue.

I am using Conan 1 and don't know how this behaves in Conan 2.

I guess it might also be a bug in the Conan client? Not sure...

### DIFF
--- a/reference/config_files/global_conf.rst
+++ b/reference/config_files/global_conf.rst
@@ -243,7 +243,7 @@ You can use package patterns to apply the configuration in those dependencies wh
 .. code-block:: text
 
     *:tools.cmake.cmaketoolchain:generator=Ninja
-    zlib:tools.cmake.cmaketoolchain:generator=Visual Studio 16 2019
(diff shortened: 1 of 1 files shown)


memsharded (approved):
Thanks for contributing this fix.

Yes, this seems an error in the docs, the ``conf`` per-package qualifiers must be patterns, so ``zlib/*`` is necessary, not just ``zlib``. I think it should be the same in Conan 2, but I will double check there.

I have moved the PR to master branch so it is published immediately after merge and not wait until next release.

PR #3757, opened 2024-06-04
languages and cstd support

memsharded:
Docs for https://github.com/conan-io/conan/pull/16028

### DIFF
--- a/reference/conanfile/attributes/binary_model.inc
+++ b/reference/conanfile/attributes/binary_model.inc
@@ -331,6 +331,33 @@ For example:
             "option2": "Describe the purpose and functionality of 'option2'. ",
         }
 
+languages
+---------
+
+.. include:: ../../common/experimental_warning.inc
+
+
+From Conan 2.4, the ``conanfile.py`` recipe attribute ``languages`` can be used to define the programming languages 
+involved in this package. At the moment the ``C`` and ``C++`` languages are the possible values. For example a
+pure C package would define something as:
+
+.. code-block:: python
+
+    class ZLib(ConanFile):
+        languages = "C"
+
+
+It is possible to define more than one language, for example ``languages = "C", "C++"`` is the correct definition when
+a package is built from both C and C++ sources.
+
+Regarding ``languages`` definition, the following will happen:
+
+- If no ``languages`` is defined or ``C`` is not a declared language, ``compiler.cstd`` subsetting will be automatically removed 
+  at package ``configure()`` time (to achieve backward compatibility).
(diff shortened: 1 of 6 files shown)


reference/conanfile/attributes/binary_model.inc
```diff
@@ -331,6 +331,33 @@ For example:
...
+        languages = "C"
+
+
+It is possible to define more than one language, for example ``languages = "C", "C++"`` is the correct definition when
+a package is built from both C and C++ sources.
+
+Regarding ``languages`` definition, the following will happen:
+
+- If no ``languages`` is defined or ``C`` is not a declared language, ``compiler.cstd`` subsetting will be automatically removed 
+  at package ``configure()`` time (to achieve backward compatibility).
```
RubenRBS, line 356:
```suggestion
  at package ``configure()`` time (to achieve backward compatibility).
  Thus Conan assumes that ``C++`` is the default value for ``languages`` if not overriden by the user.
```

reference/conanfile/methods/configure.rst
```diff
@@ -29,6 +29,12 @@ recipe to remove them so they are not used in the recipe:
         # No need to delete those settings here, they were already deleted
         pass
 
+.. note::
+
+    From Conan 2.4, the above ``configure()`` is not necessary if defined ``languages = "C"`` recipe
```
RubenRBS, line 34:
```suggestion
    From Conan 2.4, the above ``configure()`` is not necessary if defining ``languages = "C"`` as a recipe
```

reference/extensions/profile_plugin.rst
```diff
@@ -14,6 +14,8 @@ This ``profile.py`` contains a default implementation that does:
   That allow users to let it undefined in profiles, and switch it conveniently in command line just with ``-s build_type=Debug``
 - Will check the ``compiler.cppstd`` value if defined to validate if the current compiler version has support for it. For example, if a developer
   tries to use ``-s compiler=gcc -s compiler.version=5 -s compiler.cppstd=20``, it will raise an error.
+- Even the ``profile.py`` plugin has some provision to handle ``compiler.cstd`` checks they are not implemented yet, so the plugin will be permissive
```
RubenRBS, line 17:
```suggestion
- Even though the ``profile.py`` plugin has some provision to handle ``compiler.cstd`` checks, they are not implemented yet, so the plugin will be permissive
```

RubenRBS (approved):

reference/conanfile/attributes/binary_model.inc
memsharded: Not super accurate, it will not be reported ``languages = "C++"`` if it is not defined in any of the places it is reported (json output, ``dependencies`` interface, etc), it will be an empty list. So it is not the default value, the default value is an empty list.

PR #3758, opened 2024-06-04
Added MacOS sdk_version 14.5

perseoGI:


jcar87 (approved):

reference/config_files/settings.rst
```diff
@@ -61,7 +61,7 @@ It looks like this:
                        "13.0", "13.1", "13.2", "13.3", "13.4", "13.5", "13.6",
                        "14.0", "14.1", "14.2", "14.3", "14.4"]
```
czoido, line 62: I think os.version should also be updated, not only the sdk?

czoido (approved):

Issue #3629, opened 2024-02-16
[bug] path argument no longer position independent for "conan lock create" in Conan 2.1.0


antoinebardoz:
### Environment details

* Operating System+version: Ubuntu 22.04
* Compiler+version: N/A
* Conan version: 2.1.0
* Python version: 3.10.8

### Steps to reproduce

1. ```conan lock create -r myremote -u path/to/conanfile```
2. Conan returns: ```ERROR: Please specify a path to a conanfile or a '--requires=<ref>'```

This doesn't happen if I don't have ```-u``` in the command. It also doesn't happen if I put the path first:
```conan lock create path/to/conanfile -r myremote -u```

This may apply to all commands. I saw it specifically with ```conan lock create```, but after checking, I saw it happen with ```conan create``` as well.

PR #3760, opened 2024-06-05
Order of arguments good practise

AbrilRBS:
Closes https://github.com/conan-io/docs/issues/3629

### DIFF
--- a/knowledge/guidelines.rst
+++ b/knowledge/guidelines.rst
@@ -43,9 +43,10 @@ Good practices
   the graph to avoid the conflicts without overrides and forces is the recommended approach.
 - **Please, do not abuse 'tool_requires'**. Those are intended only for executables like ``cmake`` and ``ninja`` running in the "build"
   context, not for libraries or library-like dependencies, that must use ``requires`` or ``test_requires``.
-- Named arguments which receive a value should use ``=`` without spaces. For example, ``-r=conancenter`` is correct, but ``-r conancenter`` is not.
 - Positional arguments when invoking Conan should be specified first, before any named argument. For example,
   ``conan install . -s="os=Windows"`` is correct, but ``conan install -s="os=Windows" .`` is not.
+  Likewise, it's recommended to use ``=`` instead of spaces between the name and value of named arguments.
+  This is to avoid some ambiguity scenarios when parsing the command line arguments.
 
 Forbidden practices
 -------------------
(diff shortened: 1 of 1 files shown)


knowledge/guidelines.rst
```diff
@@ -43,6 +43,8 @@ Good practices
   the graph to avoid the conflicts without overrides and forces is the recommended approach.
 - **Please, do not abuse 'tool_requires'**. Those are intended only for executables like ``cmake`` and ``ninja`` running in the "build"
   context, not for libraries or library-like dependencies, that must use ``requires`` or ``test_requires``.
+- Positional arguments when calling Conan should be specified first, before any named argument. For example,
+  ``conan install . -s os=Windows`` is correct, but ``conan install -s os=Windows .`` is not.
```
RubenRBS, line 47:
```suggestion
  ``conan list "zlib/*" -r=conancenter`` is correct, but ``conan -r=conancenter "zlib/*"`` is not.
```

maybe this is a bit more clear?

memsharded: I like the ``conan install .`` example, why do you think the ``conan list`` is better?

memsharded (approved):

knowledge/guidelines.rst
RubenRBS: The positional argumwent being a point might get missed by readers, whereas the pattern in the list is a bit more prominent, but let's merge as-is if it looks good for you

RubenRBS, line 47:
```suggestion
  ``conan install . -s=os=Windows`` is correct, but ``conan install -s=os=Windows .`` is not.
```

knowledge/guidelines.rst
```diff
@@ -43,6 +43,9 @@ Good practices
   the graph to avoid the conflicts without overrides and forces is the recommended approach.
 - **Please, do not abuse 'tool_requires'**. Those are intended only for executables like ``cmake`` and ``ninja`` running in the "build"
   context, not for libraries or library-like dependencies, that must use ``requires`` or ``test_requires``.
+- Named arguments which receive a value should use ``=`` without spaces. For example, ``-r=conancenter`` is correct, but ``-r conancenter`` is not.
+- Positional arguments when invoking Conan should be specified first, before any named argument. For example,
+  ``conan install . -s="os=Windows"`` is correct, but ``conan install -s="os=Windows" .`` is not.
```
memsharded, line 48: Not fully sure. The ``-r conancenter`` does work, so maybe we want to reword a bit? I'd do:
- Use just 1 item to define both good practices for CLI args, not 2 bullet items
- Define the good practice without saying that the other is not correct (positive recommendation), just recommend one over the other (it can be said that there can be arg parsing errors if not in some cases)

memsharded (approved):

czoido (approved):

PR #3763, opened 2024-06-05
fix syntax of `conan config install --args` example

nmasseyKM:
Passing an argument to "--args" as a separate string generates:

    conan config install: error: argument -a/--args: expected one argument

if the argument looks like a separate option. Which pretty much anything you want to pass to "git clone" will.


memsharded:
I have moved the PR to ``release/2.4`` branch, so it is published immediately without having to wait until next release

PR #3495, opened 2023-12-19
Docs for replace_requires and replace_tool_requires profile sections

danimtb:
Docs for https://github.com/conan-io/conan/pull/15136/files

PR #3765, opened 2024-06-06
Fix replace requires for zlib in profile

uilianries:
The actual [replace-requires](https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires) usage in the profile shows `zlib/1.2.123: zlib/*`, which does not work. My guess is "replace zlib/1.2.123 by latest version available", but then we need to use `[*]` instead.

Related to the PR #3495

/cc @RubenRBS

### DIFF
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -36,7 +36,7 @@ They have this structure:
     tools.build:jobs=2
 
     [replace_requires]
-    zlib/1.2.123: zlib/[*]
+    zlib/1.2.12: zlib/[*]
 
     [replace_tool_requires]
     7zip/*: 7zip/system
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -36,7 +36,7 @@ They have this structure:
     tools.build:jobs=2
 
     [replace_requires]
-    zlib/1.2.123: zlib/*
+    zlib/1.2.123: zlib/[*]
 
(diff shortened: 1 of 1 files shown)


reference/config_files/profiles.rst
```diff
@@ -36,7 +36,7 @@ They have this structure:
     tools.build:jobs=2
 
     [replace_requires]
-    zlib/1.2.123: zlib/*
+    zlib/1.2.123: zlib/[*]
```
memsharded, line 39: The previous syntax is supported, and it means to replace the existing version with the same version, as a place holder, but needs to be something like ``zlib/*: zlib-ng/*``, which is different than replacing something by a version range.

memsharded (approved):

reference/config_files/profiles.rst
RubenRBS:
```suggestion
    zlib/1.2.12: zlib/[*]
```

RubenRBS (approved):

reference/config_files/profiles.rst
uilianries: > The previous syntax is supported, and it means to replace the existing version with the same version, as a place holder, but needs to be something like zlib/*: zlib-ng/*

Yes, but `zlib/*: zlib-ng/*` is documented already in https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires. Nice to know, thank you.

PR #3779, opened 2024-06-20
Colour code like Conan 1 docs, ensure highlight works

RubenRBS:
After talking with @memsharded, this colour scheme looks nicer, it's more prominent to the user that they are reading code

image

The only thing I'm not completely sold on is the highlight colour. I'm open to suggestions! :)

Also, this will only work for the new 2.5 release. Should we backport this to at least 2.4?

### DIFF
--- a/_static/css/conan.css
+++ b/_static/css/conan.css
@@ -25,3 +25,7 @@ h1, h2, .rst-content .toctree-wrapper p.caption, h3, h4, h5, h6, legend {
     text-align: left;
     font-weight: bold;
 }
+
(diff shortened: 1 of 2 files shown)


memsharded (approved):
Looks good, so finally no light orange, but the most standard light green? Did this already changed in Conan 1 docs? It seems so.

czoido:
What about using `pygments_style = 'default'` instead of `'sphinx'` and not modifying the highlight color?

image

czoido:
this should go to master to regenerate all the branches

PR #3799, opened 2024-07-15
Feature/devops ci

memsharded:
Very initial draft of the first sections, to get team early feedback.

It is better reviewed generating the html docs, not in code.

Depends on https://github.com/conan-io/examples2/pull/153

Close https://github.com/conan-io/docs/issues/3849
Close https://github.com/conan-io/conan/issues/3003
Close https://github.com/conan-io/conan/issues/6589
Close https://github.com/conan-io/conan/issues/6045
Close https://github.com/conan-io/conan/issues/8587
Close https://github.com/conan-io/conan/issues/16309


devops.rst
```diff
@@ -14,6 +14,7 @@ If you plan to use Conan in production in your project, team, or organization, t
 
    devops/using_conancenter
    devops/devops_local_recipes_index
+   devops/continuous_integration/tutorial
```
czoido, line 17: Maybe we can use the `guide` instead `tutorial` to avoid confusion with the `tutorial` section?

czoido: or moving it to the tutorial would be another option?

devops/continuous_integration/tutorial.rst
```diff
@@ -0,0 +1,141 @@
+Continuous Integration (CI) tutorial
+====================================
+
+Continuous Integration has different meanings for different users and organizations. In this tutorial we will cover the scenarios when users
+are doing changes to the source code of their packages and want to automatically build new binaries for those packages and also compute if those new package changes integrate cleanly or break the organization main products.
+
+We will use in this tutorial this small project that uses several packages (static libraries by default) to build a couple of applications, a video game and a map viewer utility:
```
czoido, line 7: Maybe we can advance here that we will refer to those applications as _products_ later

devops/continuous_integration/tutorial.rst
```diff
@@ -0,0 +1,141 @@
...
+  above, when some developer did source changes in a new ``ai/1.1.0`` recipe, the different binaries for Windows and Linux will be built in different
+  servers. These jobs can upload their respective binaries for Windows and Linux to the ``packages`` binary repository. Note that these individual
+  binaries will not disrupt other developers or CI jobs, as they don't use the ``packages`` repository.
+- ``products``: It is possible that some changes create new package versions or revisions correctly. But these new versions might break consumers
+  of those packages, for example some changes in the new ``ai/1.1.0`` package might unexpectedly break ``engine/1.0``. Or even if they don't
+  necessarily break, they might still need to build a new binary from source for ``engine/1.0`` and/or ``game/1.0``. The ``products`` binary
+  repository will be the place where binaries for different packages are uploaded to not disrupt or break the ``develop`` repository, until
+  the "products pipeline" can build necessary binaries from source and verify that these packages integrate cleanly.
```
czoido, line 89: I have the impression that this would be better if summarized in bullet points or something like that, maybe just giving the most important attributes of each server and later readers will have more data when they read the full story? Or at least summarize a bit the ``packages`` that I think ends up being hard to follow.

czoido: I think that all this part maybe is hard to follow because is trying to advance too much about the story, maybe we should limit here to explain briefly what kind of binaries is storing each repo and then making sense of it later?

devops/continuous_integration/tutorial.rst
```diff
@@ -0,0 +1,141 @@
...
+
+
+.. graphviz::
```
czoido, line 92: What about adding the pipelines to the diagram?

devops/continuous_integration/project_setup.rst
```diff
@@ -0,0 +1,91 @@
...
+.. code-block:: bash
+
+    $ git clone https://github.com/conan-io/examples2.git
+    $ cd examples2/devops/ci/game
+
+
+Server repositories setup
+-------------------------
+
+We need 3 different repositories in the same server. Make sure to have an Artifactory running and available. You can download the free  `JFrog Artifactory Community Edition (CE) <https://conan.io/downloads.html>`_ and run it in your own computer. If you have another available Artifactory, it can be used too if you can create new repositories there. Log into the web UI and create 3 different local repositories called ``develop``, ``packages`` and ``products``.
```
czoido, line 17: maybe it's good to say that we have to create three _local_ repositories

devops/continuous_integration/packages_pipeline/single_configuration.rst
```diff
@@ -0,0 +1,98 @@
...
+As we described before while presenting the different server binary repositories, the idea is that package builds
+will use by default the ``develop`` repo only, which is considered the stable one for developer and CI jobs.
+
+Let's make sure we start from a clean state:
+
+.. code-block:: bash
+
+    $ conan remove "*" -c  # Make sure no packages from last run
+    $ conan remote remove "*"  # Make sure no other remotes defined
+    $ conan remote add develop <url-develop-repo>  # Add only the develop repo
```
czoido, line 16: I think that most users will run all these commands for a local artifactory, so it would be great to offer them the full line for copying and just add a comment to change this if if differs

```suggestion
    # adjust these lines for your artifactory instance
    $ conan remote add develop http://localhost:8081/artifactory/api/conan/develop
    $ conan remote login develop admin -p password
```

devops/continuous_integration/tutorial.rst
memsharded: Doing "clusters" or subgraphs is quite challenging, adding the pipelines should be 2 overlapping subgraphs, no idea if this is possible.

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+    $ conan remote add develop <url-develop-repo>  # Add only the develop repo
+
+
+We will create the packages for the 2 configurations sequentially in our computer, but note these will typically run
+in different computers, so it is typical for CI systems to launch the builds of different configurations in parallel.
+
+.. code-block:: bash
+    :caption: Release build
+
+    $ conan create . --build="missing:ai/*" -s build_type=Release --format=json > graph.json
```
czoido, line 25:
```suggestion
    $ cd ai
    $ conan create . --build="missing:ai/*" -s build_type=Release --format=json > graph.json
```

devops.rst
AbrilRBS: I agree that guide might be a tad less confusing. Also, I'd set this as the first item in the devops tab

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+    $ conan list --graph=graph.json --graph-binaries=build --format=json > upload_release.json
+    $ conan remote add packages "<url-packages-repo>"
+    $ conan upload -l=upload_release.json -r=packages -c --format=json > upload_release.json
```
czoido, line 28:
```suggestion
    $ conan list --graph=graph.json --graph-binaries=build --format=json > upload.json
    $ conan remote add packages "<url-packages-repo>"
    $ conan upload -l=upload.json -r=packages -c --format=json > upload_release.json
```
Using the same name makes it fail because the file is deleted for output before it's read by Conan

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+Likewise, the Debug build will do the same steps:
+
+
+.. code-block:: bash
+    :caption: Debug build
+
+    $ conan create . --build="missing:ai/*" -s build_type=Debug --format=json > graph.json
+    $ conan list --graph=graph.json --graph-binaries=build --format=json > upload_debug.json
+    $ conan remote add packages "<url-packages-repo>" -f  # Can be ommitted, it was defined above
+    $ conan upload -l=upload_debug.json -r=packages -c --format=json > upload_debug.json
```
czoido, line 50:
```suggestion
    $ conan create . --build="missing:ai/*" -s build_type=Debug --format=json > graph.json
    $ conan list --graph=graph.json --graph-binaries=build --format=json > upload.json
    $ conan remote add packages "<url-packages-repo>" -f  # Can be ommitted, it was defined above
    $ conan upload -l=upload.json -r=packages -c --format=json > upload_debug.json
```

devops/continuous_integration/project_setup.rst
czoido: > > nd create 3 different local repositories called
> 
> I think that's already the case no?

🤦

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+.. code-block:: bash
+    :caption: Promoting from packages->product
+
+    # aggregate the package list
+    $ conan pkglist merge -l upload_release.json -l upload_debug.json --format=json > promote.json
+
+    $ conan remote add packages "<url-packages-repo>" -f  # Can be ommitted, it was defined above
+    $ conan remote add products "<url-products-repo>" -f  # Can be ommitted, it was defined above
+
+    # Promotion with Artifactory CE (slow, can be improved with art:promote)
```
czoido, line 116:
```suggestion
    # Promotion using Conan download/upload commands (slow, can be improved with art:promote custom command)
```

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+
+If the build of all configurations for ``ai/1.1.0`` were succesfull, then the ``packages pipeline`` can proceed and promote
+them to the ``products`` repository:
+
+.. code-block:: bash
+    :caption: Promoting from packages->product
+
+    # aggregate the package list
+    $ conan pkglist merge -l upload_release.json -l upload_debug.json --format=json > promote.json
+
```
czoido, line 112:
```suggestion
    $ conan pkglist merge -l upload_release.json -l upload_debug.json --format=json > merged.json

```

devops/continuous_integration/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+    $ conan remote add packages "<url-packages-repo>" -f  # Can be ommitted, it was defined above
+    $ conan remote add products "<url-products-repo>" -f  # Can be ommitted, it was defined above
+
+    # Promotion with Artifactory CE (slow, can be improved with art:promote)
+    $ conan download --list=promote.json -r=packages --format=json > promote.json
```
czoido, line 117:
```suggestion
    $ conan download --list=merged.json -r=packages --format=json > promote.json
```
same problem with the file being deleted before read

memsharded: Oh, maybe it works differently with our ``run(..., redirect_output=filename)`` implementation. 
Understood, I need to change this. I will change it first in the ``examples2`` repo, then update it here.

devops/continuous_integration/project_setup.rst
```diff
@@ -0,0 +1,91 @@
...
+
```
czoido, line 18: I think it would be nice to add here the lines to launch Artifactory CPP from Docker (in a version that works), so that users that are a bit experienced with using it do not have to move to other place to get the line.

ci_tutorial/packages_pipeline.rst
```diff
@@ -0,0 +1,28 @@
...
+  so let`s edit the ``ai/conanfile.py`` file and define ``version = "1.1.0"`` there (instead of the previous ``1.0``). Note that if we
+  did some breaking changes to the ``ai`` public API, the recommendation would be to change the major instead and create a new ``2.0`` version.
+
+
+The ``packages pipeline`` will take care of building the different packages binaries for the new ``ai/1.1.0`` and upload them to the ``packages``
+binary repository. If the pipeline succeed it will copy them to the ``products`` binary repository, and stop otherwise.
+
+There are different aspects that need to be taken into account when building these packages. The following tutorial sections do the same
+job, but under different hypothesis. They are explained in increasing complexity.
+
```
AbrilRBS, line 21: I would note somewhere (not necessarily here) that the examples2 repo in fact has all the commands being ran, not just only the packages, and let users know that they are free to take those as a starting reference when implementing their own CI

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+In this example we are using a slow ``conan download`` + ``conan upload`` promotion. This can be way more efficient with 
+the ``conan art:promote`` extension command.
+
+After running the promotion we will have the following packages in the server:
+
+.. graphviz::
+    :align: center
+
+    digraph repositories {
+        node [fillcolor="lightskyblue", style=filled, shape=box]
```
AbrilRBS, line 134: We can tweak the colors at a later point until I can see what is going on :P

ci_tutorial/packages_pipeline/single_configuration.rst
```diff
@@ -0,0 +1,98 @@
...
+    ai/1.1.0: SUPER BETTER Artificial Intelligence for aliens (Release)!
+    ai/1.1.0: Intelligence level=50
+
+
+Note the ``--build="missing:ai/*"`` might not be fully necessary in some cases, but it can save time in other situations.
+For example, if the developer did some changes just to the repo README, and didn't bump the version at all, Conan will not 
+generate a new ``recipe revision``, and detect this as a no-op, avoiding having to unnecessarily rebuild binaries from source.
+
+If we are in a single-configuration scenario and it built correctly, for this simple case we won't need a promotion at all,
+and just uploading directly the built packages to the ``products`` repository will be enough:
```
AbrilRBS, line 38: I think I have a bit missing as to re-clarify how this package would then be used. As it now lives in the products remote, it won't be picked up by any job, so users that have a simpler need might not feel stuck as to what to do

ci_tutorial/tutorial.rst
```diff
@@ -0,0 +1,141 @@
...
+but this concept will be further explained later, and specially why it is important to think in terms of "products" instead of trying to explicitly
+model the dependencies top-bottom in the CI.
+
+
+Repositories and promotions
+---------------------------
+
+The concept of multiple server side repositories is very important for CI. In this tutorial we will use 3 repositories:
+
+- ``develop``: This repository is the main one that developers have configured in their machines to be able to ``conan install`` dependencies
```
AbrilRBS, line 73: I don't know if "develop" is a great name for this. Users might not necessarily associate develop with something that is supposed to be stable and the front facing remote for their daily usage

ci_tutorial/tutorial.rst
```diff
@@ -0,0 +1,141 @@
...
+  not its reference. Using user/channel to denote stages or maturity is discouraged.
+- Then, when packages are in the ``products`` repository, the ``products pipeline`` can be triggered. This job will make sure that both the
+  organization products ``game/1.0`` and ``mapviewer/1.0`` build cleanly with the new ``ai/1.1.0`` package, and build necessary new package
+  binaries, for example if ``engine/1.0`` needs to do a build from source to integrate the changes in ``ai/1.1.0`` the ``products pipeline``
+  will make sure that this happens.
+- When the ``products pipeline`` build all necessary new binaries for all intermediate and product packages and check that every is correct, then
+  these new packages can be made available for all other developers and CI jobs. This can be done with a promotion of these packages, copying
+  them from the ``products`` repository to the ``develop`` repository. As the changes have been integrated and tested consistently for the main
+  organization products, developers doing ``conan install`` will start seeing and using the new packages and binaries.
+
```
AbrilRBS, line 132: I think an advanced version of this could mention where users would expect to pull their final products from to release it to their users, it's somewhat counterintuitive that they should do it from the develop remote

AbrilRBS (commented):
Some thoughts

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,178 @@
...
+.. code-block:: bash
+    :caption: Promoting from packages->product
+
+    # aggregate the package list
+    $ conan pkglist merge -l upload_release.json -l upload_debug.json --format=json > promote.json
+
+    $ conan remote add packages "<url-packages-repo>" -f  # Can be ommitted, it was defined above
+    $ conan remote add products "<url-products-repo>" -f  # Can be ommitted, it was defined above
+
+    # Promotion with Artifactory CE (slow, can be improved with art:promote)
```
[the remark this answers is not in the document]
memsharded: Sounds good, I will add a "promotions" page in Devops section.

ci_tutorial/packages_pipeline.rst
memsharded: What do you mean with ``has all the commands being ran, not just only the packages,``?

devops/continuous_integration/tutorial.rst
memsharded: Done

ci_tutorial/packages_pipeline.rst
```diff
@@ -0,0 +1,41 @@
...
+✍️ **Let's do the following changes in the ai package**:
+
+- Let's change the implementation of the ``ai/src/ai.cpp`` function and change the message from ``Some Artificial`` to ``SUPER BETTER Artificial``
+- Let's change the default ``intelligence=0`` value in ``ai/include/ai.h`` to a new ``intelligence=50`` default.
+- Finally, let's bump the version. As we did some changes to the package public headers, it would be adviced to bump the ``minor`` version,
+  so let`s edit the ``ai/conanfile.py`` file and define ``version = "1.1.0"`` there (instead of the previous ``1.0``). Note that if we
+  did some breaking changes to the ``ai`` public API, the recommendation would be to change the major instead and create a new ``2.0`` version.
+
+
+The ``packages pipeline`` will take care of building the different packages binaries for the new ``ai/1.1.0`` and upload them to the ``packages``
```
czoido, line 25:
```suggestion
The **packages pipeline** will take care of building the different packages binaries for the new ``ai/1.1.0`` and upload them to the ``packages``
```

ci_tutorial/packages_pipeline/single_configuration.rst
```diff
@@ -0,0 +1,102 @@
+Package pipeline: single configuration
+======================================
+
+We will start with the most simple case, in which we only had to build 1 configuration, and that configuration 
+can be built in the current CI machine.
+
+As we described before while presenting the different server binary repositories, the idea is that package builds
+will use by default the ``develop`` repo only, which is considered the stable one for developer and CI jobs.
+
+Let's make sure we start from a clean state:
```
czoido, line 10: Maybe at this point it would be good to point to the configuration instructions or remind our users that they should have cloned the examples2 repo that has the .conanrc, to prevent someone that starts here and deletes the entire cache and remotes?

ci_tutorial/packages_pipeline/single_configuration.rst
```diff
@@ -0,0 +1,102 @@
...
+The removal and addition of repos accross this tutorial can be a bit tedious, but it is important for the correct
+behavior. Also, there might be other configurations that can be even more efficient for some cases, like re-triggering
+a broken job because of CI malfunction, but we will keep it simple at the moment and try to focus on the main concepts.
+
+With this configuration the CI job could just do:
+
+.. code-block:: bash
+
+    $ cd ai
+    $ conan create . --build="missing:ai/*"
```
czoido, line 29: maybe this saves one step?
```suggestion
    $ conan create ai --build="missing:ai/*"
```

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,195 @@
...
+the ``package pipeline`` can promote them to the ``products`` binary repository. As all other developers and CI use the ``develop`` repository,
+no one will be broken at this stage either:
+
+.. code-block:: bash
+    :caption: Promoting from packages->product
+
+    # aggregate the package list
+    $ conan pkglist merge -l uploaded_release.json -l uploaded_debug.json --format=json > uploaded.json
+
+    # Promotion using Conan download/upload commands
```
czoido, line 119: The step for adding the products repo is missing
```suggestion
    $ conan pkglist merge -l uploaded_release.json -l uploaded_debug.json --format=json > uploaded.json
    $ conan remote add products http://localhost:8081/artifactory/api/conan/products
    # Promotion using Conan download/upload commands 
```

ci_tutorial/packages_pipeline/multi_configuration_lockfile.rst
```diff
@@ -0,0 +1,147 @@
+Package pipeline: multi configuration using lockfiles
+=====================================================
+
+In the previous example, we built both ``Debug`` and ``Release`` package binaries for ``ai/1.1.0``. In real world scenarios the binaries to build would be different platforms (Windows, Linux, embedded), different architectures, and very often it will not be possible to build them in the same machine, requiring different computers.
+
+The previous example had an important assumption: the dependencies of ``ai/1.1.0`` do not change at all during the building process. In many scenarios, this assumption will not hold, for example if there are any other concurrent CI jobs, and one succesfull job publish a new ``mathlib/1.1`` version in the ``develop`` repo.
```
czoido, line 6:
```suggestion
The previous example had an important assumption: the dependencies of ``ai/1.1.0`` do not change at all during the building process. In many scenarios, this assumption will not hold, for example if there are any other concurrent CI jobs, and one succesfull job publishes a new ``mathlib/1.1`` version in the ``develop`` repo. 
```

ci_tutorial/packages_pipeline/multi_configuration_lockfile.rst
```diff
@@ -0,0 +1,147 @@
...
+        ],
+        "build_requires": [],
+        "python_requires": [],
+        "config_requires": []
+    }
+    
+As we can see, it is locking the ``mathlib/1.0`` dependency version and revision.
+
+
+With the lockfile, the creating of the different configurations is exactly the same, but providing the ``--lockfile=conan.lock`` argument to the ``conan create`` step, it will guarantee that ``mathlib/1.0#f2b05681ed843bf50d8b7b7bdb5163ea`` will always be the exact dependency used, irrespective if there exist new ``mathlib/1.1`` versions or new revisions available. The following builds could be launched in parallel but executed at different times, and still they will always use the same ``mathlib/1.0`` dependency:
```
czoido, line 60:
```suggestion
With the lockfile, creating of the different configurations is exactly the same, but providing the ``--lockfile=conan.lock`` argument to the ``conan create`` step, it will guarantee that ``mathlib/1.0#f2b05681ed843bf50d8b7b7bdb5163ea`` will always be the exact dependency used, irrespective if there exist new ``mathlib/1.1`` versions or new revisions available. The following builds could be launched in parallel but executed at different times, and still they will always use the same ``mathlib/1.0`` dependency:
```

ci_tutorial/packages_pipeline/multi_configuration_lockfile.rst
```diff
@@ -0,0 +1,147 @@
...
+
+
+.. code-block:: bash
+    :caption: Release build
+
+    $ cd ai
```
czoido, line 66: we already were executing things inside ai, so far at this point of the tutorial I find it more clear to launch everything from the root doing `conan create ai...` because with `conan create . `can be a bit confusing and you lose track of where you are

ci_tutorial/packages_pipeline/multi_configuration_lockfile.rst
```diff
@@ -0,0 +1,147 @@
...
+
+Creating the lockfile
+---------------------
+
+Let's make sure as usual that we start from a clean state:
+
+.. code-block:: bash
+
+    $ conan remove "*" -c  # Make sure no packages from last run
+    $ conan remote remove "*"  # Make sure no other remotes defined
```
czoido, line 22: I think you also need to clean the packages repo, otherwise at some point, before the second upload it will download the Debug package and create an empty package list (step in line 77)
```suggestion
    $ conan remote remove "*"  # Make sure no other remotes defined
    $ conan remove "*" -c -r packages
```

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,195 @@
...
+
+When all the different binaries for ``ai/1.1.0`` have been built correctly, the ``package pipeline`` can consider its job succesfull and decide
+to promote those binaries. But further package builds and checks are necessary, so instead of promoting them to the ``develop`` repository,
+the ``package pipeline`` can promote them to the ``products`` binary repository. As all other developers and CI use the ``develop`` repository,
+no one will be broken at this stage either:
+
+.. code-block:: bash
+    :caption: Promoting from packages->product
+
+    # aggregate the package list
```
czoido, line 116: the products remote is missing
```suggestion
    $ conan remote add products http://localhost:8081/artifactory/api/conan/products
    # aggregate the package list
```

ci_tutorial/products_pipeline/multi_product.rst
```diff
@@ -0,0 +1,131 @@
...
+.. code-block:: bash
+
+    $ conan graph build-order --requires=game/1.0 --build=missing 
+      --order-by=recipe --format=json > game_release.json
+    $ conan graph build-order --requires=game/1.0 --build=missing 
+      --order-by=recipe -s build_type=Debug --format=json > game_debug.json
+
+These commands are basically the same than in the previous section, each one with a different configuration and creating a different output file ``game_release.json`` and ``game_debug.json``. These files will be similar to the previous ones, but as we haven't used the ``--reduce`` argument (this is important!) they will actually contain a "build-order" of all elements in the graph, even if only some contain the ``binary: Build`` definition, and others will contain other ``binary: Download|Cache|etc``.
+
+.. code-block:: bash
```
czoido, line 42: A sentence here would be nice?
```suggestion
Now, let's compute the build-order for ``mapviewer/1.0``:

.. code-block:: bash
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+Using the information in the ``graph_build_order.json`` file, it is possible to execute the build of the necessary packages, in the same way that the previous section ``--build=missing`` did, but not directly managed by us.
+
+Taking the arguments from the json, the commands to execute would be:
+
+.. code-block:: bash
+
+    $ conan install --requires=engine/1.0 --build=engine/1.0
+    $ conan install --requires=game/1.0 --build=game/1.0
+
+We are executing these commands manually, but in practice, it would be a ``for`` loop in CI executing over the json output. We will see some Python code later for this. At this point we wanted to focus on the ``conan graph build-order`` command, but we haven't really explained how the build is distributed.
```
Minimonium, line 109: Maybe some re-ordering of these parts is in order, since we name it `distributed build` but do not in fact explain it

devops/versioning/default.rst
```diff
@@ -0,0 +1,70 @@
...
+the consumers binaries. See :ref:`how to customize the dependencies package_id modes<reference_binary_model_custom_compatibility_dependencies>`.
+
+
+.. note::
+
+    **Best practices**
+
+    - It is not recommended to use other package reference fields, as the ``user`` and ``channel`` to represent changes in the source code,
+      or other information like the git branch, as this becomes "viral" requiring changes in the ``requires`` of the consumers. Furthermore,
+      they don't implement any logic in the build model with respect to which consumers need to be rebuilt.
```
Minimonium, line 68: A good idea would be to provide some insight for where `user` and `channel` _can_ be used. After all these years, I still believe override is a useful dev tool which matches nicely with custom user/channel which helps making dev packages public without exposing dev remote.

ci_tutorial/tutorial.rst
```diff
@@ -0,0 +1,127 @@
...
+            "packages\n repository" -> "products\n repository" -> "develop\n repository" [ label="promotion" ];
+        }
+       
+    }
+
+Promotions are the mechanism used to make available packages from one pipeline to the other. Connecting the above packages and product pipelines
+with the repositories, there will be 2 promotions:
+
+- When all the different binaries for the different configurations have been built for a single package with the ``packages pipeline``, and uploaded
+  to the ``packages`` repository, the package changes and package new version can be considered "correct" and promoted (copied) to the ``products``
```
AbrilRBS, line 103:
```suggestion
  to the ``packages`` repository, the new version and changes to the package can be considered "correct" and promoted (copied) to the ``products``
```

ci_tutorial/tutorial.rst
```diff
@@ -0,0 +1,127 @@
...
+  the packages can be promoted (copied) from the ``products`` repo to the ``develop`` repo, to make them available for all other developers and CI.
+
+
+.. note::
+
+  - The concept of **immutability** is important in package management and devops. Modifying ``channel`` is strongly discouraged, see :ref:`Package promotions<devops_package_promotions>`.
+  - The versioning approach is important. This tutorial will be following :ref:`the default Conan versioning approach, see details here<devops_versioning_default>`
+
+This tutorial is just modeling the **development** flow. In production systems, there will be other repositories
+and promotions, like a ``testing`` repository for the QA team, and a final ``release`` repository for final users and packages can
```
AbrilRBS, line 116:
```suggestion
and promotions, like a ``testing`` repository for the QA team, and a final ``release`` repository for final users, such that packages can
```

devops/devops.rst
```diff
@@ -0,0 +1,24 @@
+.. _devops:
+
+
+Devops guide
+============
+
+The previous :ref:`tutorial<tutorial>` section was aimed at users in general and developers.
+
+The :ref:`Continuous Integration tutorial<ci_tutorial>` explained the basics how to implement Continuous Integration involving Conan packages.
```
AbrilRBS, line 9:
```suggestion
The :ref:`Continuous Integration tutorial<ci_tutorial>` explained the basics on how to implement Continuous Integration involving Conan packages.
```

ci_tutorial/tutorial.rst
```diff
@@ -0,0 +1,127 @@
...
+
+.. note::
+
+    **Important notes**
+
+    - This section is written as a hands-on tutorial. It is intended to be reproduced by copying the commands in your machine.
+    - The tutorial presents some of the tools, good practices and common approaches to the CI problem. But there are no silver bullets.
+      This tutorial is not the unique way that things should be done.
+      Different organizations might have different needs and priorities, different build services power and budget, different sizes, etc.
+      The principles and practices presented in the tutorial might need to be adapted.
```
AbrilRBS, line 48: Maybe here in the same line of "your millage may vary", it's a good place to remind readers that they can contact us by opening an issue in gh if they have question on the approach for their use-case?

devops/versioning/default.rst
```diff
@@ -0,0 +1,70 @@
...
+  to the ``conanfile.py`` recipe are done. As the source code is the same, we might want to keep the same version number, and
+  just have a new revision of that version.
+- **Patch**: Increasing the **patch** version of a package means that only internal changes were done, in practice it means change to files
+  that are not public headers of the package. This "patch" version can avoid having to re-build consumers of this package, for
+  example if the current package getting a new "patch" version is a static library, all other packages that implement static
+  libraries that depend on this one do not need to be re-built from source, as depending on the same public interface headers
+  guarantee the same binary.
+- **Minor**: If changes are done to package public headers, in an API source compatible way, then the recommendation would be to increase
+  the **minor** verson of a package. That means that other packages that depend on it will be able to compile without issues, 
+  but as there were modification in public headers (that could contain C++ templates or other things that could be inlined in
```
AbrilRBS, line 43:
```suggestion
  but as there were modifications in public headers (that could contain C++ templates or other things that could be inlined in
```

devops/package_promotions.rst
```diff
@@ -0,0 +1,162 @@
...
+            {
+                edge[style=invis];
+                "zlib/1.3.1" -> "release" ; 
+                rankdir="BT";    
+            }
+        }
+    }
+
+
+When QA team tests and approves these packages, they can be promoted to the "release" repository.
```
AbrilRBS, line 77:
```suggestion
When the QA team tests and approves these packages, they can be promoted to the "release" repository.
```

devops/package_promotions.rst
```diff
@@ -0,0 +1,162 @@
...
+If we have a package list ``pkglist.json`` that contains the above ``zlib/1.3.1`` and ``openssl/3.2.2`` packages, then
+the command would look like:
+
+.. code-block:: bash
+    :caption: Promoting from testing->release
+
+    $ conan art:promote pkglist.json --from=testing --to=release --url=https://<url>/artifactory --user=<user> --password=<password>
+
+
+Note that the ``conan art:promote`` doesn't work with ArtifactoryCE, but need pro editions of Artifactory.
```
AbrilRBS, line 96:
```suggestion
Note that the ``conan art:promote`` command doesn't work with ArtifactoryCE, Pro editions of Artifactory are needed.
```

ci_tutorial/products_pipeline/single_configuration.rst
```diff
@@ -0,0 +1,151 @@
...
+  ai/1.1.0: SUPER BETTER Artificial Intelligence for aliens (Release)!
+  ai/1.1.0: Intelligence level=50
+  graphics/1.0: Checking if things collide (Release)!
+  engine/1.0: Computing some game things (Release)!
+  game/1.0:fun game (Release)!
+
+We can see that the new ``game/1.0`` binary incorporates the improvements in ``ai/1.1.0``, and links correctly with the new binary for ``engine/1.0``.
+
+And this is a basic "products pipeline", we manage to build and test our main products when necessary (recall that ``mapviewer`` wasn't really affected, so no rebuilds were necessary at all).
+In general, a production "products pipeline" will finish uploading the built packages to the repository and running a new promotion to the ``develop`` repo. But as this was a very basic and simplify pipeline, let's wait a bit for that, and let's continue with more advanced scenarios.
```
AbrilRBS, line 151:
```suggestion
In general, a production "products pipeline" will finish uploading the built packages to the repository and running a new promotion to the ``develop`` repo. But as this was a very basic and simple pipeline, let's wait a bit for that, and let's continue with more advanced scenarios.
```

ci_tutorial/products_pipeline/multi_product.rst
```diff
@@ -0,0 +1,131 @@
+Products pipeline: multi-product multi-configuration builds
+===========================================================
+
+In the previous section we computed a ``conan graph build-order`` with several simplifications, we didn't take the ``mapviewer`` product into account, and we processed only 1 configuration.
+
+In real scenarios, it will be necessary to manage more than one product and the most common case is that there is more than on configuration for every product. If we build these differents cases sequentially it will be much slower and inefficient, and if we try to build them in parallel there will easily be many duplicated and unnecessary builds of the same packages, wasting resources and even producing issues as race conditions or traceability problems.
+
+To avoid this issue, it is possible to compute a single unified "build-order" that aggregates all the different build-orders that are compute for the different products and configurations.
```
AbrilRBS, line 8:
```suggestion
To avoid this issue, it is possible to compute a single unified "build-order" that aggregates all the different build-orders that are computed for the different products and configurations.
```

ci_tutorial/products_pipeline/multi_product.rst
```diff
@@ -0,0 +1,131 @@
...
+    # (in different machines, or different Conan caches)
+    $ conan install --requires=engine/1.0 --build=engine/1.0
+    $ conan install --requires=engine/1.0 --build=engine/1.0 -s build_type=Debug
+
+    # Once engine/1.0 builds finish, it is possible
+    # to build these 2 binaries in parallel (in different machines or caches)
+    $ conan install --requires=game/1.0 --build=game/1.0
+    $ conan install --requires=game/1.0 --build=game/1.0 -s build_type=Debug
+
+In this section we have still omitted some important implementation details that will follow in next sections. The goal was to focus on the ``conan graph build-order-merge`` command and how different products and configurations can be merged in a single "build-order".
```
AbrilRBS, line 131: > some important implementation details that will follow in next sections

I'd add some more info, like a "next sections such as..." to give a bit por a peek into the whole picture

ci_tutorial/products_pipeline/multi_product.rst
```diff
@@ -0,0 +1,131 @@
...
+                            }
+                        ]
+                    ]
+                }
+            ]
+        ]
+    }
+
+
+This build order summarizes the necessary builds. First it is necessary to build all different binaries for ``engine/1.0``. This recipe contains 2 different binaries, one for Release and the other for Debug. These binaries belong to the same element in the ``packages`` list, which means they do not depend on each other and can be built in parallel. Each binary tracks its own original build-order file with ``"filenames": ["game_release"],`` so it is possible to deduce the necessary profiles to apply to it.
```
AbrilRBS, line 113: Update this line if/when https://github.com/conan-io/conan/pull/17102 is merged

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+
+    $ conan graph build-order-merge 
+      --file=game_release.json --file=game_debug.json 
+      --file=mapviewer_release.json --file=mapviewer_debug.json 
+      --reduce --format=json > build_order.json
+
+
+    
+So far, this process has been almost identical to the previous section one, just with the difference of capturing and using a lockfile.
+Now, we will explain the "core" of the ``products`` pipeline: iterating the build-order and distributing the build, gathering the
```
AbrilRBS, line 86:
```suggestion
Now, we will explain the "core" of the ``products`` pipeline: iterating the build-order and distributing the build, and gathering the 
```

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+      --file=mapviewer_release.json --file=mapviewer_debug.json 
+      --reduce --format=json > build_order.json
+
+
+    
+So far, this process has been almost identical to the previous section one, just with the difference of capturing and using a lockfile.
+Now, we will explain the "core" of the ``products`` pipeline: iterating the build-order and distributing the build, gathering the 
+resulting built packages.
+
+This would be some Python code that performs the iteration sequentially (a real CI system would distribute the builds to different agents in parallel):
```
AbrilRBS, line 89:
```suggestion
This would be an example of some Python code that performs the iteration sequentially (a real CI system would distribute the builds to different agents in parallel):
```

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+  $ conan upload -l=built.json -r=products -c --format=json > uploaded3.json
+
+  # game/1.0 debug
+  $ conan install --requires=game/1.0 --build=game/1.0 --lockfile=conan.lock 
+    -s build_type=Debug --format=json > graph.json
+  $ conan list --graph=graph.json --format=json > built.json
+  $ conan upload -l=built.json -r=products -c --format=json > uploaded4.json
+
+
+After this step the new built packages will be in the ``products`` repo and we will have 4 ``uploaded1.json`` - ``uploaded4.json`` files.
```
AbrilRBS, line 161:
```suggestion
After this step the newly built packages will be in the ``products`` repo and we will have 4 ``uploaded1.json`` - ``uploaded4.json`` files.
```

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Simplifying the  different release and debug configurations, the state of our repositories would be something like:
```
AbrilRBS, line 163:
```suggestion
Simplifying the different release and debug configurations, the state of our repositories would be something like:
```

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+We can now accumulate the different ``uploadedX.json`` files into a single package list ``uploaded.json`` that contains everything:
+
+.. code-block:: bash
+
+    $ conan pkglist merge -l uploaded0.json -l uploaded1.json 
+      -l uploaded2.json -l uploaded3.json 
+      --format=json > uploaded.json
+
+
+And finally, if everything worked well and we consider this new set of versions and new package binaries is ready to be used by developers and other CI jobs, then, we can run the final promotion from the ``products`` to the ``develop`` repository:
```
AbrilRBS, line 231:
```suggestion
And finally, if everything worked well, and we consider this new set of versions and new package binaries is ready to be used by developers and other CI jobs, then we can run the final promotion from the ``products`` to the ``develop`` repository:
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
+Products pipeline: distributed build
+====================================
+
+
+The previous section used ``--build=missing`` to build all the necessary packages in the same CI machine.
+This is not always desired, or even possible, and in many situations it is preferable to do a distributed build, to achieve faster builds and better use the CI resources. The most natural distribution of the build load is to build different packages in different machines. Let's see how this is possible with the ``conan graph build-order`` command.
```
AbrilRBS, line 6:
```suggestion
This is not always desired, or even possible, and in many situations it is preferable to do a distributed build, to achieve faster builds and better usage the CI resources. The most natural distribution of the build load is to build different packages in different machines. Let's see how this is possible with the ``conan graph build-order`` command.
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+    # If you did in previous sections, NO need to repeat this
+    $ conan remote remove "*"  # Make sure no other remotes defined
+    # Add products repo, you might need to adjust this URL
+    # NOTE: The products repo is added first, it will have higher priority.
+    $ conan remote add products http://localhost:8081/artifactory/api/conan/products
+    # Add develop repo, you might need to adjust this URL
+    $ conan remote add develop http://localhost:8081/artifactory/api/conan/develop
+
+
+We will obviate by now the ``mapviewer/1.0`` and focus this section in the ``game/1.0`` product.
```
AbrilRBS, line 29:
```suggestion
We will obviate by now the ``mapviewer/1.0`` product and focus this section in the ``game/1.0`` product.
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+The first step is to compute the "build-order", that is, the list of packages that need to be build, and in what order.
```
AbrilRBS, line 30:
```suggestion
The first step is to compute the "build-order", that is, the list of packages that need to be built, and in what order.
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+
+.. code-block:: bash
+
+    $ conan graph build-order --requires=game/1.0 --build=missing 
+      --order-by=recipe --reduce --format=json > game_build_order.json
+
+Note a few important points:
+
+- It is necessary to use the ``--build=missing``, in exactly the same way than in the previous section. Failing to provide the intended ``--build`` policy and argument will result in incomplete or erroneous build-orders.
+- The ``--reduce`` eliminates all elements in the result that doesn't have the ``binary: Build`` policy. This means that the resulting "build-order" cannot be merged with other build order files for aggregating them into a single one, which is important when there are multiple configurations and products.
```
AbrilRBS, line 41:
```suggestion
- The ``--reduce`` argument eliminates all elements in the resulting order that don't have the ``binary: Build`` policy. This means that the resulting "build-order" cannot be merged with other build order files for aggregating them into a single one, which is important when there are multiple configurations and products.
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+- The ``--order-by`` argument allows to define different orders, by "recipe" or by "configuration". In this case, we are using the ``--order-by=recipe`` which is intended to parallelize builds per recipe, that means, that all possible different binaries for a given package like ``engine/1.0`` should be built first before any consumer of ``engine/1.0`` can be built.
```
AbrilRBS, line 42:
```suggestion
- The ``--order-by`` argument allows to define different orders, by "recipe" or by "configuration". In this case, we are using ``--order-by=recipe`` which is intended to parallelize builds per recipe, that means, that all possible different binaries for a given package like ``engine/1.0`` should be built first before any consumer of ``engine/1.0`` can be built.
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+
+
+.. image:: ./build_order_simple.png
+   :width: 500 px
+   :align: center
+
+
+The resulting json contains an ``order`` element which is a list of lists. This arrangement is important, every element in the top list is a set of packages that can be built in parallel because they do not have any relationship among them. You can view this list as a list of "levels", in level 0, there are packages that have no dependencies to any other package being built, in level 1 there are packages that contain dependencies only to elements in level 0 and so on.
+
+Then, the order of the elements in the top list is important and must be respected. Until the build of all the packages in one list item has finished, it is not possible to start the build of the next "level".
```
AbrilRBS, line 98:
```suggestion
Then, the order of the elements in the outermost list is important and must be respected. Until the build of all the packages in one list item has finished, it is not possible to start the build of the next "level".
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+
+Using the information in the ``graph_build_order.json`` file, it is possible to execute the build of the necessary packages, in the same way that the previous section ``--build=missing`` did, but not directly managed by us.
```
AbrilRBS, line 100:
```suggestion
Using the information in the ``graph_build_order.json`` file, it is possible to execute the build of the necessary packages, in the same way that the previous section's ``--build=missing`` did, but not directly managed by us.
```

AbrilRBS (approved):
Great job! I have a few suggestions, but otherwise this is mostly a wording review

Artalus (commented):
A magnificent read! Really glad that what I had in mind for our CI so far fits nicely with many points described here! Left couple of comments on something that I think the tutorial should explain a bit more.

Out of curiosity: how does a CCI fork fit into this picture? Does it essentially become just one more source of `packages` that then get promoted into `products`?

ci_tutorial/packages_pipeline.rst
```diff
@@ -0,0 +1,41 @@
+Packages pipeline
+==================
+
+
+The **packages pipeline** will build, create and upload the package binaries for the different configurations and platforms, when some
+developer is submitting some changes to one of the organization repositories source code. For example if a developer is doing some changes
+to the ``ai`` package, improving some of the library functionality, and bumping the version to ``ai/1.1.0``. If the organization needs to
+support both Windows and Linux platforms, then the package pipeline will build the new ``ai/1.1.0`` both for Windows and Linux, before
+considering the changes are valid. If some of the configurations fail to build under a specific platform, it is common to consider the
+changes invalid and stop the processing of those changes, until the code is fixed.
```
jasal82, line 10: What I'm missing in this workflow is a solution for 'nightly' and MR/PR builds. We're usually implementing new features on feature branches which then run MR pipelines to verify compilation and unit test integrity. Packages generated from these feature branches should be consumable by others for integration testing, etc. Also, after integrating an MR into the mainline branch we don't want to create a release yet. We want to have a nightly build and package that can also be consumed by other projects. In Conan 1 we used channels to distinguish between the different package types, but as far as I understood, these should no longer be used for this purpose. However, I don't see a way to do the same thing by leveraging multiple repositories, because in this scenario there is no way to mix dependencies from release, feature branch and nightly channels without creating temporary repositories and fiddling around with the remote order.

jasal82: It's also unclear how to version nightly and MR builds. We haven't decided about a release version yet so we could either use
- prerelease suffixes like `pkg/1.2-pre.1`, but they can only be enabled/disabled in the consumer conanfile - which means changing the conanfile all the time - or globally for all packages, which we don't want, because it's not selective enough
- 'nightly' as a version string, like `pkg/nightly` - this is difficult for the consumer because it's unclear what the current nightly revision is based on, the string is semver incompatible, older revision cannot be told apart easily and there's no automatic resolving possible against '*' or version ranges
- user or channels, like `pkg/1.2@nightly` (discouraged)

jasal82: And another open question is how to disambiguate between packages with the same name coming from different contexts. For example, we internally mirror Conan Center Index and make these recipe available for consumption in our recipes. However, there are some cases where we have existing company-specific recipes that happen to have the same name as recipes in the Conan Center Index. This would cause clashes when trying to use both the index mirror and the internal deployment remotes in the same configuration. I think it would be good to be able to specify a source remote for each of the requirements to make Conan consider only that remote. Other dependency managers use this solution (e.g. Cargo).

jcar87: > This would cause clashes when trying to use both the index mirror and the internal deployment remotes in the same configuration. I think it would be good to be able to specify a source remote for each of the requirements to make Conan consider only that remote. Other dependency managers use this solution (e.g. Cargo).

There are a few ways of handling that with Conan currently - when solving a graph to generate a lockfile, you can specify which remotes are considered - so the lockfile would then only show the revisions that are available in the remote you want, and subsequent calls would only resolve those.

Alternatively, the remote configurations do support a filter of which packages are allowed to be considered from a remote, see the `--allowed-packages` option in `conan remote add` and `conan remote update`.
You can restrict a remote to only serve certain packages, or to exclude some packages, for example the following would prevent the cmake package from being considered from the Conan Center remote.


```
conan remote update --allowed-packages='~cmake/*' conancenter
```

jasal82: @jcar87 Thanks for the quick answer. Both options could work but I think they're moving the decision to the wrong layer, from an architectural point of view. I'd rather have the recipe decide about the package sources because only the recipe knows what package it needs to be buildable. This should not be injected via environment or Conan config. It is good that it CAN be injected that way, because sometimes you want to influence the build from the pipeline for example, but it should not be the recommended way for handling package source selection for recipe dependencies.

ci_tutorial/products_pipeline.rst
```diff
@@ -0,0 +1,103 @@
+Products pipeline
+==================
+
+The **products pipeline** responds a more challenging question: does my "products" build correctly with the latest changes and new versions that have been done
```
mmmfarrell, line 4:
```suggestion
The **products pipeline** responds to a more challenging question: does my "products" build correctly with the latest changes and new versions that have been done
```

mmmfarrell: in general a few grammar / wording choice things in this paragraph. consider taking a second look

including "tested against the organization's important products" (note the missing `'s`)

ci_tutorial/products_pipeline.rst
```diff
@@ -0,0 +1,103 @@
...
+different customers, we could have ``game/1.0`` and ``game/2.3`` as well as different versions of ``mapviewer`` as products.
+
+The "products" approach, besides the advantage of focusing on the business value, has another very important advantage: it avoids having to model
+the dependency graph at the CI layer. It is a frequent attempt trying to model the inverse dependency model, that is, representing at the CI level
+the dependants or consumers of a given package. In our example, if we had configured a job for building the ``ai`` package, we could have another
+job for the ``engine`` package, that is triggered after the ``ai`` one, configuring such topology somehow in the CI system.
+
+But this approach does not scale at all and have very important limitations:
+  
+- The example above is relatively simple, but in practice dependency graphs can have many more packages, even several hundrends, making it very tedious and error prone to define all dependencies among packages in the CI
```
mmmfarrell, line 36:
```suggestion
- The example above is relatively simple, but in practice dependency graphs can have many more packages, even several hundreds, making it very tedious and error prone to define all dependencies among packages in the CI
```

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+          }
+        }
+    }
+
+
+This state of the ``develop`` repository will have the following behavior:
+
+- Developers installing ``game/1.0`` or ``engine/1.0`` will by default resolve to latest ``ai/1.1.0`` and use it. They will find pre-compiled binaries for the dependencies too, and they can continue developing using the latest set of dependencies.
+- Developers and CI that were using a lockfile that was locking ``ai/1.0`` version, will still be able to keep working with that dependency without anything breaking, as the new versions and package binaries do not break or invalidate the previous existing binaries.
+
```
mmmfarrell, line 303: It'd be nice to have some commentary on what failure modes are likely to arise and how they should be handled.

e.g. what if an intermediate dependency build fails due to one of the changed packages during the product pipeline?

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+.. note::
+
+  - This code is specific for the ``--order-by=recipe`` build-order, if chosing the ``--order-by=configuration``, the json
+    is different and it would require a different iteration.
+
+
+These are the tasks that the above Python code is doing:
+
+- For every ``package`` in the build-order, a ``conan install --require=<pkg> --build=<pkg>`` is issued, and the result of this command is stored in a ``graph.json`` file
```
NilsKlimanis, line 126: Is there a special reason, why `conan install`is used here, instead of `conan create`?
I'm also asking because `conan install` does not seem to execute `test`...

devops/package_promotions.rst
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+
+When the QA team tests and approves these packages, they can be promoted to the "release" repository.
+Basically, a promotion is a copy of the packages, including all the artifacts and metadata from the
+"testing" to the "release" repository.
+
+
+There are different ways to implement and execute a package promotion. Artifactoy has some APIs that can be
```
mmmfarrell, line 82:
```suggestion
There are different ways to implement and execute a package promotion. Artifactory has some APIs that can be
```

ci_tutorial/products_pipeline/full_pipeline.rst
mmmfarrell: @memsharded More specific question/example about rolling out a major version change.

If packages pipeline only use the `develop` repository, and never the `products` repository, how are major versions rolled out?
e.g. if we create a breaking API change in ai/2.0.
Based on my read, this would be the order of events:
- ai/2.0 passes the packages pipeline, using mathlib/1.0 from develop
- ai/2.0 is promoted to the products repository
- products pipeline would run and pass, but not use ai/2.0 b/c engine depends on ai/[>=1.0 <2]
- so ai/2.0 is stuck in products repository until engine updates its version range
- engine updates its version range to ai/[>=2.0 <3], but its package pipeline won't be able to find ai/2.0 b/c it is only looking in develop

It seems like the packages pipeline needs to look at both develop and product repositories so changes like this can work up the chain.

My questions are:
- does this match your experience? should packages pipeline look at both develop and product repositories? (if so, probably good to mention that in the tutorial)
- What order should it look at develop/product repositories? Currently the product pipeline adds product repo first,then develop. This makes sense for the product pipeline that is trying to promote packages from product to develop. However, it may make sense for the packages pipeline to be reversed, first look at the stable develop, second look at the products repo only if a suitable stable package isn't found

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,195 @@
...
+
+Let's begin cleaning our cache and initializing only the ``develop`` repo:
+
+
+.. code-block:: bash
+
+    $ conan remove "*" -c  # Make sure no packages from last run
+    $ conan remote remove "*"  # Make sure no other remotes defined
+    # Add develop repo, you might need to adjust this URL
+    $ conan remote add develop http://localhost:8081/artifactory/api/conan/develop
```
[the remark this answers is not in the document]
memsharded: I agree that readers should be aware of better practices, and re-defining remotes is not great here.
On the other hand, I'd like the tutorial to use as much as possible basic commands, not high level operations defined exclusively for the tutorial. So I have done the following:
- Rewrote the full example and docs to not add and remove remotes anymore
- The remotes are always defined
- The remotes are ``conan remote enable/disable`` at very specific points. This is better practice
- Also, it no longer requires changing URLs, as they are always defined

ci_tutorial/packages_pipeline/multi_configuration.rst
memsharded: I have re-written the tutorial to fully avoid URLs and adding/removing repos unnecessary, just ``conan remote enable/disable`` repos, in less occasions.

ci_tutorial/packages_pipeline/multi_configuration.rst
memsharded: Fixed

ci_tutorial/packages_pipeline/single_configuration.rst
memsharded: It has a purpose. The generated files in the commands are created inside the ``ai`` folder, if we move location. So they pollute less the overall project and it makes more evident that they are due to some source code changes of the ``ai`` package, but not others.

ci_tutorial/packages_pipeline/multi_configuration_lockfile.rst
memsharded: I see that, but as commented above, the problem is that if we don't ``cd ai``, the generated files will pollute the root of the project, making it worse in my opinion: ``graph.json``, ``built.json``, ``uploaded_debug.json``, ``uploaded_release.json``, ``uploaded.json``, and it is not evident if they are in the root to which package or event they belong to, at least if they are inside ``ai`` folder, it is more clear they are part of the "package pipeline" of ``ai``.

memsharded: I am adding a comment like:
```
$ cd ai  # If you were not inside "ai" folder already
```

ci_tutorial/products_pipeline/distributed_build.rst
```diff
@@ -0,0 +1,113 @@
...
+    $ conan remote add develop http://localhost:8081/artifactory/api/conan/develop
+
+
+We will obviate by now the ``mapviewer/1.0`` and focus this section in the ``game/1.0`` product.
+The first step is to compute the "build-order", that is, the list of packages that need to be build, and in what order.
+This is done with the following ``conan graph build-order`` command:
+
+.. code-block:: bash
+
+    $ conan graph build-order --requires=game/1.0 --build=missing
```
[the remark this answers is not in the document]
memsharded: Done

ci_tutorial/products_pipeline/distributed_build.rst
memsharded: I have renamed this section, now it is called "the build-order", which I think is better than "distributed build", because it is a bit too early to introduce that concept here, and it is delayed until last scenario.

devops/versioning/default.rst
memsharded: The ``user/channel`` and other versioning, CI triggering and branch merging, and similar strategies will follow in a later section of the tutorial, at this stage I think it is better to keep it as simple as possible and focus on the pure integration and build.

ci_tutorial/products_pipeline/full_pipeline.rst
memsharded: Yes, building from Conan recipes in the server is way simpler. Otherwise it is necessary to obtain the exact ``scm`` coordinates (recipes must implement the scm-capture strategy), do the clone+checkout then the ``conan create``. 

It has the advantage that it runs the ``test_package``, but this ``test_package`` is focused on checking the package correctness (the headers have been packaged, the libs too, etc), it is not a functional test. As in a CI pipeline packages are being re-built, but their code or recipes is not changed at all, it is almost impossible that the ``test_package`` would catch any significant package issue, as they are extremely unlikely. And the regular testing that happens in the ``build()`` when activated, and the testing performed by the consumers of packages that have been re-built is even a better check of the package correctness, as they are the real consumers.

memsharded: We will most likely add examples in a later section of the docs with the full ``scm`` + clone + checkout + ``conan create`` flow. Specially because it will most likely be highly simplified by new ``conan workspace`` commands.

But at this point, it is better to focus on the basic concepts of the CI process, the simpler the examples and commands the better.

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,195 @@
...
+
+We will create the packages for the 2 configurations sequentially in our computer, but note these will typically run
+in different computers, so it is typical for CI systems to launch the builds of different configurations in parallel.
+
+.. code-block:: bash
+    :caption: Release build
+
+    $ cd ai
+    $ conan create . --build="missing:ai/*" -s build_type=Release --format=json > graph.json
+    $ conan list --graph=graph.json --graph-binaries=build --format=json > built.json
```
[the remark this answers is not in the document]
memsharded: Not sure what does this mean in this context. We are not doing ``--build=missing``, but just for ``ai``. Dependencies aren't built and won't be uploaded.

ci_tutorial/packages_pipeline/multi_configuration.rst
```diff
@@ -0,0 +1,195 @@
...
+- Finally, we will upload the ``built.json`` package list to the ``packages`` repository, creating the ``uploaded_release.json``
+  package list with the new location of the packages (the server repository).
+
+Likewise, the Debug build will do the same steps:
+
+
+.. code-block:: bash
+    :caption: Debug build
+
+    $ conan create . --build="missing:ai/*" -s build_type=Debug --format=json > graph.json
```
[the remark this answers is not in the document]
memsharded: Pre exporting recipes might help in some scenarios like reducing concurrency issues if building in parallel in the same machine, and it can also help when there are changes in multiple repositories/packages simultaneously. In this case, we are using the basic case of only ``ai`` having changes, so a ``conan create`` should be good enough, it would be simpler for readers.

ci_tutorial/tutorial.rst
memsharded: Done

ci_tutorial/products_pipeline/multi_product.rst
memsharded: Done

ci_tutorial/products_pipeline/multi_product.rst
memsharded: Done

ci_tutorial/products_pipeline/multi_product.rst
```diff
@@ -0,0 +1,124 @@
...
+    # (in different machines, or different Conan caches)
+    $ conan install --requires=engine/1.0 --build=engine/1.0
+    $ conan install --requires=engine/1.0 --build=engine/1.0 -s build_type=Debug
+
+    # Once engine/1.0 builds finish, it is possible
+    # to build these 2 binaries in parallel (in different machines or caches)
+    $ conan install --requires=game/1.0 --build=game/1.0
+    $ conan install --requires=game/1.0 --build=game/1.0 -s build_type=Debug
+
+In this section we have still omitted some important implementation details that will follow in next sections. The goal was to focus on the ``conan graph build-order-merge`` command and how different products and configurations can be merged in a single "build-order". Next section will with more details how this build-order can be really distributed in CI, using lockfiles to guarantee constant dependencies.
```
AbrilRBS, line 124:
```suggestion
In this section we have still omitted some important implementation details that will follow in next sections. The goal was to focus on the ``conan graph build-order-merge`` command and how different products and configurations can be merged in a single "build-order". The next section will show with more details how this build-order can be really distributed in CI, using lockfiles to guarantee constant dependencies.
```

ci_tutorial/products_pipeline.rst
```diff
@@ -0,0 +1,103 @@
...
+- The "inverse" dependency model, that is, asking what are the "dependants" of a given package is extremely challeging in practice, specially in a decentralized
+  approach like Conan in which packages can be stored in different repositories, including different servers, and there isn't a central database of all packages and their relations.
+  Also, the "inverse" dependency model is, similar to the direct one, conditional. As a dependency can be conditional on any configuration (settings, options), the inverse is
+  also conditioned to the same logic, and such logic also evolves and changes with every new revision and version.
+
+In C and C++ projects the "products" pipeline becomes more necessary and critical than in other languages due to the compilation model with headers textual inclusions becoming part of the consumers' binary artifacts and due to the native artifacts linkage models. 
+
```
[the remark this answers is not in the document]
memsharded: Thanks for the feedback!!

> Out of curiosity: how does a CCI fork fit into this picture? Does it essentially become just one more source of packages that then get promoted into products?

Our own CCI uses the exact same principles, tools and flows. There are some differences, to handle with the scale, the reporting, security, etc.

For users CCI forks, the flow might change a little bit, because it is likely that the main flow is to update, merging from upstream, and there might be several packages affected at the same time. For this flow, doing a batch ``conan export`` operation for all the intended packages, then doing the ``conan graph build-order`` for the different configurations, using the exported changes, that part would be basically the same. So mainly the initial part, as the tutorial focused on changes per-package that lives in different git repos, one repo per package, while conan-center-index updates will likely be multi-package, and in the same repo.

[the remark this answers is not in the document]
memsharded: > Do I, as a CI engineer, just talk to the Important People and then proclaim "game and mapviewer has been declared as the main and only products, and thus we have CI only for them"?
Or should this be a more dynamic and programmatic approach, in essence determining all the root nodes in the dependency graph?
Or must I not even think about traversing the graph at all, since I will inevitably fall victim to the same issue of conditional dependencies?

Almost any attempt to try to deduce dynamically the "products" will most likely be worse than explicit modeling of those. Note that for a start, organizations have packages distributed accross many repositories, some global to the org, some specific for teams, etc. Packages often count in many thousands of them, each one with multiple versions and revisions. It becomes intractable pretty easily. 
Also, any automatic approach is completely unable to capture any business logic and decision.

So mostly, it is a explicit definition, not by the CI engineer, but it comes from the business: This org/team is delivering these 3 products, ``app1``, ``app2`` and ``app3``. For each one of them we might need to maintain different versions, as we have customers that depend on them. So our products are ``app1/2.3``, ``app1/3.8``, ``app2/1.1`` and ``app3/1.2``, ``app3/1.5``. When our customers upgrade we can safely remove some of these versions from our products, and we can also incorporate new versions of the products for the forecoming new releases. 

These products are the things that shouldn't be broken when we create new versions of the packages. For legacy products no longer in our current product list, we will be able to re-build them, if we captured the recommended lockfiles when they were last built, but if we don't use the lockfile to constraint new dependencies versions, they might indeed break if trying to use latest versions. This is expected and desired, we didn't want to burn CI for a product that is no longer maintained and evolved!

ci_tutorial/products_pipeline/full_pipeline.rst
```diff
@@ -0,0 +1,303 @@
...
+            rankdir="BT";    
+          }
+        }
+    }
+
+
+This state of the ``develop`` repository will have the following behavior:
+
+- Developers installing ``game/1.0`` or ``engine/1.0`` will by default resolve to latest ``ai/1.1.0`` and use it. They will find pre-compiled binaries for the dependencies too, and they can continue developing using the latest set of dependencies.
+- Developers and CI that were using a lockfile that was locking ``ai/1.0`` version, will still be able to keep working with that dependency without anything breaking, as the new versions and package binaries do not break or invalidate the previous existing binaries.
```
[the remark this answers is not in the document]
memsharded: > I think the doc may also clarify what happens with the lockfile later in this workflow...

That is a good question. I will add some lines to discuss this, but in short, I don't think there is a unique good approach for this, and it might depend on many other business and development factors. We will also try to elaborate a bit more on future sections of the docs.

> Is it supposed to be updated later, once the Important People decide that updating game to use ai/1.1.0 is a Good Idea?

The ``conan.lock`` used at this point has already been updated. It contains the ``ai/1.1.0``. So no scripting magic necessary at this point.

I have added a couple of paragraphs commenting this, but still very general, this will most likely get further details in the future.

ci_tutorial/products_pipeline/full_pipeline.rst
memsharded: Very good questions, thanks very much for your feedback @mmmfarrell 

> so ai/2.0 is stuck in products repository until engine updates its version range

Not really!
If you create ``ai/2.0``, and everything is green, the products are ok, etc, because they are not using ``ai/2.0`` the end result is that ``ai/2.0`` gets promoted to the ``develop`` repo. From there it is available for everyone else to keep working downstream to adapt and start using the new ``ai/2.0`` package.
In this case, developers can start working on a ``engine/1.1`` (it is a new version, as it will need source code changes to adapt to the braking ``ai/2.0``), that uses ``ai/2.0``. Only when this new ``engine`` version is green, it will promote to ``develop`` so down to ``game``, the whole app uses the new ``ai/2.0``.

It is true that if a major version happens in a higher package, for example in ``mathlib/2.0``, the CI will need to manage changes from more than 1 PR:
- 1 PR for the ``ai`` package, to use the new ``mathlib/2.0``. It will be blocked in ``products`` pipeline, because the products will break, as ``graphics`` hasn't still upgraded. The packages will be in the ``products`` repos, as they passed correctly the ``packages`` piepline.
- Another PR for the ``graphics`` package, that will also put the binaries in the ``products`` repository.
- Now that both new binaries for both ``ai`` and ``graphics`` have accounted for the changes in ``mathlib/2.0``, now a products pipeline can succeed, and promote both the packages of those PRs that were in the products pipeline.

Basically, devs can work on the different packages PRs, that will accumulate in the ``products`` repository, until they pass all together.

This is exactly the type of discussions and examples that will follow after, the current section is focusing on the most common "happy path" of the CI.

memsharded: I have added some "Final remarks" to the tutorial, to clarify the scope and what will follow in later new sections.

czoido (approved):

ci_tutorial/packages_pipeline/multi_configuration.rst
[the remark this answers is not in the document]
Todiq: @memsharded 
Would it please be possible to also show how to decouple the `conan create` into `conan build` + `conan export-pkg` with the formatting? Should we create a graph for both of these commands then merge them?

It can be especially insteresting in CI, since creating an artifact with the whole conan cache (if built with `conan create`) can be cumbersome. This allows cases where one can create an artifact with only the `build` folder (if built with `conan build`)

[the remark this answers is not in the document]
memsharded: Hi @Todiq 

This would be a different topic than the one covered in this tutorial. This kind of things will probably follow in later versions, talking about implementation strategies like this.

However, in CI a ``conan build + conan export-pkg`` would be identical to ``conan create``, there will be no difference in computing time or storage in disk or anything like that. It only changes the location of the build folder, but actually it would be very recommended to use ``create`` because that garantees that future re-builds with ``conan install --requires=thispkg/version --build=thispkg/version`` work correctly, because those are also executed in the cache.

So I am not sure about the value of your proposal. ``conan build + conan export-pkg`` has some value for developers trying things locally, but it shouldn't have any advantage in the CI, I might be missing something.

PR #3804, opened 2024-07-30
Fix version mismatch in package_prebuilt_binaries.rst

hosseini92:
There was a mismatch between the version provided in the 'conan new ...' command and the rest of the documentation, including the 'conan test ...' command.


memsharded (approved):
Thanks very much for fixing this!

PR #3807, opened 2024-08-07
Update compression formats

andrey-zherikov:
Synchronize with [`conan.tools.files.unzip`](https://github.com/conan-io/conan/blob/c55ab23/conan/tools/files/files.py#L285-L299)


memsharded (approved):
Thanks for the contribution

PR #3811, opened 2024-08-13
Document how to preserve PkgConfig data from package method

uilianries:
Hello! 

Originally we have few issues in ConanCenterIndex affected by the scenario where `pkg-config` is not installed installed in the system, so users can not consume those packages that are simple wrappers for system libraries:

- https://github.com/conan-io/conan-center-index/issues/23009
- https://github.com/conan-io/conan-center-index/issues/14909

Indeed I was caught by this error when using a raw docker image to validate a reported bug in Ubuntu 24.04. This is my motivation for writing this PR.

As possible alternative (experimental), we can run `PkgConfig` from `package()` and store its data in a json file, then, restore that information in `self.cpp_info`. This workaround was commented by Memsharded 1 year ago: https://github.com/conan-io/conan/issues/14422#issuecomment-1665166761

This scenario is tested already in Conan client, but not well explored, still, is working in minimal case: https://github.com/conan-io/conan/blob/c55ab23d01f1336c850816e07c68ae19704201d6/test/functional/toolchains/gnu/test_pkg_config.py#L99

Related to https://github.com/conan-io/conan/issues/14422.

Current preview (commit c6c8f72):

Screenshot 2024-08-13 at 12-15-11 PkgConfig — conan 2 6 0 documentation

### DIFF
--- a/reference/tools/gnu/pkgconfig.rst
+++ b/reference/tools/gnu/pkgconfig.rst
@@ -25,8 +25,11 @@ Read a ``pc`` file and access the information:
     print(pkg_config.variables['prefix']) # something like'/usr/local'
 
 
-Use the ``pc`` file information to fill a ``cpp_info`` object:
+Using PkgConfig to fill a ``cpp_info`` object
+---------------------------------------------
 
+The ``PkgConfig`` class can be used to fill a ``CppInfo`` object with the information that will be consumed by ``PkgConfigDeps`` generator later.
+This is a useful feature when packaging a system library that provides a ``.pc`` file, or when a proprietary package has a build system that only outputs ``.pc`` files.
 
 .. code-block:: python
 
@@ -35,6 +38,45 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
         pkg_config.fill_cpp_info(self.cpp_info, is_system=False, system_libs=["m", "rt"])
 
 
+However, ``PkgConfig`` will invoke the ``pkg-config`` executable to extract the information from the ``.pc`` file.
+The ``pkg-config`` executable must be available in the system path for this case, otherwise, it will fail when installing the consumed package.
+
+
+Using pkg-config from Conan package instead of system
+-----------------------------------------------------
+
+.. include:: ../../../common/experimental_warning.inc
+
+In case not having ``pkg-config`` available in the system, it is possible to use the ``pkg-config`` executable provided by a Conan package:
+
+.. code-block:: python
+
+    import os
+    from conan import ConanFile
+    from conan.tools.gnu import PkgConfig
+    from conan.tools import CppInfo
+
+    class Pkg(ConanFile):
+
+        def tool_requires(self):
+            self.requires("pkgconf/[*]")
+
+        ...
+
+        def package(self):
+            pkg_config = PkgConfig(self, "libastral", pkg_config_path=".")
+            cpp_info = CppInfo(self)
+            pkg_config.fill_cpp_info(cpp_info, is_system=False, system_libs=["m", "rt"])
+            cpp_info.save(os.path.join(self.package_folder, "cpp_info.json"))
+
+        def package_info(self):
+            self.cpp_info = CppInfo(self).load(os.path.join(self.package_folder, "cpp_info.json"))
+
+
+The ``pkg-config`` executable provided by the Conan package ``pkgconf`` will be invoked only when creating the Conan binary package.
+The ``.pc`` information will be extracted from the ``cpp_info.json`` file located in the package folder, it will fill the ``self.cpp_info`` object.
+This way, the ``PkgConfig`` will not need to invoke the ``pkg-config`` executable again to extract the information from the ``.pc`` file,
+when consuming the package.
 
 Reference
 ---------


reference/tools/gnu/pkgconfig.rst
```diff
@@ -35,6 +38,45 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
...
+However, ``PkgConfig`` will invoke the ``pkg-config`` executable to extract the information from the ``.pc`` file.
+The ``pkg-config`` executable must be available in the system path for this case, otherwise, it will fail when installing the consumed package.
+
+
+Using pkg-config from Conan package instead of system
+-----------------------------------------------------
+
+.. include:: ../../../common/experimental_warning.inc
+
+In case not having ``pkg-config`` available in the system, it is possible to use the ``pkg-config`` executable provided by a Conan package:
```
AbrilRBS, line 50:
```suggestion
In case of not having ``pkg-config`` available in the system, it is possible to use the ``pkg-config`` executable provided by a Conan package:
```

reference/tools/gnu/pkgconfig.rst
```diff
@@ -25,8 +25,11 @@ Read a ``pc`` file and access the information:
     print(pkg_config.variables['prefix']) # something like'/usr/local'
 
 
-Use the ``pc`` file information to fill a ``cpp_info`` object:
+Using PkgConfig to fill a ``cpp_info`` object
+---------------------------------------------
 
+The ``PkgConfig`` class can be used to fill a ``CppInfo`` object with the information that will be consumed by ``PkgConfigDeps`` generator later.
```
AbrilRBS, line 31:
```suggestion
The ``PkgConfig`` class can be used to fill a ``CppInfo`` object with the information that will be consumed by the ``PkgConfigDeps`` generator later.
```

reference/tools/gnu/pkgconfig.rst
```diff
@@ -35,6 +38,45 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
...
+
+    import os
+    from conan import ConanFile
+    from conan.tools.gnu import PkgConfig
+    from conan.tools import CppInfo
+
+    class Pkg(ConanFile):
+
+        def tool_requires(self):
+            self.requires("pkgconf/[*]")
```
AbrilRBS, line 62:
```suggestion
        def build_requirements(self):
            self.tool_requires("pkgconf/[*]")
```

AbrilRBS (approved):

reference/tools/gnu/pkgconfig.rst
uilianries: Autopilot!

reference/tools/gnu/pkgconfig.rst
```diff
@@ -35,6 +38,45 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
...
+        ...
+
+        def package(self):
+            pkg_config = PkgConfig(self, "libastral", pkg_config_path=".")
+            cpp_info = CppInfo(self)
+            pkg_config.fill_cpp_info(cpp_info, is_system=False, system_libs=["m", "rt"])
+            cpp_info.save(os.path.join(self.package_folder, "cpp_info.json"))
+
+        def package_info(self):
+            self.cpp_info = CppInfo(self).load(os.path.join(self.package_folder, "cpp_info.json"))
```
jcar87, line 73: If this is a "system" requirement (rather than parsing the `.pc` output generated inside a package folder), this can cause issues on systems where the contents of `.pc` files can change with system updates - and I wouldn't recommend it. This could end up with a situation where the `cpp_info.json` inside a pacakge contains information about a system dependency that is wrong and out of sync with what the system actually has - and could be quite difficult to troubleshoot.

This is less likely to happens on distros that keep a closed set of versions for a given version of the distro, but it can be problematic on distros that have rolling version updates, or if a user generates a package today, and tries to use it after a `dist-upgrade`. 

Remember that a call to `pkg-config` to locate "foo" may cause pkg-config to locate other libraries for transitive dependencies - and these can change across versions. I wouldn't recommend this for system dependencies.

On a Linux system, while inconvenient, I see no issue recommending that pkg-config is installed at the system level

memsharded: I totally agree this feature should be exclusively for compiled packages, not system dependencies. What are the use cases in ConanCenter?

jcar87: I think the motivating issue in Conan Center is precisely system packages (e.g. OpenGL). I can see 10 such cases in Conan Center where `pkg_config.fill_cpp_info` is called inside `package_info()`

memsharded: Thats bad. The only way it could be allowed is by marking these packages as ``upload = "skip"`` and ``build_policy = "missing"``, to make sure they are always locally created in the target machine, but that ``.json`` file is never shared or uploaded.

jcar87: Ha! I would haver never thought of that - that would work - the dependency on the Conan-provided pkfcong would satisfy this at the time of use, I think.

My alterantive was more blunt: if a "system" recipe lists the system package manager requirements, and we need `pkg-config` to parse the `.pc` files... just install `pkg-config` as part of the system dependency set. At the end of the day, if a user is completely happy doing `tools.system.package_manager:mode=install`, then I can't really see a scenario where they are happy doing that, but they draw the line at having pkg-config installed at the system level. I imagine that what we want here is convenience, but I don't think this is a pressing issue - we should do a better job documenting (in Conan Center, not necessarily in Conan) - which tools users are expected to have installed, and this is one of them.

uilianries:
@jcar87 @memsharded I just addressed your comments into the commit https://github.com/conan-io/docs/pull/3811/commits/e23b7c217dd462e992e392d53ce1b2aa2b8c5a18

- No more system package recommendation in the feature description
- Added Warning box to clarify that using system package with cached information is risky.

Please, review it again.

reference/tools/gnu/pkgconfig.rst
uilianries: > If this is a "system" requirement (rather than parsing the .pc output generated inside a package folder), this can cause issues on systems where the contents of .pc files can change with system updates

Agreed. Real example: CCI packages are using Ubuntu 16.04, the package vaapi uses the version 1.x. Ubuntu 22.04 supports the library 2.x. They have same library names etc, but are ABI incompatible. 

> The only way it could be allowed is by marking these packages as upload = "skip" and build_policy = "missing", to make sure they are always locally created in the target machine, but that .json file is never shared or uploaded.

Are you saying for any scenario? Should it be included by default in the documentation as well? 

> My alterantive was more blunt: if a "system" recipe lists the system package manager requirements, and we need pkg-config to parse the .pc files... just install pkg-config as part of the system dependency set... in each of those 10 recipes. At the end of the day, if a user is completely happy doing tools.system.package_manager:mode=install, then I can't really see a scenario where they are happy doing that, but they draw the line at having pkg-config installed at the system level.

Yes, totally true, for who is configuring a CI or build environment, is just one more command line to install it. Users have their excuses like: It's a dependency manager, but does not solve pkg-config as requirement, etc ...

In the past we needed to document about CMake not being present in every recipe as build requirement, because we had PRs adding it by default. I see pkg-config a very similar case. https://github.com/conan-io/conan-center-index/blob/f31b98bae9c200494240f9b475c16e055ff8077e/docs/faqs.md#why-recipes-that-use-build-tools-like-cmake-that-have-packages-in-conan-center-do-not-use-it-as-a-build-require-by-default

memsharded: The main issue is that the ``upload_policy = "skip"`` is only Conan 2, so it will be incorrect in Conan 1.X, so we cannot allow this until we have a Conan 2-only pipeline.

uilianries: But this documentation will be merged to Conan 2.x only (target is release/2.6). We have others features only compatible for Conan 2.x in docs. For CCI I got that we can not use it, because our main problem are system packages.

memsharded: Sure, as this was a ConanCenter driven thing, I was talking about CCI

reference/tools/gnu/pkgconfig.rst
```diff
@@ -35,6 +38,50 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
         pkg_config.fill_cpp_info(self.cpp_info, is_system=False, system_libs=["m", "rt"])
 
 
+However, ``PkgConfig`` will invoke the ``pkg-config`` executable to extract the information from the ``.pc`` file.
+The ``pkg-config`` executable must be available in the system path for this case, otherwise, it will fail when installing the consumed package.
+
+
+Using pkg-config from Conan package instead of system
```
memsharded, line 45: I'd say using ``pkg-config`` from a Conan ``tool_requires``.

reference/tools/gnu/pkgconfig.rst
```diff
@@ -35,6 +38,50 @@ Use the ``pc`` file information to fill a ``cpp_info`` object:
...
+
+
+The ``pkg-config`` executable provided by the Conan package ``pkgconf`` will be invoked only when creating the Conan binary package.
+The ``.pc`` information will be extracted from the ``cpp_info.json`` file located in the package folder, it will fill the ``self.cpp_info`` object.
+This way, the ``PkgConfig`` will not need to invoke the ``pkg-config`` executable again to extract the information from the ``.pc`` file,
+when consuming the package.
+
+.. warning::
+
+    It's not recommended to use this approach when creating a "system" package recipe, as the packaged information may not be compatible with the host system,
```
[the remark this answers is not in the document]
uilianries: Done, changes addressed to the commit https://github.com/conan-io/docs/pull/3811/commits/c3139187527d33139748bc9dda08c6ba74a2acb1

reference/tools/gnu/pkgconfig.rst
uilianries: Done, changes addressed to the commit https://github.com/conan-io/docs/pull/3811/commits/c3139187527d33139748bc9dda08c6ba74a2acb1

memsharded (approved):

czoido (approved):

rohel01:
Hello, this documentation update does not show up on the documentation site. I looked at the documentation for releases 2.6, 2.7, 2.15 and 2.16.
Am I missing something?

memsharded:
Hi @rohel01 

Thanks for noticing this!
It seems this PR was never merged, not sure what happened, let us check

rohel01:
Hello, any update on this?
The added documentation is very useful

jcar87:
> Hello, any update on this? The added documentation is very useful

Hi @rohel01, thanks for reaching out.

This has not been merged as we have gained a better understanding of pkg-config and have identified some issues in Conan Center.

Our **current** recommendations are as follows:
- All of the recipes in Conan Center that are wrappers for the system package manager on Linux and FreeBSD - the users must have the system's `pkg-config` (or equivalent) installed - this has always been the case, but is now reflected in the Conan Center documentation: https://github.com/conan-io/conan-center-index/blob/05bebf9e25d4af0abe883ad15c517a0803f0dd0c/docs/consuming_recipes.md#expected-environment . Would very strongly recommend to set the `tools.gnu:pkg_config` config in the profile, pointing to the system-provided `pkg-config` executable. There are mainly two reasons for this:
   - The `pkg-config` provided by the system package manager designed to work with the `.pc` files provided by the same system package manager. That is, it ***KNOWS*** to **not** emit `-I/usr/include`, `-L/usr/lib/`, `-L/usr/lib/x86_64-linux-gnu` or similar as flags. This is very important.
   - A `package_info()` method that invokes `pkg-config` (in Conan Center this only happens for recipes where the version is simply "system" - is not able to call `pkg-config` provided by a recipe - because `package_info()` is often invoked in a context where the graph is not resolved.
   

The approach proposed in this PR would be suitable for:
- a library that is built from source
- that produces a .pc file
- where we want to "lift" the data from the `.pc` file to fill the contents of `cpp_info` in the `package_info()` method
- there are no dependencies on "system" libraries that need to be resolved

let us know if these are the patterns that are relevant to you, or if we are missing something, thanks!

PR #3814, opened 2024-08-23
Fix version range syntax in conanfile_txt.rst

cubanpit:
I used this page as reference for version ranges, it took me a while to realize it is the only page in the documentation using a comma instead of a space to define a version range. I looked for `,<` in the documentation and this seems to be the only location with the wrong syntax.


memsharded (approved):
You are right, thanks very much for fixing this!

This will be in next release docs.

PR #3833, opened 2024-09-10
Updated docs for TriCore (#3831)

### LINKED ISSUE #3831: [bug] TriCore architecture inconsistencies in documentation
### Describe the bug

Hello!

I guess that *bug* is perhaps the wrong label for this issue, but it was the best suited label in my opinion!

In the documentation for TriCore (https://docs.conan.io/2/examples/cross_build/tricore.html), TriCore is denoted as a compiler. This is not strictly true, TriCore is instead the architecture: https://en.wikipedia.org/wiki/Infineon_TriCore, https://www.infineon.com/cms/en/product/microcontroller/32-bit-tricore-microcontroller/?gad_source=1&gclid=EAIaIQobChMIs-bByJe1iAMVWGWRBR3vnj0hEAAYASAAEgL_qfD_BwE&gclsrc=aw.ds.

There are multiple compilers for TriCore, some of which can be found here: https://www.infineon.com/cms/en/tools/aurix-tools/Compilers/. I guess that the one listed in the Conan documentation refers to this one: https://github.com/mc-cat-tty/tricore-gcc/tree/aeeeafeaa20072d7f1811890b113fe4e9c3f5407?

I don't know much about the nomenclature of these microcontrollers and architectures, but I suggest that you change the documentation to state that TriCore is the architecture and not the compiler. At a quick glance the profile suggestion looks correct, though, with `gcc` being stated as the compiler.

### How to reproduce it

_No response_

aander80:
TriCore was mentioned as a compiler instead of an architecture (for which there are multiple compilers). This has now been clarified.

This will close #3831.

### DIFF
--- a/examples/cross_build/tricore.rst
+++ b/examples/cross_build/tricore.rst
@@ -1,14 +1,18 @@
 .. _example_cross_build_tricore:
 
 
-Building packages with Tricore compiler
+Building packages for TriCore
 =======================================
 
 .. include:: ../../common/experimental_warning.inc
 
 
-Tricore compiler is a popular compiler for embedded in some domains such as automotive.
-Since Conan 2.7 there is some built-in support for this compiler:
+TriCore is an embedded microcontroller architecture used in multiple domains such as automotive.
+There are multiple compilers for TriCore, some of which can be found `here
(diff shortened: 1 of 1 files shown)


memsharded (approved):
Excellent, many thanks!

memsharded:
This will be published in next Conan 2.8 release

Issue #3832, opened 2024-09-09
self.output in Conan 2.2


Ed-CMI:
### What is your question?

The following link shows some information about different methods for self.output. Is there any more documentation regarding the different methods. For ex, what would be the benefits or difference between using the debug verse trace method in self.output?

https://docs.conan.io/2.2/reference/conanfile/running_and_output.html

### Have you read the CONTRIBUTING guide?

- [ ] I've read the CONTRIBUTING guide

ErniGH:
Hey @Ed-CMI 👋 , unfortunately, we don't have specific documentation for each method. In the link you provided, you can see all the methods related to output, ordered from the most verbose to the least. They all display a message on the screen, but the difference between them lies in when they are shown, depending on the verbosity level you set for Conan.

Most commands have the option `-v [V]` option with valid levels from less to more verbose: `-vquiet`, `-verror`, `-vwarning`, `-vnotice`, `-vstatus`, `-v or -vverbose`, `-vv or -vdebug`, `-vvv` or `-vtrace`.

For example, `-vquiet` doesn't display any messages, `-verror` only shows error messages, `-vwarning` shows both error and warning messages, and so on.

You can also set the verbosity using the `CONAN_LOG_LEVEL` environment variable.
https://docs.conan.io/2/reference/environment.html#logging

I hope this helps! Feel free to reach out if you have any other questions.

Ed-CMI:
Can you please more details on trace, debug, and verbose methods? I would like to understand the difference between these methods to try and select the best method to print out generic information for the user that might help them troubleshoot issues in the recipe files.

ErniGH:
Essentially, the three methods are the same; the only difference is the level of verbosity at which they are displayed. By default, none of them are shown. From least specific to most specific, we have:

* **Verbose**: With a moderately high level of detail, it is typically used to obtain general information about the functioning of the process. In general, you will want to use this for most of the additional information you occasionally want to display.

* **Debug**: With a high level of detail, it is usually reserved for showing the values of variables. As the name suggests, it is appropriate to use when you are in a debugging scenario and want to view important variables.

 * **Trace**: With an extremely high level of detail, it is reserved for extremely granular details. Here we can include everything from function entry and exit, values of less important variables, and other very specific events. The use of **trace** is rare, but on special occasions when we need incredibly detailed information, we can use it.

Ed-CMI:
Thanks for the explanation.  Can you please update the training material for conan with these types of details. I will close this ticket.

[...]

PR #3838, opened 2024-09-12
Add extra information for the output methods

ErniGH:
Add extra information about the output methods.

This was asked in #3832


czoido:
It would be better that this PR is pointed to the release/2.7 branch as it's not related with the next release and would be published right after merging.

AbrilRBS (approved):

PR #3841, opened 2024-09-16
Add lost space to fix inline formatting in whatsnew

Alexolut:
```
that is used by``mylib_b``
               ^
```
In case if it's not clearly visible on diff


AbrilRBS:
Hi @Alexolut thanks a lot for taking the time to create the PR, we really appreciate it :)

Could you target this PR for the `release/2.7` branch? That way we get the fix in the current version instead of waiting for the next release, thanks!

Alexolut:
> Could you target this PR for the `release/2.7` branch? 
Done

memsharded (approved):
Thanks!

PR #3843, opened 2024-09-16
Fix typo in cmake presets example code

i-ilak:
Running the code as described here:

```bash
$ conan new cmake_exe -d name=foo -d version=1.0
$ conan install .
$ conan install . -s build_type=Debug
$ cmake --preset conan-default
$ cmake --build --preset conan-debug
```
Will not work because the preset `conan-default` does not exist, but `conan-debug` exists.

### DIFF
--- a/examples/tools/cmake/cmake_toolchain/build_project_cmake_presets.rst
+++ b/examples/tools/cmake/cmake_toolchain/build_project_cmake_presets.rst
@@ -50,7 +50,7 @@ If you are using a multi-configuration generator:
 
 .. code:: bash
 
-    $ cmake --preset conan-default
+    $ cmake --preset conan-debug
     $ cmake --build --preset conan-debug
     $ build\Debug\foo.exe
     foo/1.0: Hello World Release!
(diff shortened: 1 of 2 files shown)


examples/tools/cmake/cmake_toolchain/build_project_cmake_presets.rst
```diff
@@ -50,7 +50,7 @@ If you are using a multi-configuration generator:
 
 .. code:: bash
 
-    $ cmake --preset conan-default
+    $ cmake --preset conan-debug
```
memsharded, line 53: Multi-configuration generators such as ``Visual Studio`` or ``Ninja Multi-Config`` will generate ``conan-default`` configure profile. It is only ``conan-debug`` if it is not a multi-config generator.

reference/conanfile/running_and_output.rst
```diff
@@ -53,6 +53,54 @@ To tag a warning, use the ``warn_tag`` argument of the ``warning()`` method in y
 
 .. _reference_conanfile_run:
 
+The common use of these functions from least specific to most specific, is:
```
memsharded, line 56: I think this was merged already, you might need to rebase your branch.

memsharded (changes requested):
I think there was some misundersanding, the instructions seems correct for multi-config generators

PR #3855, opened 2024-09-26
Note for --env-generation=false

AbrilRBS:
For https://github.com/conan-io/conan/pull/16935

### DIFF
--- a/reference/tools/env/virtualbuildenv.rst
+++ b/reference/tools/env/virtualbuildenv.rst
@@ -70,6 +70,8 @@ will create the following files:
 .. note::
 
     Disabling the automatic generation of these files for consumers can be achieved by passing ``--env-generation=false`` to the ``conan install`` command.
+    **This feature is experimental and subject to change**, and disabling this file can lead to unexpected build errors,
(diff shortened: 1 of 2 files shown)


reference/tools/env/virtualbuildenv.rst
```diff
@@ -67,6 +67,10 @@ will create the following files:
 
     To create ``.ps1`` files required for Powershell it is necessary to set to True the following conf: ``tools.env.virtualenv:powershell``.
 
+.. note::
+
+    Disabling the automatic generation of these files for consumers can be achieved by passing ``--env-generation=false`` to the ``conan install`` command.
```
memsharded, line 72: I'll add that suppressing these files means that some things won't work in some cases, and this should only be done if you are sure you don't need tool_requires and that stuff

I'd also add that this is experimental and might change

AbrilRBS: Good call about the experimental status

czoido (approved):

memsharded (approved):

PR #3862, opened 2024-10-04
fix preset name used

sdmg15:
After following the example I think it should be `conan-debug` here.


memsharded (changes requested):
Hi @sdmg15 

Thanks for your contribution.
I think it is already corret if you are using a multi-config generator (like Visual Studio in Windows), please read the line above, and the section below that uses a single-config generator (like makefiles in Linux)

sdmg15:
Thanks yes indeed :)

PR #3865, opened 2024-10-11
Wrong port order in partial docker example

Ibmurai:
The "inside" port goes on the ~left~ right.

### DIFF
--- a/reference/conan_server.rst
+++ b/reference/conan_server.rst
@@ -86,7 +86,7 @@ Server Parameters
   
     ..  code-block:: bash 
         
-        docker run ... -p9300:9999 ... # Check Docker docs for that
+        docker run ... -p9999:9300 ... # Check Docker docs for that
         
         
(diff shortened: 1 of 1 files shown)


reference/conan_server.rst
```diff
@@ -86,7 +86,7 @@ Server Parameters
   
     ..  code-block:: bash 
         
-        docker run ... -p9300:9999 ... # Check Docker docs for that
+        docker run ... -p9999:9300 ... # Check Docker docs for that
```
memsharded, line 89: But the ``conan_server`` is running on port 9300 by default, so the inside port should be 9300, isn't it? What am I missing?

Ibmurai: Left is "outside" right is "inside" :)

memsharded: that was it, the text was confusing, thanks for clarifying!

PR #3872, opened 2024-10-22
Clarify that you can set_property in CMakeDeps everything

AbrilRBS:


reference/tools/cmake/cmakedeps.rst
```diff
@@ -288,9 +288,7 @@ Overwrite properties from the consumer side using CMakeDeps.set_property()
 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
 Using ``CMakeDeps.set_property()`` method you can overwrite the property values set by the
-Conan recipes from the consumer. This can be done for `cmake_file_name`,
-`cmake_target_name`, `cmake_find_mode`, `cmake_module_file_name` and
-`cmake_module_target_name` properties. Let's see an example of how this works:
+Conan recipes from the consumer. This can be done for every property listed above.
```
jcar87, line 291: I have a vague recollection of a case for a property that didnt have this behaviour - let me double check

AbrilRBS: Note that in 2.8, `cmake_target_aliases` is _not_ overrideable,  but I'm opening a follow-up PR to fix that

PR #3880, opened 2024-10-28
Add custom generators thru tool_requires docs

AbrilRBS:
For https://github.com/conan-io/conan/pull/17129

### DIFF
--- a/reference/extensions/custom_generators.rst
+++ b/reference/extensions/custom_generators.rst
@@ -5,7 +5,7 @@ Custom Conan generators
 
 In the case that you need to use a build system or tool that is not supported by Conan
 off-the-shelf, you could create your own custom integrations using a custom generator.
-This can be done in two different ways.
+This can be done in three different ways.
 
 Custom generators as python_requires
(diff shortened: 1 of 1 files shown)


czoido (approved):

reference/extensions/custom_generators.rst
```diff
@@ -98,3 +98,47 @@ argument:
 .. code-block:: bash
 
     conan install --requires=zlib/1.2.13 -g MyGenerator
+
+
+Generators from tool_requires
```
memsharded, line 103: Maybe we also want to document that this can be used to inject built-in generators too?

AbrilRBS: Does the note in https://github.com/conan-io/docs/pull/3880/files#diff-4fe044d0a3df1239f28bea0a55f5277b3f0cc08f8f17e887fc9b69c07571f449R141 work?

memsharded: Yes, sounds good

PR #3899, opened 2024-11-17
fix core.cache examples

kambala-decapitator:
I copied example blindly and was getting

> ERROR: [conf] Either 'core.cache.storage_path' does not exist in configuration list or the conf format introduced is not valid. Run 'conan config list' to see all the available confs.


memsharded (approved):
Many thanks for fixing this!

memsharded:
Changed it to the release/2.9 branch, so it goes live immediately

PR #3901, opened 2024-11-19
[conanfile.txt] use recommended syntax for options

kambala-decapitator:


memsharded (approved):
Thanks for your contribution!

PR #3913, opened 2024-11-21
fix typo in replace_requires example

kambala-decapitator:


memsharded (commented):
Thanks for your PR @kambala-decapitator 

However, this one I think it is correct. The replacement is happening that specific version is replaced by a ``requires("zlib/[*]")``, that is, a version range, which is not the same as the placeholder to replace the version.

kambala-decapitator:
Could you explain the meaning of such construct via a use case (or possibly a real world example)? I don't think I understand what kind of replacement should occur and what would be the difference to a plain `*`.

Also, not sure I saw other examples of `[*]` in docs.

memsharded:
It is a version range, as defined in https://docs.conan.io/2/tutorial/versioning/version_ranges.html

There are some explicit examples in some points of the docs as: https://docs.conan.io/2/examples/tools/meson/create_your_first_package.html

In short, it is a version range that means "any", and as such it will resolve to the latest available version.

The ``replace_requires`` "*" is a place holder replacement: https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires

Means:
```
[replace_requires]
zlib/*: zlibng/*
```

It will replace an occurence of ``zlib/1.2.3`` for a new require to ``zlibng/1.2.3``.
If you put ``zlib/*: zlibng/[*]``  that means ``zlib/1.2.3`` would be replaced by ``zlibng/[*]`` which could resolve to any later version, such as ``zlibng/2.3``.

kambala-decapitator:
> If you put zlib/*: zlibng/[*] that means zlib/1.2.3 would be replaced by zlibng/[*] which could resolve to any later version, such as zlibng/2.3

aha, so this is the key meaning: to be able to ignore the range defined in the original package. Thanks a lot!

valgur:
> > If you put zlib/*: zlibng/[*] that means zlib/1.2.3 would be replaced by zlibng/[*] which could resolve to any later version, such as zlibng/2.3
> 
> 
> 
> aha, so this is the key meaning: to be able to ignore the range defined in the original package. Thanks a lot!

The docs could be a bit clearer here, though. The replacement applies only to the single specific version computed from that version range, not all versions that might match it. E.g. `zlib/[*]` will only match `zlib/1.2.3` and not `zlib/1.2.2` if both are available on the repo. I found this part a bit unexpected.

PR #3963, opened 2025-01-27
--build=compatible and other --build docs

memsharded:
Docs for https://github.com/conan-io/conan/pull/17637

### DIFF
--- a/reference/commands/install.rst
+++ b/reference/commands/install.rst
@@ -234,6 +234,8 @@ Note that the ``--update`` argument will look into all the remotes specified in 
 and won't stop at the first newer one found.
 
 
+.. _reference_commands_build_modes:
+
 Build modes
 -----------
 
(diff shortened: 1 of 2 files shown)


reference/commands/install.rst
```diff
@@ -232,3 +232,58 @@ to only update those packages, which avoids the re-evaluation of the whole graph
...
+.. code-block:: bash
+
+    --build=never      Disallow build for all packages, use binary packages or fail if a binary
+                       package is not found, it cannot be combined with other '--build' options.
+    --build=missing    Build packages from source whose binary package is not found.
+    --build=cascade    Build packages from source that have at least one dependency being built from
+                       source.
+    --build=[pattern]  Build packages from source whose package reference matches the pattern. The
+                       pattern uses 'fnmatch' style wildcards, so '--build="*"' will build everything
+                       from source.
```
AbrilRBS, line 256: Maybe we need to clarify (Or warn in the client?) the fact that `--build=zlib` _won't_ build zlib, as the pattern is checked against name/version

memsharded: Done.
Also added new section in ``conan create`` too, to clarify some behavior.

czoido (approved):

PR #3980, opened 2025-02-11
Consuming system requirements only when building a package

### LINKED ISSUE #3975: [system_requirements] How to consume system requirements only when building a package
Related to https://github.com/conan-io/conan/issues/16625 and #3974 

The Conan documentation could have those cases reported when system package managers should execute only when build the package, not when installing with pre-built binaries. 

The issue original issue has a basic explanation, but could be integrated to the official documentation, with a better view of the recipe example.

uilianries:
Added similar content as described on https://github.com/conan-io/conan/issues/16625#issuecomment-2624321525, but with more details.

closes #3975

Preview from commit [95b4c57](https://github.com/conan-io/docs/pull/3980/commits/95b4c57e6d7733e5080df954f934d50736f4f457)

Screenshot 2025-02-12 at 09-20-47 Consuming system requirements only when building a package — conan 2 12 1 documentation


examples/tools.rst
```diff
@@ -14,3 +14,4 @@ Conan recipe tools examples
    tools/scm/git/capture_scm/git_capture_scm
    tools/microsoft/msbuild
    tools/system/package_manager
+   tools/system/consuming_system_packages
```
memsharded, line 17: These need to be put into the same ``tools/system`` examples directory, not a first level entry for each one.

The entry point can summarize and redirect to the different examples below it

examples/tools/system/consuming_system_packages.rst
```diff
@@ -0,0 +1,111 @@
...
+Consuming system requirements only when building a package
+==========================================================
+
+In some cases, you may want to consume system requirements only when building a package, but not when installing it.
+It can be useful when you want to build a package in a CI/CD pipeline, but you don't want to run the system package
+manager when installing the Conan package in a different environment.
+For those cases, there are few approaches that can be used to achieve this goal.
+
+
+Consume a Conan package wrapper for a system package using hidden visibility
```
memsharded, line 12: Isn't this mostly what is discussed in the other example, with the system wrapper? Why repeating it here, not better linking to the other example?

examples/tools/system/consuming_system_packages.rst
```diff
@@ -0,0 +1,111 @@
...
+        ...
+
+This ensures that downstream consumers of the package *mypackage* will not directly invoke the system package manager (e.g., apt-get).
+Only the direct package consumer of the system wrap package for ``ncurses`` will execute the system package manager.
+
+Centralizing and wrapping ``ncurses`` in a single recipe, as mentioned in this example, makes it reusable across multiple
+cases and is good practice to avoid code duplication.
+
+
+Package the system requirement as a header-library or consume only the headers
```
memsharded, line 40: This also reads like a derivative or particular case for the other example.

memsharded (commented):
I am a bit confused about this, it seems it has some overlap with the other existing example with code, not as clear as the other one, like a bit of a mix.

examples/tools.rst
uilianries: Do you mean only extending the file `package_manager.rst` with this new content, instead of adding an entire new?

This new example/how-to is under the same examples tools/system folder as the other example.

examples/tools/system/consuming_system_packages.rst
uilianries: Yes, it's a particular case related to the previous example.

examples/tools.rst
memsharded: I mean, every other item in this section is a sub-section, see https://docs.conan.io/2/examples/tools.html, containing further articles/examples. The missing part is the equivalent to https://docs.conan.io/2/examples/tools/google/bazel.html, but for "system-package-manager"

examples/tools/system/consuming_system_packages.rst
memsharded: That is exactly my point. We have a full working and reproducible example in https://docs.conan.io/2/examples/tools/system/package_manager.html. Why adding a new partial ``conanfile.py`` here, just for the ``visible=True``? I think the best would be to provide the real world case:
- Provide a full ``pkg`` conanfile.
- Then use the ``tool_requires()`` instead of ``visible=False``, it is more idiomatic and adequate for the task.
- ``conan create`` the package
- Consume the package, show the output to see that the ``ncurses/system`` is skipped, and that consequently the ``Apt()`` or equivalent is not called.

examples/tools/system/consuming_system_packages.rst
memsharded: I think this section is more confusing than beneficial at this point, I'd probably remove it.

examples/tools.rst
uilianries: Restructured on the commit [6f50b49](https://github.com/conan-io/docs/pull/3980/commits/6f50b49f3f99828bb2dcf2e3269a51958c1ac25a)

examples/tools/system/consuming_system_packages.rst
uilianries: Removed by the commit [95b4c57](https://github.com/conan-io/docs/pull/3980/commits/95b4c57e6d7733e5080df954f934d50736f4f457)

examples/tools/system/consuming_system_packages.rst
uilianries: Simplified by the commit [95b4c57](https://github.com/conan-io/docs/pull/3980/commits/95b4c57e6d7733e5080df954f934d50736f4f457)

memsharded (approved):

czoido (approved):

PR #4005, opened 2025-03-01
add note about `languages = "C"` since v2.4

kambala-decapitator:
copied the note from `configure()` reference


memsharded (approved):
Thanks for your contribution!

PR #4006, opened 2025-03-05
fixing .gitattributes

memsharded:


AbrilRBS (approved):
From https://git-scm.com/docs/gitattributes#_end_of_line_conversion:

Captura de pantalla 2025-03-05 a las 13 06 52

memsharded:
> From https://git-scm.com/docs/gitattributes#_end_of_line_conversion:

Yes, but that is not fully clear to me. Normalization is for entering the git database, but the eol is for the checkout.

PR #4009, opened 2025-03-06
Correct type-o in documentation

KUGA2:
This one is quite important to fix, because one might also think that `ai` was meant here.


memsharded (approved):
Thanks for the fix!

Issue #4014, opened 2025-03-10
Open an integration section for third parties


memsharded:
- Rust-Conan integration
- Conan GUI
- Gradle C++ plugin

PR #4019, opened 2025-03-14
new integrations-community section

memsharded:
Close https://github.com/conan-io/docs/issues/4014

cc @awakecoding @markfinal

### DIFF
--- a/integrations/community.rst
+++ b/integrations/community.rst
@@ -1,0 +1,28 @@
+.. _integrations_community:
+
+Community
+=========
+
+
+.. important::
+
+    This section contains links to different integrations created and supported by the community.
+    Conan is not responsible for any of them, and does not provide any guarantee of compatibility or support.
+
+- **conan-rs Rust-Conan integration**: A Rust wrapper of the conan C/C++ package manager (conan.io) to simplify usage in build scripts.
+  For source and support go to the `conan-rs Github page <https://github.com/Devolutions/conan-rs>`_. 
+  Created by `@awakecoding <https://github.com/awakecoding>`_
+- **Cruiz Conan GUI**. A Graphical User Interface for Conan client from `Mark Final <https://github.com/markfinal>`_. 
+  Source and support in `cruiz Github page <https://github.com/markfinal/cruiz>`_.
+- **Gradle Conan plugin**: A plugin to use Conan from the Gradle build system. See it in the `Gradle plugins site <https://plugins.gradle.org/plugin/net.sf.loggersoft.cpp-conan>`_.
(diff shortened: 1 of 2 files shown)


integrations/community.rst
```diff
@@ -0,0 +1,28 @@
...
+  For source, suppport (tickets), go to `project Sourceforge page <https://sourceforge.net/projects/gradle-cpp/>`_
+
+
+.. note::
+
+    **Notes**
+
+    - If you have an integration that you think could be valuable for the community, and you are willing to support it,
+      feel free to open a Pull Request to this page in the Github repo.
+    - Recall that these integrations are not supported by Conan team. If you have any question, issue or feedback about
```
AbrilRBS, line 27:
```suggestion
    - Recall that these integrations are not supported by the Conan team. If you have any questions, issues or feedback about 
```

AbrilRBS (approved):

awakecoding:
thanks for including me, but it would probably be better to remove conan-rs from the list and include conan2-rs from @ravenexp instead: https://github.com/ravenexp/conan2-rs

conan-rs is for conan 1.x only, and since we're no longer using conan from Rust, we have never ported it to conan 2. conan2-rs forked it and has someone maintaining it for conan 2.x, so it would be better to point new users to it instead.

memsharded:
> thanks for including me, but it would probably be better to remove conan-rs from the list and include conan2-rs from @ravenexp instead: https://github.com/ravenexp/conan2-rs

Thanks very much for pointing that out, I knew it was a good idea to ping you in this PR :)

ravenexp:
Thanks for mentioning my project. `conan2-rs` is not a fork in the classical sense, but a greenfield implementation for Conan 2.0 only, even though it was largely inspired by `conan-rs` (thanks to @awakecoding).

It has never supported Conan 1.x, nor does it provide any migration path for current `conan-rs` users.
I'm currently using it for my mixed Rust + C++ projects, so it is passively maintained ATM.

memsharded:
Thanks for info @ravenexp and for releasing that integration as open source. It is perfectly fine if it doesn't support Conan 1 nor have an upgrade path.
I have updated the PR to link to your repo instead, feel free to comment or suggest changes.

markfinal:
Thanks also for the mention. Time has not been kind on finding free moments to do all the things I want to do in cruiz. I'll get there eventually. But I know people who use it most days, so I'm happy it's being put to use.

memsharded:
> Thanks also for the mention. Time has not been kind on finding free moments to do all the things I want to do in cruiz. I'll get there eventually. But I know people who use it most days, so I'm happy it's being put to use.

Excellent, thanks for the feedback and for sharing this integration Mark!

czoido (approved):

PR #4022, opened 2025-03-16
use --out-file param instead of redirecting output

kambala-decapitator:
TMU this is a more correct way now


memsharded (commented):
Thanks for the contribution @kambala-decapitator !

But not really, we still think the stdout redirection is a more idiomatic way than a custom argument. The ``--out-file`` was added for possible corner cases, but the redirection is still good, it is massively used in the docs, in our trainings, demos, conference talks, etc, and at the moment we are not planning to change it for ``--out-file``.

kambala-decapitator:
Thanks a lot for the explanation!

PR #4033, opened 2025-03-23
Fix incorrect statement at export() method

aagor:
* The text described export_sources(), not export().
* The equivalent attributes exports and exports_sources have correct documentation.


memsharded (approved):
Thanks for fixing this @aagor !

memsharded:
Merged, will be included in next 2.15 release

PR #4034, opened 2025-03-23
Add cross reference to export_sources() method

aagor:
* Similar to cross reference at exports attribute.


memsharded (approved):
Thanks @aagor !!

PR #4037, opened 2025-03-24
Mention `ConanFile.conf` validity

tttapa:
This addresses comments from @memsharded in https://github.com/conan-io/conan/issues/18008 about which methods are allowed to rely on the data in `ConanFile.conf`.

I just added a quick note, but it would be great to discuss some more helpful suggestions for people who need these configuration values elsewhere in their recipe.

### DIFF
--- a/reference/conanfile/attributes.rst
+++ b/reference/conanfile/attributes.rst
@@ -87,6 +87,10 @@ The profile entries have priority.
     The ``conf`` attribute is a **read-only** attribute. It can only be defined in profiles and command lines, but it should never be set by recipes.
     Recipes can only read its value via ``self.conf.get()`` method.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


AbrilRBS:
I think it would be easier to note in which methods access to `conf` is not allowed, or only a certain kind of confs are allowed to be queried, as the current list would be missing quite a few methods, let me check with the team :)

memsharded (commented):
This might be a bit more complicated than this.
The issue is that the ``conf`` is available in other places, as long as it comes from profiles files.

It is the usage of ``conf`` coming from ``tool_requires`` that is expected to be limited, because those ``tool_requires`` might not be necessarily available (skipped), and their ``package_info()`` won't even execute.

PR #4038, opened 2025-03-28
[reference] Add tools.files.chmod description and reference

uilianries:
Related to https://github.com/conan-io/conan/pull/17800

It requires Conan develop2 branch to be generated.

Preview - Docs commit [2985991](https://github.com/conan-io/docs/pull/4038/commits/29859916a1c613cbd96f405d8d16dd65ec1e9c06) + Conan commit [63e193b](https://github.com/conan-io/conan/pull/18038/commits/63e193bd8b4e2f592f36f7738583513d16523541)

conan_ref_files_chmod

### DIFF
--- a/reference/tools/files/basic.rst
+++ b/reference/tools/files/basic.rst
@@ -103,6 +103,27 @@ Usage:
     replace_in_file(self, os.path.join(self.source_folder, "folder", "file.txt"), "foo", "bar")
 
 
+conan.tools.files.chmod()
+-----------------------------------
+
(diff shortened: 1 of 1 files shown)


czoido:
If we accept None as value for read, write and execute, shouldn't the type hints be as follows?

```
read: Optional[bool] = None,
write: Optional[bool] = None,
execute: Optional[bool] = None,
```

Also, I think that it should be documented in the docstring that `True` will add the permission, `False` will remove it and the default that is `None` does not modify it.

czoido (approved):

PR #4039, opened 2025-03-29
Various minor tweaks and fixes

chausner:
This is a collection of small grammar tweaks and fixes.

PR #4047, opened 2025-04-01
Various minor tweaks and fixes (continued)

chausner:
Second batch of minor grammar tweaks and fixes, continued from #4039.


memsharded (approved):
Thanks very much for contribution such a thorough and detailed review @chausner !

PR #4056, opened 2025-04-11
Update settings.rst for apple-clang 17

daniel-heater-imprivata:
I updated XCode yesterday and now I have:
```
❯ clang --version
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.3.0
```

This adds support for apple-clang 17


memsharded (commented):
Thanks for contributing this @daniel-heater-imprivata !

It would be great to explicitly verify that this new version hasn't change the C++ standard default, as there are some helper functions that would require this information if this is the case. Have you checked this? A link would be appreciated.

AbrilRBS:
Thanks for the PR @daniel-heater-imprivata

Note that the documentation usually gets updated for the next release shortly before it's due, that's why this was still missing from the `develop2` branch here.

Note that as per my initial comment in https://github.com/conan-io/conan/pull/18051, the default C++ std **has** changed - but I'm not quite sure what the necessary changes for that are in the docs, I couldn't find anything

cc @memsharded

czoido:
Hi @daniel-heater-imprivata,

Thanks a lot for the contribution, we have decided to release a patch version 2.15.1 adding the new versions, this is the PR to add those to the release https://github.com/conan-io/conan/pull/18122

Let me close this PR, and I'll add a new one branching from the correct branch to add also other changes related with the release.

PR #4091, opened 2025-05-14
[FAQ] fix .gitattributes code block

kambala-decapitator:
maybe it'd also be better to narrow the example?

    conanfile.* text eol=lf

### DIFF
--- a/knowledge/faq.rst
+++ b/knowledge/faq.rst
@@ -122,7 +122,7 @@ Conan will not normalize or change in any way the source files, it is not its re
 
 .. code-block:: ini
 
-* text eol=lf
+    text eol=lf
 
 
 Other approach would be to change the ``.gitconfig`` to change it globally. Modern editors (even Notepad) in Windows can perfectly work with files with ``LF``, it is no longer necessary to change the line endings.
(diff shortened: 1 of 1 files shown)


memsharded:
> conanfile.* text eol=lf

I am afraid not enough.
Many recipes often have other extra files, like patches, an extra ``CMakeLists.txt``, some ``mydeps.json`` file, or even other imported ``mycode.py`` files. All of them are text based files that cannot be checked-out with CRLF on Windows, as that would also break the revision hash

knowledge/faq.rst
```diff
@@ -122,7 +122,7 @@ Conan will not normalize or change in any way the source files, it is not its re
 
 .. code-block:: ini
 
-* text eol=lf
```
memsharded, line 125: Why exactly? Isn't the ``*`` pattern working correctly, or is it just to simplify?
Thanks!

kambala-decapitator: this is how it looks at the moment:

Screenshot 2025-05-14 at 14 36 48

kambala-decapitator: oh, I thought that `*` was a mistakenly used list element. It should be a wildcard pattern, right?

PR #4102, opened 2025-05-23
Propagate master layout to all branches to enable cookies

czoido:


_templates/layout.html
```diff
@@ -1,6 +1,8 @@
 {% extends "!layout.html" %}
 
 {%- block extrahead %}
+
+{% if enable_cookies|default(true) %}
```
czoido, line 5: this comes from here: https://github.com/conan-io/docs/pull/3856 but with the filter it should work even if we don't define the variable in the conf.py

franramirez688 (approved):

PR #4115, opened 2025-06-04
Added Emscripten documentation

perseoGI:
Examples2 PR: https://github.com/conan-io/examples2/pull/184
Client PR: https://github.com/conan-io/conan/pull/18400


examples/cross_build/emscripten.rst
```diff
@@ -0,0 +1,256 @@
...
+
+.. code-block:: text
+
+   include(default)
+   [settings]
+   arch=asm.js
+   build_type=Release
+   os=Emscripten
+
+   [tool_requires]
```
memsharded, line 34: Maybe in the basic example would be better not to depend on a ConanCenter package such as ``emsdk``, but use the one installed by the user?. Then a second example can use the one from the package. 

Same use case as AndroidNDK in https://docs.conan.io/2/examples/cross_build/android/ndk.html, we can assume that devs building for webassembly will have their own emscripten installed in their machines and want to use it?

examples/cross_build/emscripten.rst
```diff
@@ -0,0 +1,349 @@
...
+
+Build the project for `asm.js`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Let's build the project for both targets, first for ``asm.js``:
+
+
+..  code-block:: bash 
+    
+   $ conan build . -profile:host ../profiles/wasm32 --build=missing
```
czoido, line 246: I think the profile here should be `asmjs`?

czoido (approved):

examples/cross_build/emscripten.rst
```diff
@@ -0,0 +1,139 @@
...
+
+    - **asm.js** is a subset of JavaScript optimized for speed. It is fully supported by all browsers (even older ones) and compiles to a large ``.js`` file.
+    - **WebAssembly (WASM)** is a binary format that is smaller and faster to load and execute. Most modern browsers support it, and it is generally recommended for new projects. **WASM** is also easier to integrate with native browser APIs compared to **asm.js**.
+
+
+Even though Emscripten is not a true runtime environment (like Linux or
+Windows), it is part of a toolchain ecosystem that compiles C/C++ to
+WebAssembly (WASM) and asm.js.
+
+Conan uses Emscripten to:
```
memsharded, line 54:
```suggestion
Conan uses ``os=Emscripten`` to:
```

memsharded (approved):

PR #4127, opened 2025-06-16
new conan report diff documentation

ErniGH:
related to: https://github.com/conan-io/conan/pull/18247


reference/commands/report/diff.rst
```diff
@@ -0,0 +1,77 @@
...
+
+When only a reference is given, Conan will first search for the recipe in the local cache; if it is not found, it will
+attempt to download it from the configured remotes. If no revision is explicitly provided, Conan will
+default to using the latest available revision.
+
+
+**Examples**
+
+
+If we want to compare versions 1.0 and 2.0 of "mylibrary" that are available on our remotes, it would be:
```
AbrilRBS, line 22:
```suggestion
If we want to compare versions 1.0 and 2.0 of "mylibrary" that are available on our `my-remote` remote, it would be:
```

AbrilRBS (approved):

reference/commands/report/diff.rst
```diff
@@ -0,0 +1,77 @@
...
+
+
+This is how it would look if one of the versions is on the remote and the other is not. The version that is not on the
+remote requires the path to the recipe in order to compare it.
+
+.. code-block:: bash
+
+    $ conan report diff -or="mylib/1.0" --old-path="path/to/recipe" -nr="mylib/2.0"
+    $ conan report diff -or="mylib/1.0" -nr="mylib/2.0" --new-path="path/to/recipe"
+    $ conan report diff -or="mylib/1.0" --old-path="path/to/recipe" -nr="mylib/2.0" --new-path="path/to/recipe"
```
czoido, line 36: this is a bit confusing, and also mixing abbreviated arguments with the ones that are not, can you elaborate a bit more on the previous paragraphs about each of the arguments and how they should be used please?

reference/commands/report.rst
```diff
@@ -0,0 +1,10 @@
+conan report
+============
+
+
+The ``conan report`` command contains subcommands that return information about packages and libraries.
+
+.. toctree::
+   :maxdepth: 1
+
+   conan report diff: Get the differences between two recipes, also comparing their sources. <report/diff>
```
czoido, line 10: This is incomplete and does not follow how we document the reference for other commands, please check any other command to complete this.

reference/commands/report/diff.rst
```diff
@@ -0,0 +1,90 @@
...
+
+When only a reference is given, Conan will first search for the recipe in the local cache; if it is not found, it will
+attempt to download it from the configured remotes. If no revision is explicitly provided, Conan will
+default to using the latest available revision.
+
+
+**Examples**
+
+
+If we want to compare versions 1.0 and 2.0 of "mylibrary" that are available on our `my-remote` remote, it would be:
```
czoido, line 22:
```suggestion
If we want to compare versions 1.0 and 2.0 of `mylib` that are available on our `my-remote` remote, it would be:
```

reference/commands/report/diff.rst
```diff
@@ -0,0 +1,90 @@
...
+This functionality allows you to compare either two versions of the same recipe or two entirely different recipes.
+Each recipe (old and new) can be identified in one of two ways: by providing both the path to its ``conanfile.py`` and
+its reference, or by specifying just the reference.
+
+When only a reference is given, Conan will first search for the recipe in the local cache; if it is not found, it will
+attempt to download it from the configured remotes. If no revision is explicitly provided, Conan will
+default to using the latest available revision.
+
+
+**Examples**
```
czoido, line 19: I find this section still a bit confusing, a couple of suggestions:

To improve clarity, would it be possible to structure the examples to show one dedicated example for each of the main use cases? I think this would make the purpose of the `--...-reference` and `--...-path` arguments much easier to understand at a glance.

For example:

* `ref` vs `ref`
* `local path` vs `ref`
* `ref` vs `local path`
* `local path` vs `local path`

Also, for consistency, sticking to either the long or short form of the arguments within each example might help readability.

PR #4144, opened 2025-06-26
fix: correct misuse of parentheses in shutil.copy example

R1cardoX:


R1cardoX:
Hello, this PR is ready to merge. Please take a look when convenient. Thanks!

PR #4159, opened 2025-07-08
Fix and clarify description of self.folders.subproject

aagor:
The old documentation was wrong, as the subproject is not required to be the folder, where the `conanfile.py` is located.
An example, where this is not the case, can be found here: https://docs.conan.io/2/examples/conanfile/layout/conanfile_in_subfolder.html

### DIFF
--- a/reference/conanfile/methods/layout.rst
+++ b/reference/conanfile/methods/layout.rst
@@ -39,8 +39,9 @@ self.folders
   **self.folders.root**.
 
 - **self.folders.subproject** (Defaulted to ``None``): Specifies a subfolder where the
-  ``conanfile.py`` is relative to the project root. This is particularly useful for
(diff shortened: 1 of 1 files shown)


memsharded (commented):
While this is technically correct, I am not sure this was the original intention and design of the feature. The ``self.folders.subproject`` intends to have the ``conanfile.py`` in that location, and then the rest of the relative things of the subproject relative to it. 
That is, the idea is that the ``conanfile.py`` is found at the root of the subproject. Maybe it doesn't work exactly like that for all cases, but for example, the docs in https://docs.conan.io/2/examples/conanfile/layout/multiple_subprojects.html show this usage.

It looks like we probably want to investigate the implications of this, if there could be some problems of explicitly documenting and encouraging this layout.

PR #4165, opened 2025-07-11
Fix architecture mapping at Apt documentation

aagor:
Just copy the mapping from [`package_manager.py`](https://github.com/conan-io/conan/blob/fdc7bfa41a183ebe015ea03460077959ab62ae16/conan/tools/system/package_manager.py#L235-L242).


memsharded (approved):
Thanks! 
It seems the dict that was copied was the YUM one, not the APT one.

PR #4190, opened 2025-08-14
Add note about bash:active=True and MSYS2

czoido:
Related with:
- https://github.com/conan-io/conan-training2/issues/21
- https://github.com/conan-io/conan/issues/18734

### DIFF
--- a/examples/tools/autotools/create_your_first_package_windows.rst
+++ b/examples/tools/autotools/create_your_first_package_windows.rst
@@ -57,8 +57,11 @@ execute the build in it.
 
   It is not necessary, and in fact it is not recommended for most cases to be already running inside an ``msys2``
   terminal. Conan will automatically run the build subprocess for autotools in the defined bash shell.
-
-  If already running in a bash shell, it is necessary to activate the ``tools.microsoft.bash:active=True`` conf.
+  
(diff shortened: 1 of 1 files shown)


examples/tools/autotools/create_your_first_package_windows.rst
```diff
@@ -57,8 +57,11 @@ execute the build in it.
 
   It is not necessary, and in fact it is not recommended for most cases to be already running inside an ``msys2``
   terminal. Conan will automatically run the build subprocess for autotools in the defined bash shell.
-
-  If already running in a bash shell, it is necessary to activate the ``tools.microsoft.bash:active=True`` conf.
+  
+  If already running in a bash shell, it is necessary to activate the
```
memsharded, line 61: Not sure this is the best place for this clarification, this example is not very visible.
Maybe add it to the ``global.conf`` description itself?

Then, it might not be fully clear, maybe turn it around: "if running in a msys2 subsystem that uses a /bin/sh or bash shell as default subprocess shell, it is necessary to activate ``active=True`` ..."

czoido: Yes, I agree, the description in the source code would be a better place, I'll do a PR in the Conan client.

PR #4205, opened 2025-08-29
Add runtime_deploy entry

uilianries:
Related to https://github.com/conan-io/conan/pull/17848

- Add a new entry for `runtime_deploy` in the deployers page.
- Update install reference about preserving subfolders.
- Moved the warning message from deployers to the top, so it will be more visible.

### DIFF
--- a/reference/commands/install.rst
+++ b/reference/commands/install.rst
@@ -130,7 +130,8 @@ There are 3 built-in deployers:
   dependencies.
 - ``runtime_deploy`` deploys all the shared libraries and the executables of the
   dependencies (like ``.so``, ``.dll``, or ``.dylib`` files) into a flat directory
-  structure. The symbolic links of libraries are copied by default and they can be managed through the boolean
+  structure, preserving subdirectories as-is.
+  The symbolic links of libraries are copied by default and they can be managed through the boolean
   configuration ``tools.deployer:symlinks``. (Available since Conan 2.5.0)
 
 Some generators might have the capability of redefining the target "package folder". That means that if some other
--- a/reference/extensions/deployers.rst
+++ b/reference/extensions/deployers.rst
@@ -22,6 +22,12 @@ Use the ``--deployer-folder`` argument to change the base folder output path for
 Built-in deployers
 ------------------
 
+.. warning::
+
+  The built-in deployers are in **preview**.
+  See :ref:`the Conan stability<stability>` section for more information.
+
+
 .. _reference_extensions_deployer_full_deploy:
 
 full_deploy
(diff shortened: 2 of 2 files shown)


reference/extensions/deployers.rst
```diff
@@ -50,18 +56,27 @@ This deployer will output your dependencies in a tree folder such as:
...
+and ``cpp_info.libdirs`` configuration.
 
 
 configuration
 ^^^^^^^^^^^^^
 
-Both the ``full_deploy`` and the ``direct_deploy`` understand when the conf ``tools.deployer:symlinks`` is set to ``False``
-to disable deployers copying symlinks. This can be convenient in systems that do not support symlinks and could fail
-if deploying packages that contain symlinks.
+Both the ``full_deploy``, ``direct_deploy`` and ``runtime_deploy`` understand when the conf ``tools.deployer:symlinks``
```
uilianries, line 77: Only added `runtime_deploy` here and reformatted the line. Otherwise, it passes +120 characters.

reference/commands/install.rst
```diff
@@ -130,7 +130,8 @@ There are 3 built-in deployers:
   dependencies.
 - ``runtime_deploy`` deploys all the shared libraries and the executables of the
   dependencies (like ``.so``, ``.dll``, or ``.dylib`` files) into a flat directory
-  structure. The symbolic links of libraries are copied by default and they can be managed through the boolean
+  structure, preserving subdirectories as-is.
```
czoido, line 133: I think it would be great to link the reference for each deployer here

reference/extensions/deployers.rst
czoido, line 77:
```suggestion
The ``full_deploy``, ``direct_deploy`` and ``runtime_deploy`` understand when the conf ``tools.deployer:symlinks``
```

reference/commands/install.rst
uilianries: @czoido I reduced the install to only point to deployers instead: [835e2b9](https://github.com/conan-io/docs/pull/4205/commits/835e2b9dbd92d06e5c3d5be693f6301722aba58b)

I believe the information is duplicated so far, we could centralize all them in the deployers page only. WDYT?

czoido: For me, having a quick summary of what the deployer does in the install reference was ok, then you can get more details if you want... but I'm fine with this too.

uilianries: Done, the summary is back: [36e5629](https://github.com/conan-io/docs/pull/4205/commits/36e56297f28174f6e69c44d350f5ffaf7a3fdf82)

I reduced a little the runtime_deploy explanation, as we have a more complete information on the deploy page.

czoido (approved):

PR #4216, opened 2025-09-03
fix version-ranges caret

memsharded:
Close https://github.com/conan-io/docs/issues/4215

PR #4218, opened 2025-09-03
Clarify docs for version-range carat operator

dskyle-shieldai:
The docs say that variability is allowed in the "last defined digit", which suggests that "^1.0.0" should include only versions of the form "1.0.X", but actually it accepts "1.X.X". The actual logic doesn't have anything to do with the "last defined" digit of any kind. Instead, it looks for the first non-zero digit, and allows the following digit to vary.


dskyle-shieldai:
Related PR here: https://github.com/conan-io/docs/pull/4216

Issue #4231, opened 2025-09-18
[question] package_type trait inferring clarification


andrey-zherikov:
[This section](https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring) of documentation describes how `package_type` affects dependency trait. Could you please clarify (ideally update the docs) whether it's about `package_type` of depender or dependee?

Does `package_type` of dependency affect traits? For example, dependency on `build-scripts` should infer `build=True`.

AbrilRBS:
No worries.

Yes, and this is something that could also be improved in the documentation:

Unless the list you linked above in the documentation sets them to `False` (Or the user does manually), the `headers`, `libs` and `visible` traits are set to `True` by default for every kind of requirements

andrey-zherikov:
I believe this should be explicitly mentioned in the documentation to avoid guessing and misunderstanding. For example, in the list where `package_type` of dependency is `*-library`.

AbrilRBS:
@andrey-zherikov I've now proposed https://github.com/conan-io/docs/pull/4234, if you could take a look and see if that solves your questions, it would be great to have your feedback, thanks!

PR #4234, opened 2025-09-20
More info on default traits of dependencies

AbrilRBS:
For https://github.com/conan-io/docs/issues/4231

### DIFF
--- a/reference/conanfile/methods/requirements.rst
+++ b/reference/conanfile/methods/requirements.rst
@@ -229,6 +229,13 @@ package_type trait inferring
 Some traits are automatically inferred based on the value of the ``package_type`` of the dependency
 if not explicitly set by the recipe.
 
+.. note::
+
+    The ``libs``, ``headers`` and ``visible`` traits are set to ``True`` by default
+    unless otherwise stated in the lists shown below, or if they are manually set by the user to ``False``.
+
+The inferring rules are:
+
  * ``application``: ``headers=False``, ``libs=False``, ``run=True``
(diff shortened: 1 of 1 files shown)


andrey-zherikov (commented):
This looks much clearer.  Thank you!

reference/conanfile/methods/requirements.rst
```diff
@@ -257,3 +264,12 @@ Each kind of requires sets some additional traits by default on top of the ones
...
  * ``test_requires``: ``headers=True``, ``libs=True``, ``build=False``, ``visible=False``, ``test=True``
+
+For example, taking all the logic shown into account, this means that if you have
+``self.requires("mypkg/1.0", visible=False)`` and ``mypkg/1.0`` has ``package_type="shared-library"``,
+the effective traits for that ``requires`` will be:
+
+ * Inferred from ``mypkg``'s ``package_type``: ``run=True``
+ * By the ``requires`` kind: ``build=False``
+ * Overridden by the user: ``visible=False``
+ * By default: ``headers=True``, ``libs=True``
```
memsharded, line 275: Note that both ``run=True`` and ``visible=False`` are kind of incompatible. Unless the shared-library isn't really used by the consumer, for example, lets say that the consumer is good with just a headers with some templates that don't have any linkage needs to the shared library, this will be problematic, as the shared-library runtime requirements will not be propagated downstream, and it will fail to run.

memsharded (commented):
I have some concern about the example

PR #4238, opened 2025-09-22
Docs for recipes_only

AbrilRBS:
For https://github.com/conan-io/conan/pull/18896

### DIFF
--- a/devops/using_conancenter.rst
+++ b/devops/using_conancenter.rst
@@ -105,8 +105,11 @@ we recommend building your own binary packages from source:
 
 If the ``conancenter`` remote is used directly, your organization might require
 that the binaries are built from source, and not downloaded directly from ConanCenter.
-This can be achieved with the ``recipes_only`` field in the remote configuration,
-see :ref:`the remotes.json docs<reference_config_files_remotes_json>`.
+This can exceptionally be achieved with the ``recipes_only`` field
+:ref:`in the remote configuration<reference_config_files_remotes_json>`, but it is
+highly likely that you will also want to build the packages from source, and not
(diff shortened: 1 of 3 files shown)


security/guidelines.rst
```diff
@@ -20,5 +20,7 @@ This is an incomplete and preliminary, not exhaustive security related recommend
   party dependencies. For this reason, the :ref:`Using ConanCenter packages in production environments <devops_consuming_conan_center>` section recommends building your own binaries
   from source and storing those binaries in your own private server, without downloading packages directly from ConanCenter.
   The :ref:`local-recipes-index feature<devops_local_recipes_index>` was designed to help in this process.
+- The ``recipes_only`` field for remotes can be used to avoid downloading binaries from untrusted remotes,
```
memsharded, line 23: Downloading recipes from untrusted remotes is a high security risk. I am not sure I'd put this here as a security guideline, it is more like a policy/regulation thing, but you cannot and you never should download even recipes only from a remote that is untrusted

devops/using_conancenter.rst
```diff
@@ -103,6 +103,11 @@ we recommend building your own binary packages from source:
 
 - Read more: :ref:`creating and hosting your own Conan Center binaries <devops_conancenter_hosting_your_own_conancenter_fork>`
 
+If the ``conancenter`` remote is used directly, your organization might require
```
memsharded, line 106: I'd add a clarification here, that this should be done for exceptional cases, but if you don't want to use binaries from ConanCenter, it is very likely that you also want to build the packages directly from source, and not depend on the recipes on ConanCenter, but from ``conan-center-index`` github repo

PR #4248, opened 2025-09-26
Some minor changes

AbrilRBS:


_static/css/conan.css
```diff
@@ -25,3 +25,7 @@ h1, h2, .rst-content .toctree-wrapper p.caption, h3, h4, h5, h6, legend {
     text-align: left;
     font-weight: bold;
 }
+
+/*.wy-nav-content {*/
```
AbrilRBS, line 29: Here to discuss if this would make sense

_themes/conan_theme/versions.html
```diff
@@ -13,11 +13,14 @@
         {% endif %}
     {% endfor %}
     <br>
+        <details>
```
czoido, line 16: I think changes to the template should go to master branch, because the template is taken from there if I recall correctly?

PR #4249, opened 2025-09-26
Fixed some typos and little tweaks

perseoGI:
While reviewing docs for the conan training, I found some things that can be improved.

### DIFF
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -285,8 +285,8 @@ the prefix of the variable. It is useful to automatically get the append/prepend
     # Prepend another PATH to "MyPath1"
     MyPath1=+(path)/other path/path12
 
-    # Unset the variable "MyPath1"
-    mypkg/*:MyPath1=!
+    # Unset the variable "MyPath2"
+    MyPath2=!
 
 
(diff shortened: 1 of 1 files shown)


reference/config_files/profiles.rst
```diff
@@ -286,7 +286,7 @@ the prefix of the variable. It is useful to automatically get the append/prepend
     MyPath1=+(path)/other path/path12
 
     # Unset the variable "MyPath1"
-    MyPath1=!
+    mypkg/*:MyPath1=!
```
memsharded, line 289: This would be introducing here the concept of per-package environment, but this section is to introduce the core syntax (add, append, remove).

It would be better to leave it as-is, and then add the per-package examples with ``mypkg/*`` separately

perseoGI: I agree with you. There were two options, modifying this line or the below line. 
See this screen capture from https://docs.conan.io/2/reference/config_files/profiles.html#profiles

I was confused about reading `MyPath=!` and then the explanation below.


ImageTT

AbrilRBS (approved):

memsharded (approved):

PR #4250, opened 2025-09-26
multicmake info added

davidsanfal:
@memsharded I don't know if this is enough info or if it's clear, but I couldn't think of a way to explain it in a few lines without ending up giving a complete example of how to use it.

### DIFF
--- a/reference/tools/cmake/cmake.rst
+++ b/reference/tools/cmake/cmake.rst
@@ -35,7 +35,10 @@ when a package is being built directly by Conan (create, install)
             cmake.configure()
             cmake.build()
 
-
+``CMake`` ``configure()``, ``build()`` and ``install()`` methods have the ``subfolder`` parameter in case you need to generate the files
+and artifacts in a subfolder inside the ``build_folder`` or ``package_folder``.
+This allows you to have more than one ``CmakeList.txt`` in the same project without mixing the generated files
+or allows you to do a ``cmake install`` only of the components you really want to package.
 
 Reference
 ---------


reference/tools/cmake/cmake.rst
```diff
@@ -35,7 +35,10 @@ when a package is being built directly by Conan (create, install)
             cmake.configure()
             cmake.build()
 
-
+``CMake`` ``configure()``, ``build()`` and ``install()`` methods have the ``subfolder`` parameter in case you need to generate the files
```
memsharded, line 38: I am not sure this is the intended use case?
The use case was having more than 1 CMakeLists.txt, in separate places, not as a single project?
I was thinking more an example like:
```python
def build(self):
     cmake1 = CMake(.., subfolder="folder1")
     cmake1.configure()
     cmake1.build()
     cmake2 = CMake(..., subfolder="folder2")
     cmake2.configure()
     cmake.build()
```

memsharded (approved):

reference/tools/cmake/cmake.rst
```diff
@@ -35,7 +35,27 @@ when a package is being built directly by Conan (create, install)
             cmake.configure()
             cmake.build()
 
+``CMake`` ``configure()``, ``build()`` and ``install()`` methods have the ``subfolder`` parameter in case you have
+more than one ``CmakeLists.txt`` in different folders.
```
czoido, line 39:
```suggestion
more than one ``CMakeLists.txt`` in different folders.
```

reference/tools/cmake/cmake.rst
```diff
@@ -35,7 +35,27 @@ when a package is being built directly by Conan (create, install)
...
+This feature allows you to call the ``configure`` ``build`` and ``install`` method of each ``CmakeLists.txt``
```
czoido, line 40:
```suggestion
This feature allows you to call the ``configure`` ``build`` and ``install`` method of each ``CMakeLists.txt``
```

reference/tools/cmake/cmake.rst
```diff
@@ -35,7 +35,27 @@ when a package is being built directly by Conan (create, install)
...
 
+In the following example, we can see what it would look like if we had two different ``CmakeLists.txt``
+in the ``folder1`` and ``folder2`` folders.
+
+.. code-block:: python
+
+    def build(self):
+        cmake = CMake(self)
+
+        # Configure and build source_folder/folder1/CmakeLists.txt
```
czoido, line 52:
```suggestion
        # Configure and build source_folder/folder1/CMakeLists.txt
```

czoido:
I think this section is a bit misleading right now. The example mixes the source folders (where the `CMakeLists.txt` live) with the output subfolders created.

From what I understand:
- `build_script_folder` → points to the location of the `CMakeLists.txt` inside the source tree (e.g., `cmake_one/`, `cmake_two/`).
- `subfolder` → defines an extra directory name inside both the `build_folder` to isolate the outputs (e.g., `one/`, `two/`).

So the text should make it explicit that the CMakeLists.txt are in the `cmake_one/` and `cmake_two/ ` source dirs, while the `build` artifacts go under `build/one/`, `build/two/`.

I would try to make this clear in the example.

Also, I think this needs to be marked as experimental.

memsharded (approved):

PR #4260, opened 2025-09-29
correct syntax for replace_requires system

lia-viam:
I may be wrong but I think `7zip/system` is a typo, it appears further in this document that the syntax should be `@system` and that `/system` is not valid

### DIFF
--- a/reference/config_files/profiles.rst
+++ b/reference/config_files/profiles.rst
@@ -39,7 +39,7 @@ They have this structure:
     zlib/1.2.12: zlib/[*]
 
     [replace_tool_requires]
-    7zip/*: 7zip/system
+    7zip/*: 7zip/*@system
 
     [platform_requires]
     dlib/1.3.22


AbrilRBS:
Hi @lia-viam thanks a lot for taking the time to submit this PR, we appreciate it, but I do think the current example is correct, let me explain

* The current `7zip/*: 7zip/system` line means: Replace every tool requirement to _any_ (the `*` in the pattern) version of 7zip, to a specific version called `system`.
* Your proposed line `7zip/*: 7zip/*@system` means: Replace ever tool requirement to _any_ version of 7zip, to that same version but add a `system` user to the reference

> and that /system is not valid

The `7zip/system` syntax means "7zip recipe with version `system`", which is usually used to refer to a recipe that just calls the system package manager to install the tool, see for example [this ncurses example](https://docs.conan.io/2/examples/tools/system/system_package/package_manager.html) on how those are used.

I hope my explanation is not too confusing, I'm closing this without merge, but please do feel free to open a new issue in the Conan client should you have any other questions regarding the replace requires feature, thanks! :)

PR #4297, opened 2025-11-05
warnings for local-recipes-index

memsharded:


AbrilRBS (approved):

devops/devops_local_recipes_index.rst
```diff
@@ -18,6 +18,26 @@ This setup is particularly useful for:
   :ref:`setup_local_recipes_index`.
 
 
+.. warning:: 
+
+    Using the ``local-recipes-index`` feature from a fork of ``conan-center-index`` Github repository,
+    without using a package server or relying on the ConanCenter package server can easily result in
+    missing dependencies due to old versions being removed by upstream ``conan-center-index``. The
+    recommendations are:
```
jcar87, line 26: I would probably tweak wording to make it more "generic" in terms of explaining the pitfalls, but more specific in terms of what happens with Conan Center.

Some suggestions:
- Mention explicity that the git repository at `conan-center-index`'s main purpose is to feed into the `conancenter` remote, and the ability to use it as a local-recipes-index is not a goal.
- Remove "building binaries from a private conan-center-index fork" from the paragraph above - and instead mention something like "test changes in multiple recipes together without exporting them before hand" (which from memory, I believe was one of the intended uses of local-recipes-index when it comes to operating a fork)
- Clarify, very explicitly, that unlike a remote Conan server, local-recipes-index does not support recipe revisions, but instead the current branch is a snapshot of all versions and revisions visible. This is already mentioned down below, but may be good to mention it here again
- Reword, that when using a local-recipes-index with a fork of `conan-center-index`, users must be careful in the two following scenarios:
    - bringing changes from upstream (`git pull`, `git merge master`, `git rebase master` etc) - because versions they rely on may no longer be exportable from master. Here the best mitigation is just to ensure the versions needed by the user are always in the repo, for example, they can always add their versions to a different subfolder (e.g. something other than "all") - which will minimise merge conflicts as the only potential change is "config.yml" in this scenario 
- when using the `conancenter` remote alongside local-recipes-index, you have two repositories that potentially export the same versions. This may be intentional or not - users need exercise care.

devops/devops_local_recipes_index.rst
```diff
@@ -18,6 +18,26 @@ This setup is particularly useful for:
...
+
+    - Use a package server to store your recipes and binaries built from your fork, and use it to resolve
+      dependencies.
```
jcar87, line 29: Since it is not possible to upload binary packages to a remote without _also_ uploading the recipes, I would probably just leave this whole section out, as it essential boils down to "if you need binaries, you need a server" therefore, using local-recipes-index AND needing to configure a remote, it is not entirely clear in which scenario `local-recipes-index` offers an advantage.

PR #4427, opened 2026-04-23
Modern in-page search on Conan 2.x docs via Pagefind

czoido:
Replaces the default Sphinx search on every Conan 2.x docs version with [Pagefind](https://pagefind.app) 
 
- 1.x legacy docs are untouched and keep the classic Sphinx form. 
- Older 2.x versions show an inline banner pointing users to the latest release.

for the latest version:
latest

if not latest, there's a warning about not searching in the latest.
not-latest

image

image

### DIFF
--- a/.ci/scripts/generate_documentation.py
+++ b/.ci/scripts/generate_documentation.py
@@ -111,8 +111,31 @@ with chdir(f"{sources_folder}"):
         # run('pip3 install colorama')
 
     # generate html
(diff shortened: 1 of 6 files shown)


franramirez688 (approved):
🫶

memsharded (approved):
Great

perseoGI (approved):
LGTM

PR #4454, opened 2026-05-14
Fix typo (double dash) in lockfiles documentation

Alexolut:


memsharded (approved):
Thanks for the fix @Alexolut !

PR #4473, opened 2026-06-25
Update Intel OneAPI documentation based on recent improvements

perseoGI:
Docs for https://github.com/conan-io/conan/pull/20075

image

image

### DIFF
--- a/reference/tools/intel.rst
+++ b/reference/tools/intel.rst
@@ -72,8 +72,8 @@ At first, ensure you are using a *profile* like this one:
         tools.intel:installation_path=C:\Program Files (x86)\Intel\oneAPI
 
 
-Compiler detection
-++++++++++++++++++
+Compiler mode
++++++++++++++
 
 Conan selects the appropriate compiler executables based on the ``compiler.mode`` setting:
 
--- a/reference/tools/intel.rst
+++ b/reference/tools/intel.rst
@@ -63,16 +63,17 @@ At first, ensure you are using a *profile* like this one:
         compiler=intel-cc
         compiler.mode=icx
         compiler.version=2026.0
+        compiler.runtime=dynamic
         build_type=Release
 
         [conf]
         tools.intel:installation_path=C:\Program Files (x86)\Intel\oneAPI
 
 
-Compiler auto-detection
-+++++++++++++++++++++++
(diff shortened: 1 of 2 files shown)


reference/tools/intel.rst
```diff
@@ -35,72 +37,99 @@ important variables in your local environment.
...
-At first, ensure you are using a *profile* like this one:
+   .. code-tab:: text Windows profile
 
-.. code-block:: text
-    :caption: *intelprofile*
+        [settings]
+        os=Windows
+        arch=x86_64
+        compiler=intel-cc
+        compiler.mode=icx
```
memsharded, line 65: Maybe be explicit and add there:

- compiler.cppstd: recommended in most cases to define it
- compiler.runtime: for Windows, even if "null" is allowed in this case, it would be very recommended to define it, like for ``msvc`` where it is mandatory.

reference/tools/intel.rst
```diff
@@ -35,72 +37,99 @@ important variables in your local environment.
...
+        tools.intel:installation_path=C:\Program Files (x86)\Intel\oneAPI
 
-    [buildenv]
-    CC=dpcpp
-    CXX=dpcpp
 
-    [conf]
-    tools.intel:installation_path=/opt/intel/oneapi
+Compiler auto-detection
++++++++++++++++++++++++
```
memsharded: Misleading title, it is not an auto-detection, but something like "compiler definition"

reference/tools/intel.rst
```diff
@@ -35,72 +37,99 @@ important variables in your local environment.
...
+        tools.intel:installation_path=/opt/intel/oneapi
+
+   .. code-tab:: text Windows profile
 
+        [settings]
+        os=Windows
+        arch=x86_64
+        compiler=intel-cc
+        compiler.mode=icx
+        compiler.version=2026.0
```
memsharded, line 129: Same as above for ``compiler.cppstd`` and ``compiler.runtime`` for windows.

reference/tools/intel.rst
perseoGI: You are right, I've updated the code

memsharded (approved):

PR #4482, opened 2026-06-25
Document Bazel 9.x support in BazelDeps

davidsanfal:
Docs for https://github.com/conan-io/conan/pull/20042


examples/tools/google/bazeltoolchain/build_simple_bazel_7x_project.rst
```diff
@@ -5,7 +5,8 @@ Build a simple Bazel 7.x project using Conan
 
 .. warning::
 
-    This example is Bazel >= 7.1 compatible.
+    This example is Bazel >= 7.1 compatible. For Bazel 9.x, use the ``include("//conan:conan_deps.MODULE.bazel")``
```
czoido: that is valid syntax starting in bazel 7.2 I think
I would have the two ways explained and clarify that the include syntax is only valid from 7.2 and say that the new is the recommended one

franramirez688: Yeah, I agree, I think that I'd keep it as it was `This example is Bazel >= 7.1 compatible.`

UPDATED: Bazel 7.2, yes, sorry

examples/tools/google/bazeltoolchain/build_simple_bazel_7x_project.rst
```diff
@@ -59,13 +60,14 @@ Let's have a look at each file's content:
 .. code-block:: python
     :caption: **MODULE.bazel**
 
-    load_conan_dependencies = use_extension("//conan:conan_deps_module_extension.bzl", "conan_extension")
-    use_repo(load_conan_dependencies, "fmt")
```
franramirez688, line 63: Here, I'd keep this one as the legacy or Bazel >=7.1 <9.x way, and then, let's explain later the new recommended way and Bazel >= 7.2 compatible

examples/tools/google/bazeltoolchain/build_simple_bazel_7x_project.rst
```diff
@@ -59,13 +60,14 @@ Let's have a look at each file's content:
...
+    include("//conan:conan_deps.MODULE.bazel")
 
 
 .. code-block:: python
     :caption: **main/BUILD**
 
+    load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
```
franramirez688, line 68: Same here: this is not needed for Bazel >=7.1 <9.x, but is mandatory for Bazel 9.x. Maybe a comment is enough?

examples/tools/google/bazeltoolchain/build_simple_bazel_7x_project.rst
franramirez688: Ok, rethinking this, I guess that we can keep the new recommended way as it's done in examples2 and that's all

franramirez688 (approved):

PR #4496, opened 2026-07-21
Document cpp_info.ignored_requires attribute

davidsanfal:
Document cpp_info.ignored_requires attribute

Docs for https://github.com/conan-io/conan/pull/20171

### DIFF
--- a/reference/conanfile/methods/package_info.rst
+++ b/reference/conanfile/methods/package_info.rst
@@ -163,18 +163,26 @@ Dependencies among components and to components of other requirements can be def
 of the component. The dependency graph for components will be calculated and values will be aggregated in the correct order for each field.
 
 Conan validates that every direct dependency of a recipe is used in some ``(cpp_info/components).requires``, raising an error otherwise.
-If a direct dependency is intentionally not propagated (for example, it is only used internally at build time), it can be added to
-``self.cpp_info.ignored_requires`` to silence that validation:
+This should not be necessary in most cases, as Conan already avoids propagating information that is not needed (for example, it
+doesn't propagate headers downstream unless ``transitive_headers=True``).
+
+If a dependency really has no relationship at all with the current package artifacts (not linked, not embedded, not exposing any of
+its headers), it can be added to ``self.cpp_info.ignored_requires`` to silence the validation:
 
 .. code-block:: python
 
     def requirements(self):
         self.requires("other/1.0")
-        self.requires("another/1.0")  # only used internally at build time
+        self.requires("another/1.0")  # not linked or embedded in this package in any way
 
     def package_info(self):
         self.cpp_info.components["cmp1"].requires = ["other::other"]
         self.cpp_info.ignored_requires = ["another"]
+
+If there is some relationship between the dependency and the package, even if it looks like a build-time only one (for example, a
+static library linked into the current shared library), it is recommended to still declare it via ``requires`` instead. Otherwise,
+changes in that dependency, like building it with a different ``shared`` value, could break the package without Conan being able to
+detect it.
 
 
 .. _reference_conanfile_methods_package_info_buildenv_info:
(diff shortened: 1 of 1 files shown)


AbrilRBS (approved):

reference/conanfile/methods/package_info.rst
```diff
@@ -162,6 +162,20 @@ If your package is composed by more than one library, it is possible to declare
 Dependencies among components and to components of other requirements can be defined using the ``requires`` attribute and the name
 of the component. The dependency graph for components will be calculated and values will be aggregated in the correct order for each field.
 
+Conan validates that every direct dependency of a recipe is used in some ``(cpp_info/components).requires``, raising an error otherwise.
+If a direct dependency is intentionally not propagated (for example, it is only used internally at build time), it can be added to
+``self.cpp_info.ignored_requires`` to silence that validation:
```
memsharded: This has to be more explicit and clear that it is not necessary to be defined in most cases.
Conan already handles the propagation correctly in many cases, avoiding to propagate headers for example downstream unless ``transitive_headers=True``.
The intention is not for "build time" thing, but for things that might really not propagate even in binary (embedded/linked) form, like for example a ``test_requires`` that is not part of the ``cpp_info`` of the current package, because it was used only to build and execute the tests, but was not linked in the current package artifacts.
The "build time" aspect has been misunderstood in the past. So better:
- Clarify this is an exception and shouldn't be necessary in most cases.
- If there is a relationship between the current ``cpp_info`` of the package or a component, and that dependency, because it was used in the build time, it is generally better to express that relationship. For example, if a shared library links a static library, it can be considered a "build-time" only, and then, user adds it to ``ignored_requires``. Then, the dependency changes to ``-o dep/*:shared=True``, and things will fail. As long as there is a relationship, the ``.requires`` is recommended.
- The ``ignored_requires`` is when there is really no relationship at all.

memsharded (changes requested):

reference/conanfile/methods/package_info.rst
davidsanfal: Updated!

reference/conanfile/methods/package_info.rst
```diff
@@ -162,6 +162,28 @@ If your package is composed by more than one library, it is possible to declare
...
+        self.requires("another/1.0")  # not linked or embedded in this package in any way
+
+    def package_info(self):
+        self.cpp_info.components["cmp1"].requires = ["other::other"]
+        self.cpp_info.ignored_requires = ["another"]
+
+If there is some relationship between the dependency and the package, even if it looks like a build-time only one (for example, a
+static library linked into the current shared library), it is recommended to still declare it via ``requires`` instead. Otherwise,
+changes in that dependency, like building it with a different ``shared`` value, could break the package without Conan being able to
+detect it.
```
memsharded, line 186:
```suggestion
Conan validates that every direct dependency of a recipe is used in some ``(cpp_info/components).requires``, raising an error otherwise.

If a dependency really has no relationship at all with the current package artifacts (not linked, not embedded, not exposing any of
its headers), it can be added to ``self.cpp_info.ignored_requires`` to silence the validation:

.. code-block:: python

    def requirements(self):
        self.requires("other/1.0")
        self.requires("another/1.0")  # not linked or embedded in this package in any way

    def package_info(self):
        self.cpp_info.components["cmp1"].requires = ["other::other"]
        self.cpp_info.ignored_requires = ["another"]

The definition of ``ignored_requires`` should not be necessary in most cases, as Conan already avoids propagating information that is not needed (for example, it doesn't propagate headers downstream unless ``transitive_headers=True``).

If there is some relationship between the dependency and the package, even if it looks like a build-time only one (for example, a
static library linked into the current shared library), it is recommended to still declare it via ``requires`` instead. Otherwise,
changes in that dependency, like building it with a different ``shared`` value, could break the package without Conan being able to
detect it.
```

memsharded (commented):
The sentence about "this should not be necessary" seems to apply to ``.requires``. I have moved it down so it applies to ``ignored_requires``.

memsharded (approved):

czoido (approved):

PR #4526, opened 2026-08-20
Docs or `package_type_traits`

AbrilRBS:
For https://github.com/conan-io/conan/pull/20228

### DIFF
--- a/reference/conanfile/attributes/consumers.inc
+++ b/reference/conanfile/attributes/consumers.inc
@@ -217,6 +217,17 @@ This attribute declares that the package has runtime artifacts that consumers sh
 This is useful for ``package_type="static-library"`` recipes that nevertheless also contain shared libraries, or by  ``package_type="header-library"`` that have some kind of resource that library needs at runtime,
 so that at runtime their folders are exposed to the consumers.
 
+It is important to still declare the ``package_type`` that best describes the package for linkage and
+``package_id`` purposes (see :ref:`the effect of dependencies on package_id<reference_binary_model_dependencies>`
+and :ref:`package_type trait inferring<reference_conanfile_package_type_trait_inferring>`), and use
+``package_type_traits`` only to fix the specific trait that ``package_type`` got wrong, instead of changing
+the ``package_type`` itself. For example, a package that is fundamentally a static library, meant to be
+embedded and to force a rebuild of its direct consumers like any other static library, should keep
+``package_type = "static-library"``, even if it also happens to ship a shared library or some other resource
+needed at runtime. Declaring it instead as ``package_type = "shared-library"`` would also change the
+``package_id`` computation mode and the linkage information propagated to consumers, when the only thing
+that actually needs fixing is the ``run`` trait.
+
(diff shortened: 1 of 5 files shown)


reference/conanfile/attributes/consumers.inc
```diff
@@ -207,3 +207,36 @@ To declare that a recipe provides the functionality of several different recipes
...
 If the attribute is omitted, the value of the attribute is assumed to be equal to the current package name. Thus, it's redundant for
 ``libjpeg`` recipe to declare that it provides ``libjpeg``, it's already implicitly assumed by Conan.
+
+.. _conan_conanfile_attributes_package_type_traits:
+
+package_type_traits
+-------------------
+
+This attribute declares that the package has runtime artifacts that consumers should be aware of.
+This is useful for ``package_type="static-library"`` recipes that nevertheless also contain shared libraries,
```
memsharded:
```suggestion
This is useful for ``package_type="static-library"`` recipes that nevertheless also contain shared libraries, or by  ``package_type="header-library"`` that have some kind of resource that library needs at runtime,
```

reference/conanfile/attributes/consumers.inc
```diff
@@ -207,3 +207,36 @@ To declare that a recipe provides the functionality of several different recipes
...
+
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "mypkg"
+        version = "1.0"
+        # This is a static library, but it also ships a shared library (e.g. a plugin) that
+        # consumers need available at runtime
+        package_type = "static-library"
+        runtime_artifacts = True
```
memsharded: This is not the implementation, seems outdated?

memsharded (commented):
This docs seems outdated, the final implementation is different, ``package_type_traits``

reference/conanfile/attributes/consumers.inc
```diff
@@ -207,3 +207,36 @@ To declare that a recipe provides the functionality of several different recipes
...
+
+    from conan import ConanFile
+
+    class Pkg(ConanFile):
+        name = "mypkg"
+        version = "1.0"
+        # This is a static library, but it also ships a shared library (e.g. a plugin) that
+        # consumers need available at runtime
+        package_type = "static-library"
+        package_type_traits = {"run": True}
```
memsharded, line 248: Add a clear note after this that only ``run: True`` is supported at the moment, and they shouldn't try to set anything else.

memsharded (commented):
I think the "intention" of the developer regarding the package-type is missing. 
The idea is that the dev wants to use the ``package_type = "static-library"`` even if it contains shared libraries, because of the ``package_id`` computation, the linkage requirement propagations, etc, that is, to generally think of it as a static-library. But also that such a static library might have some runtime needs (not necessarily a shared library, but a data file needed at runtime maybe?), and then, via the traits, it is possible to do that.

Otherwise, the reader might wonder, why I wouldn't just use ``package_type = "shared-library"``?

memsharded (approved):

